PR #911, opened 2019-01-08
#162246711 rejectConnectionWeirdBehaviour

diablourbano:
Implement solution to filter/remove rejectedInvitations on requester account, (also add a fix on a filter done on updatedContacts).

Apparently the original implementation of this was done when backend still didn't have an implementation and the consideration for what's implemented on frontend was that each event was part of a stream of events.

Current implementation of backend is a db state update (i.e. it removes the inviteConnection request from the db) so this implementation takes in count the type of the stream (TYPE_REJECTED) and based on that request for `fetchInvitations` with a param that's the rejectedConnection in order to exclude it and be able to update local contacts and local invitations accordingly.


diablourbano:
This PR is outdated

PR #966, opened 2019-02-11
prevent tx-note multiple decryption at the same time

poocart:


src/actions/txNoteActions.js
```diff
@@ -110,15 +116,23 @@ export const getTxNoteByContactAction = (username: string) => {
...
-      .filter(wsMessage => wsMessage.source === username && wsMessage.tag === 'tx-note')
-      .forEach(async (wsMessage) => {
-        await chat.client.decryptSignalMessage('tx-note', JSON.stringify(wsMessage));
-        await chat.deleteMessage(wsMessage.source, wsMessage.timestamp, wsMessage.requestId);
-      });
+    const data = await chat.client.receiveNewMessagesByContact(username, 'tx-note')
+      .then(JSON.parse)
+      .catch(() => {});
 
-    await chat.client.receiveNewMessagesByContact(username, 'tx-note').catch(() => null);
```
poocart: This call was not needed at all.

src/actions/txNoteActions.js
```diff
@@ -110,15 +116,23 @@ export const getTxNoteByContactAction = (username: string) => {
       }
     });
 
-    await webSocketMessagesReceived
```
poocart: We're not putting tx-note messages to reducer anymore, they're being decoded on the fly once received.

PR #969, opened 2019-02-11
Fix/163556069 scroll issue

diablourbano:
This PR upgrade RN 0.56 to 0.57.8 which adds estability, fixes on the RN platform and the fix (which is part of RN issue with react-native-gifted-chat) for [scrolling fast on android](https://www.pivotaltracker.com/story/show/163556069).

Something to notice: several attempts to upgrade to RN 0.58 were done but there're some dependencies on Android that cause exceptions making impossible this moment to upgrade to the latest RN version.


src/utils/shim.js
```diff
@@ -1,21 +1,3 @@
-/*
```
jsidorenko: Are you sure we don't need this here?

diablourbano: @jsidorenko because this file is autogenerated it removed the `Copyright notice` ... I'm updating the changes to include this `notice`

PR #973, opened 2019-02-14
Biometrics-#163867104

jsidorenko:
This adds support for the biometrical login on the wallet, ticket [#163867104](https://www.pivotaltracker.com/story/show/163867104).
These parts of the app were modified:
1) login logic
2) on user settings screen the new setting to enable/disable biometrics was added


jsidorenko:
This branch needs to be rebased and force pushed ;) Ask for the details, as you'll need to do the same with your old branches

diablourbano:
Oh, it makes sense @jsidorenko thanks 🙂

PR #977, opened 2019-02-14
HotFix/circleci PodFile.lock fix 

alazarevski:
Removed caching steps for pod install to avoid the updated version mismatch error with the node modules referenced in the latest react-native version update.
This is a temporary solution and will increase the hockeyapp staging deployment time on iOS to 30 minutes. 
TODO: add a conditional checksum check and use latest Podfile.lock from the repository in the case of a mismatch

### DIFF
--- a/src/utils/shim.js
+++ b/src/utils/shim.js
@@ -1,3 +1,21 @@
+/*
+    Pillar Wallet: the personal data locker
+    Copyright (C) 2019 Stiftung Pillar Project
+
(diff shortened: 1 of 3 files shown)


src/utils/shim.js
```diff
@@ -1,21 +1,3 @@
-/*
```
diablourbano: @alazarevski revert this removal, it's important to have this `copyright notice` besides this #LGTM

PR #982, opened 2019-02-17
fix intercom push messages being skipped

tkulpa:


android/app/src/main/java/com/pillarproject/wallet/PillarFirebaseMessagingService.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+
+    @Override
+    public void onMessageReceived(RemoteMessage remoteMessage) {
+        Map message = remoteMessage.getData();
+        if (intercomPushClient.isIntercomPush(message)) {
+            Log.d(TAG, "onMessageReceived intercom message");
+            intercomPushClient.handlePush(getApplication(), message);
+        } else if (remoteMessage.getNotification() != null) {
```
poocart: What if for `else`we add `super.onMessageReceived(remoteMessage)` only instead of duplicating default handler behaviour?

tkulpa: Yes, can be improved on this part

poocart: @tkulpa will you handle this part or we can proceed?

android/app/src/main/java/com/pillarproject/wallet/PillarFirebaseMessagingService.java
```diff
@@ -0,0 +1,62 @@
...
+import io.intercom.android.sdk.push.IntercomPushClient;
+import io.invertase.firebase.Utils;
+import io.invertase.firebase.messaging.RNFirebaseBackgroundMessagingService;
+import io.invertase.firebase.messaging.RNFirebaseMessagingService;
+
+public class PillarFirebaseMessagingService extends RNFirebaseMessagingService {
+    private static final String TAG = "PillarFirebaseMessaging";
+    private final IntercomPushClient intercomPushClient = new IntercomPushClient();
+    
+    @Override public void onNewToken(String refreshedToken) {
```
poocart: This method doesn't take any action as `RNFirebaseMessagingService` doesn't contain method `onNewToken`, do we really need it? @tkulpa

android/app/src/main/java/com/pillarproject/wallet/PillarFirebaseMessagingService.java
```diff
@@ -0,0 +1,62 @@
...
+        intercomPushClient.sendTokenToIntercom(getApplication(), refreshedToken);
+    }
+
+
+    @Override
+    public void onMessageReceived(RemoteMessage remoteMessage) {
+        Map message = remoteMessage.getData();
+        if (intercomPushClient.isIntercomPush(message)) {
```
poocart: Are you sure that this check works as data type `RemoteMessage` is different from `Map<String, String>` and seems to be incompatible?

PR #993, opened 2019-02-20
save draft text

diablourbano:
This PR implements logic to save draft for chat texts under following circumstances:

- user wants to change from screen (similar to whatsapp)
- user needs/wants to change app
- user receives a call

The `draft` will be stored until the user clear it or send it.


src/actions/chatActions.js
```diff
@@ -143,6 +149,49 @@ export const sendMessageByContactAction = (username: string, message: Object) =>
...
 };
 
+export const getChatDraftByContactAction = (contactId: string) => {
+  return async (dispatch: Function) => {
+    try {
+      const { drafts } = await storage.get('chat');
+      const [chatDraft, chatDrafts] = partition(drafts, { contactId });
+      const { draftText = '' } = chatDraft[0] || {};
+
+      if (!draftText || draftText === '') { return; }
```
jsidorenko:
```suggestion
      if (!draftText) return;
```

jsidorenko: `!draftText` covers empty strings as well

src/actions/chatActions.js
```diff
@@ -143,6 +149,49 @@ export const sendMessageByContactAction = (username: string, message: Object) =>
   };
 };
 
+export const getChatDraftByContactAction = (contactId: string) => {
+  return async (dispatch: Function) => {
+    try {
+      const { drafts } = await storage.get('chat');
```
jsidorenko: I think if we put here `const { drafts = [] }` then we could avoid wrapping everything into try..catch.
What do you think?

src/actions/chatActions.js
```diff
@@ -143,6 +149,49 @@ export const sendMessageByContactAction = (username: string, message: Object) =>
...
+  return async (dispatch: Function) => {
+    dispatch({
+      type: CLEAR_CHAT_DRAFT,
+    });
+  };
+};
+
+export const saveDraftAction = (contactId: string, draftText: string) => {
+  return async (dispatch: Function) => {
+    try {
```
jsidorenko: I'm wondering if Sentry would be able to catch the error when we wrapped everything into try..catch?

src/screens/Chat/Chat.js
```diff
@@ -409,6 +467,8 @@ class ChatScreen extends React.Component<Props, State> {
             </View>}
           {!this.state.isFetching &&
             <GiftedChat
+              text={chatText}
+              onInputTextChanged={(text) => this.updateChatInput(text)}
```
jsidorenko:
```suggestion
              onInputTextChanged={this.updateChatInput}
```

src/actions/chatActions.js
diablourbano: The reason for the try/catch is mostly for the `storage.get('chat')` part, if for some reason Storage crashes.

Setting a default value should be there, so thanks for noticing it.

jsidorenko: If for some reason Storage crashes we should probably block the app and notify the user...
If you think `storage.get` could crash, then it's better to wrap the code of this `.get()` method inside the storage.js file as we're using `storage.get()` in a lot of places across the app

jsidorenko: Btw Sentry wrapps the whole codebase with the try..catch by default

diablourbano: ok 👍

PR #999, opened 2019-02-21
revert call to binded method

diablourbano:
@jsidorenko in the [previous PR](https://github.com/pillarwallet/pillarwallet/pull/993) I did an update based on one of your suggestions, however that breaks and causes a crash in the chat screen.

This PR solves that. It's important to have this PR merge ASAP otherwise the chat is unusable.

### DIFF
--- a/src/screens/Chat/Chat.js
+++ b/src/screens/Chat/Chat.js
@@ -468,7 +468,7 @@ class ChatScreen extends React.Component<Props, State> {
           {!this.state.isFetching &&
             <GiftedChat
               text={chatText}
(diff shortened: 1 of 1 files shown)


src/screens/Chat/Chat.js
```diff
@@ -435,7 +435,7 @@ class ChatScreen extends React.Component<Props, State> {
     );
   };
 
-  updateChatInput(text) {
+  updateChatInput = (text) => {
```
jsidorenko: 👍

PR #1015, opened 2019-03-07
Feature/discourse login modal email

poocart:
Covers Discourse integration scenario when user doesn't have email.


src/screens/Home/Home.js
```diff
@@ -331,6 +343,14 @@ class HomeScreen extends React.Component<Props, State> {
     });
   };
 
+  goToProfileEmailSettings = () => {
+    const { navigation } = this.props;
+    this.setState({ forceCloseLoginApprovalModal: true }, () => {
```
jsidorenko: Could you explain why you 're doing this trick?
And it would be good to put this answer into a comment in the code. Use `// HACK: ...` or `// NOTE: ...`

PR #1028, opened 2019-03-13
Feature/163525244 deterministic key pair generation

alazarevski:
Added react-native-threads and additional thread bundle with the generation function
Added app entry point callers for connection key generation. Extended app storage to hold a pool of pre-connection key pairs.
Added additional package.json scripts to generate thread bundle jsbundles
Added accessor function and tests
Added new invitation functions and extended old type invitations to support connections between old and updated wallets.


.circleci/config.yml
```diff
@@ -797,14 +797,14 @@ workflows:
           filters:
             branches:
               only:
-                  - develop
+                  - feature/163525244_Deterministic-Key-Pair-Generation
```
antonioned: before merging, we need to change this back to develop

PR #1032, opened 2019-03-14
Feature/sdk private key removal

poocart:
Contains implementation of Pillar wallet SDK version in which wallet private key is removed from SDK.


src/actions/onboardingActions.js
```diff
@@ -276,19 +282,12 @@ export const validateUserDetailsAction = ({ username }: Object) => {
       type: UPDATE_WALLET_STATE,
       payload: CHECKING_USERNAME,
     });
-    const { mnemonic, importedWallet } = currentState.wallet.onboarding;
+    const { mnemonic } = currentState.wallet.onboarding;
     const mnemonicPhrase = generateMnemonicPhrase(mnemonic.original);
     dispatch(generateWalletMnemonicAction(mnemonicPhrase));
     await delay(200);
 
-    let wallet = importedWallet;
```
poocart: Became unused.

alazarevski: This will conflict another PR's handling of preConnectionKey generation call https://github.com/pillarwallet/pillarwallet/pull/1028/commits/a5a48269ae63e0d37146d8120ac720124a5ccb69

src/services/api.js
```diff
@@ -71,14 +71,18 @@ export default function SDKWrapper() {
...
+  tokensFailed?: ?Function,
+) {
   this.pillarWalletSdk = new PillarSdk({
-    privateKey: privateKey.slice(2),
     apiUrl: SDK_PROVIDER, // ONLY if you have platform running locally
     notificationsUrl: NOTIFICATIONS_URL,
     investmentsUrl: INVESTMENTS_URL,
     updateOAuthFn: updateOAuth,
     oAuthTokens: oAuthTokensStored,
+    tokensFailedCallbackFn: tokensFailed,
```
jsidorenko: Maybe it's better to call `onTokensExpired`? Or they (tokens) could fail because of other reasons?

src/actions/authActions.js
```diff
@@ -50,7 +58,7 @@ const Crashlytics = firebase.crashlytics();
 const storage = Storage.getInstance('db');
 const chat = new ChatService();
 
-export const loginAction = (pin: string) => {
+export const loginAction = (pin: string, successCallback?: Function) => {
```
jsidorenko: I would rename the `successCallback` to something more manifest, like 'onLoginSuccess'

src/services/api.js
poocart: They can actually if they are revoked or signature cert has changed, I can rename it though to onTokensFailed, ok?

src/actions/onboardingActions.js
poocart: I think it won't, although the code in your PR only needs `importedWallet` which is still there at the point you need it, is this ok?

src/actions/authActions.js
poocart: OK.

PR #1044, opened 2019-03-19
Fix/fetch all collectibles

vyvil:
This PR includes:
1. Updates on the way collectibles are fetched:
Open Sea API returns only 300 items (max) per one request. 
(Default is 100)
In order to fetch all user's collectibles multiple calls are executed in an interval of 1 second.

2. Updates on mapping collectibles' transaction data with transactions' data from BCX:
In order not to have long transaction history on the wallet, only last 100 is being fetched.
Now only those 100 will be visible on the wallet and its data will be extended using data fetched from BCX (transaction fee info).

3. Grammar update on transaction notification - added missing 'a' as per [164657456](https://www.pivotaltracker.com/story/show/164657456) ticket.


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -328,14 +341,15 @@ class ActivityFeed extends React.Component<Props, State> {
 
     const tokenTxHistory = history.filter(({ tranType }) => tranType !== 'collectible');
     const collectibleTxHistory = history.filter(({ tranType }) => tranType === 'collectible');
-    // merging BCX and OpenSea transaction data by hash
-    const mergedCollectibleTxHistory = values(
-      merge({}, keyBy(collectibleTxHistory, 'hash'), keyBy(collectiblesHistory, 'hash')));
+
+    // extending OpenSea transaction data with BCX data
+    const extendedCollectiblesTransactionData =
```
jsidorenko: #pickyJegor
Maybe we should call it like `const collectiblesHistory = this.mapOpenSeaAndBCXTransactionsHistory(openSeaTxHistory, bcxCollectiblesTxHistory)`?
and choose one option, either it's 'collectiblesHistory' or 'collectibleHistory' =)

src/screens/Assets/Assets.js
```diff
@@ -658,6 +664,9 @@ class AssetsScreen extends React.Component<Props, State> {
             <EmptyStateParagraph {...emptyStateInfo} />
           </EmptyStateWrapper>
         }
+        initialNumToRender={10}
+        removeClippedSubviews
+        viewabilityConfig={viewConfig}
```
jsidorenko: 👍

src/services/api.js
```diff
@@ -243,17 +244,14 @@ SDKWrapper.prototype.assetsSearch = function (query: string, walletId: string) {
...
-    .then(data => data.json())
-    .catch(() => ({ error: true }));
+  return new Promise((resolve, reject) => {
+    getLimitedData(`${OPEN_SEA_API}/assets/?owner=${walletAddress}&order_by=listing_date&order_direction=asc`,
+      [], 300, 0, 'assets', resolve, reject);
+  })
+    .then(response => {
+      return { assets: response };
+    })
+    .catch(() => { return { error: true }; });
```
jsidorenko:
```suggestion
    .catch(() => ({ error: true }));
```

src/services/api.js
```diff
@@ -243,17 +244,14 @@ SDKWrapper.prototype.assetsSearch = function (query: string, walletId: string) {
...
-      method: 'GET',
-      headers: {
-        Accept: 'application/json',
-        'Content-Type': 'application/json',
-        'X-API-KEY': OPEN_SEA_API_KEY,
-      },
-    }))
-    .then(data => data.json())
-    .catch(() => ({ error: true }));
+  return new Promise((resolve, reject) => {
```
jsidorenko: not sure, but it look like the `{` is obsolete

vyvil:
> Well done 👍

Thank you and thanks for suggestions :)

PR #1045, opened 2019-03-19
changed intercom logout calling place, added intercom fcm update

poocart:
Covers following Intercom Push Notification issues:
1. Intercom logout is called when app becomes inactive or is locked which unsubscribes from Push Notifications.
2. Intercom Push Notifications token is never updated after registration.

### DIFF
--- a/src/actions/authActions.js
+++ b/src/actions/authActions.js
@@ -47,6 +47,7 @@ import Storage from 'services/storage';
 import { navigate, getNavigationState, getNavigationPathAndParamsState } from 'services/navigation';
 import ChatService from 'services/chat';
 import firebase from 'react-native-firebase';
+import Intercom from 'react-native-intercom';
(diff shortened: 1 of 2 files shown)


src/actions/authActions.js
```diff
@@ -271,6 +273,7 @@ export const lockScreenAction = (onLoginSuccess?: Function, errorMessage?: strin
 
 export const logoutAction = () => {
   return async (dispatch: Function) => {
+    Intercom.logout();
```
jsidorenko: Maybe we should call it when we delete the wallet, and not on logout?

poocart: I think it's ok as the current flow of wallet delete fires logoutAction.

jsidorenko: you're right :D we should rename this action one day

PR #1046, opened 2019-03-19
Feature/badges

jsidorenko:


src/screens/Home/Home.js
```diff
@@ -630,6 +691,20 @@ class HomeScreen extends React.Component<Props, State> {
...
             <RecentConnectionsSpacer />
           }
+          {badges && badges.length ?
+            <BadgesWrapper>
+              <BadgesBlock>
+                <View style={{ backgroundColor: baseColors.snowWhite }}>
+                  <BadgesSubtitle subtitle title="game of badges." />
+                </View>
+                <BadgesScrollView horizontal nestedScrollEnabled overScrollMode="always">
+                  {this.renderBadges()}
```
vyvil: If user might have lots of badges in the long run, maybe we should consider using horizontal FlatList here to optimise rendering :)

src/screens/Home/Home.js
```diff
@@ -630,6 +691,20 @@ class HomeScreen extends React.Component<Props, State> {
...
+                </BadgesScrollView>
+              </BadgesBlock>
+            </BadgesWrapper> :
+
+            <BadgesSpacer />
```
vyvil: Or you could not render here anything and set stickyHeaderIndices to 3 only if badges exists

src/screens/Home/Home.js
jsidorenko: Good idea, as for now user could have max 5-6 badges, so this would be easy to render. When we move to the new Dmitryj's layout, then it would be good to convert this into the FlatList

PR #1051, opened 2019-03-20
Fix/nav top title overlap

poocart:
https://www.pivotaltracker.com/story/show/164678792

This PR includes few fixes in following order:
1. Fixed full width title overlapping top bar left/right controls.
2. Fixed Android cropping text issue that needed both to define other font and add `fontFamily` style to component level which solves text overflow ellipsis method as described here – https://github.com/facebook/react-native/issues/15114#issuecomment-364458149.
3. Fixed iOS issue that appeared after solving No 1 – needed to realign title to center.

### DIFF
--- a/src/components/Title/Title.js
+++ b/src/components/Title/Title.js
@@ -54,7 +54,7 @@ const Text = styled(BoldText)`
   font-size: ${props => props.subtitle ? fontSizes.medium : fontSizes.large};
   font-weight: ${fontWeights.bold};
   ${({ align }) => align === 'center' && `
-    width: 100%;
+    flex: 1;
     text-align: center;
   `}
 `;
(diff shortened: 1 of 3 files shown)


src/components/Title/Title.js
```diff
@@ -87,6 +91,13 @@ const Title = (props: Props) => {
 
   const noBlueDotNeeded = noBlueDot || !title;
 
+  /**
+   *  `fontFamily` has to stay here as it affects font rendering
+   *  otherwise if it's taken then once font is being shorten with ellipsis
+   *  on Android it gets cut
+   */
+  const titleStylesFix = { ...titleStyles, fontFamily: `Aktiv Grotesk App${Platform.OS === 'android' ? '_bold' : ''}` };
```
diablourbano: I would suggest to create an styled component from Text and set this styles there. Also `titleStylesFix` could be renamed to, for example, `AktivText`

```
const AktivText = styled(Text)`
  fontFamily: 'Aktiv Grotesk' App${Platform.OS === 'android' ? '_bold' : ''};
`
```

Otherwise this looks good.

poocart: 👍

PR #1054, opened 2019-03-22
Feature/offline calls - #164085440

jsidorenko:
This feature would allow making so-called 'offline api calls'.
It's when we need to post some data to the backend server (usually only once) and be sure that call would happen immediately if the user is online, or will be put in a queue and stored locally if the user is offline.

Usage example:
```
// in action:
dispatch(offlineApiCall('selfAwardBadge', walletId, 'for-backup-phrase-set'));
```

P. S. This would resolve the ticket [164085440](https://www.pivotaltracker.com/story/show/164085440) ("Handle the situation when the user backed up his wallet beeing offline")


src/actions/appActions.js
```diff
@@ -76,6 +77,10 @@ export const initAppAndRedirectAction = (appState: string, platform: string) =>
       const { collectiblesHistory = [] } = await storage.get('collectiblesHistory');
       dispatch({ type: SET_COLLECTIBLES_TRANSACTION_HISTORY, payload: collectiblesHistory });
 
+      const { offlineQueue = [] } = await storage.get('offlineQueue');
+      dispatch({ type: UPDATE_OFFLINE_QUEUE, payload: offlineQueue });
+      dispatch({ type: START_OFFLINE_QUEUE });
```
jsidorenko: Maybe it's better to rename to `PROCESS_OFFLINE_QUEUE`?

alazarevski: It is ok to have it as "start" for this action, some queues use "digest", "process" it's up to you.

PR #1060, opened 2019-03-27
chore/ios swift version increase 4.2

alazarevski:
Added latest signal protocol node_module version
Added 4.2 version instead of 3.0 for Swift


.circleci/config.yml
```diff
@@ -808,7 +810,7 @@ workflows:
           filters:
             branches:
               only:
-                  - master
+                  - chore/ios_swift_version_increase4.2
```
poocart: change this back to master?

antonioned:
@alazarevski I think that this PR is obsolete now and we can just close it ?

alazarevski:
> @alazarevski I think that this PR is obsolete now and we can just close it ?

yes, this should be deleted

PR #1074, opened 2019-04-12
Update/Android gradle and libraries update for RN0.59.4

alazarevski:
Second part of RN version Upgrades to Android gradle and dependencies


jsidorenko:
I would try to remove the existing libs from rn-libs folder as they were put here mostly because of the old graddle version

PR #1075, opened 2019-04-14
add and configure Reactotron

diablourbano:
The idea is not to replace `RN Debugger` but include the option to use `Reactotron` which is more effective to watch logs and investigate the app state.

These are some screenshots of what you'll be able to see:

Screen Shot 2019-04-27 at 18 02 48

Screen Shot 2019-04-27 at 18 01 27


reactotron.config.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+//
+global.Reactotron = null;
+
+/* eslint no-undef: 0, global-require: 0 */
+export const ReactotronConfig = function ReactotronConfig() {
+  const isTest = !!process.env['IS_TEST']; // eslint-disable-line dot-notation
```
jsidorenko:
```suggestion
  const isTest = !!process.env.IS_TEST;
```

jsidorenko:
@diablourbano Can you provide some links/examples on how to use this tool?
1) I tried to put `Reactotron.log('test');` in the arbitrary js file and got the `Can't find variable: Reactotron` error
2) I found how to see all the ongoing api calls, but can't find the way to see all the redux actions those were dispatched
3) What is `Image overlay` in React Native tab?

src/testUtils/jestSetup.js
```diff
@@ -25,6 +25,8 @@ import { utils, HDNode } from 'ethers';
 import StorageMock from './asyncStorageMock';
 import FirebaseMock from './firebaseMock';
 
+process.env['IS_TEST'] = 'TEST'; // eslint-disable-line dot-notation
```
jsidorenko: when you start jest it sets the process.env.NODE_ENV to 'test', I think you can re-use that. Additionally, you can create a simple function in `utils/environment.js` file to check if the current env is `test`

diablourbano: > when you start jest it sets the process.env.NODE_ENV to 'test', I think you can re-use that.

actually no, that's the reason I set it there, because of the following:

When the app starts, it doesn't matter if it's `yarn start` or `yarn test`, `process.env.NODE_ENV` is set to `development` and `test` respectively, however before anything `./utils./shim.js` is being invoked and it fixes `process.env.NODE_ENV=development` so after that in `jest` if you check `process.env.NODE_ENV` it'll be `development`. Even more, you won't be able to modify it, that value is *fixed*. I tested and tried several alternatives and that's the reason here I'm adding a different variable `process.env.IS_TEST` to be able to use TEST environment and conditions.

> Additionally, you can create a simple function in utils/environment.js file to check if the current env is test

Thanks for the suggestion, I'll add that

grilix: I'd suggest avoiding checking if we're running in test environment as much as we can. I'm trying to understand why this is required, but I'm pretty sure there should be a way of avoiding it.

grilix: Maybe we can have an environment variable like `USE_REACTOTRON`, for example? (Assuming this is used just for that)

diablourbano: > I'd suggest avoiding checking if we're running in test environment as much as we can.

I agree with you when you say "as much as we can" but there're moments when the system should know it's running in a different environment (similar to differentiate between `prod` and `dev` environment) and this is one of those cases.

The idea behind Reactotron is to run anytime you want while you're running your app in `development mode` and check `logs` or the `redux state` while developing, however you don't want to include it (and its dependencies) while you're testing it, for two reasons:

1) it's unnecessary since you're not testing it
2) in this particular case you'll get an error on testing trying to use something outside `nodejs/console` environment as you can see in the following screenshot.

image

That crash happens because `App.js` tests how the app loads, and Reactotron is configured and started at that point. You can check that [here](https://github.com/pillarwallet/pillarwallet/pull/1075/files#diff-04298622441e55a9d9b5691873f8490bR34) and [here](https://github.com/pillarwallet/pillarwallet/pull/1075/files#diff-c375b4b4e917a1fddd16650469169fcdR20)

> Maybe we can have an environment variable like USE_REACTOTRON
We could, but once we add that variable into `.env` file we're still under the previous scenario, `jest` runs and `node.process.env` is development (because of the `shim.js` detail [commented above](https://github.com/pillarwallet/pillarwallet/pull/1075#discussion_r279015053)) and we don't have a way to tell jest to not include `Reactotron` dependency.

The thing is, even you're right implying the app shouldn't behave differently in `test`, `dev` or `prod` environment, the importance to know in which environment you are is not excluded, that's important on web and so does on native.

PR #1076, opened 2019-04-17
Feature/add hobbes ui for rapid ui dev

diablourbano:
While `Personas` feature (screens) were being implemented [HobbesUI]() was used to improve the UI development, this PR doesn't include those screens, but include the integration with the library and some components that were used for `Personas` screens.

Those components include some fixes like Header title position and title wrapping, it also includes useful views like Colors and Fonts used across the app.

If you check README you'll see that the library is not invasive, basically if you want to use it, change the `.env` variable `HOBBESUI=NOHOBBESUI` to `HOBBESUI=HOBBESUI` and restart the app.

This is available only on `dev` (i.e. local) environment.

Screenshots attached to show what you can expect to find.

Screen Shot 2019-04-14 at 02 13 18
Screen Shot 2019-04-14 at 02 13 29
Screen Shot 2019-04-14 at 02 13 42
Screen Shot 2019-04-14 at 02 13 48
Screen Shot 2019-04-14 at 02 14 19


diablourbano:
Based on last conversation with @alazarevski this branch is already tested with branch `RN0.59` and it works without any problem.

App.js
```diff
@@ -122,7 +122,11 @@ class App extends React.Component<Props, *> {
 
   render() {
     const { isFetched } = this.props;
-    if (!isFetched) return null;
+    const isHobbesUI = HOBBESUI === 'HOBBESUI' &&
+      process.env['NODE_ENV'] === 'development'; // eslint-disable-line dot-notation
```
jsidorenko: 1) process.env.NODE_ENV
2) I would move this to the utils/environment.js file

grilix:
We'd need to clean this up to remove unrelated complexity if we want to merge. We'll close it for now.

PR #1077, opened 2019-04-18
Fixed pod repo update command before pod install

alazarevski:
This can be postponed because of the react-native module versions updates which will be done in the future

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ aliases:
   - &install_pods
     run:
       name: Install CocoaPods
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -797,7 +803,7 @@ workflows:
           filters:
             branches:
               only:
-                  - develop
+                  - fix/pod_repo_update
```
jsidorenko: Don't forget to update this back to `develop` before merging

diablourbano:
The changes proposed on this PR were already included in PR #1088

PR #1085, opened 2019-04-23
Added referral code and phone verification

PeterKow:
Added: 
- code referral
- phone verification

phonev5
ref


src/actions/userActions.js
```diff
@@ -43,6 +43,56 @@ export const updateUserAction = (walletId: string, field: Object, callback?: Fun
...
 
+export const createOneTimePasswordAction = (walletId: string, field: Object, callback?: Function) => {
+  return async (dispatch: Function, getState: Function, api: Object) => {
+    const response = await api.createOneTimePassword({ walletId, ...field });
+    const { responseStatus } = response;
+
+    if (responseStatus === 200 && callback) {
+      callback();
+    } else {
+      dispatch(({
```
jsidorenko: Why do we need those double brackets?

src/navigation/appNavigation.js
```diff
@@ -64,6 +64,7 @@ import BackupPhraseScreen from 'screens/BackupPhrase';
 import BackupPhraseValidateScreen from 'screens/BackupPhraseValidate';
 import CollectibleScreen from 'screens/Collectible';
 import BadgeScreen from 'screens/Badge';
+import OTPScreen from 'screens/OTP/OTP';
```
jsidorenko:
```suggestion
import OTPScreen from 'screens/OTP';
```

src/screens/OTP/OTP.js
```diff
@@ -53,24 +57,30 @@ class OTP extends React.Component<Props> {
...
   };
 
   handleOTPResend = () => {
+    const { createOneTimePassword, user } = this.props;
+    createOneTimePassword(user.walletId, { phone: user.phone });
   };
 
   render() {
+    const { navigation } = this.props;
+    const phone = this.props.navigation.getParam('phone');
```
jsidorenko:
```suggestion
    const phone = navigation.getParam('phone');
```

src/screens/OTP/OTP.js
```diff
@@ -79,7 +89,13 @@ class OTP extends React.Component<Props> {
 }
 
 const mapDispatchToProps = (dispatch) => ({
-  confirmOTP: (code: string) => dispatch(confirmOTPAction(code)),
+  confirmOTP: (props: Object, callback: Function) => dispatch(verifyPhoneAction(props, callback)),
```
jsidorenko: It would be good to type the first argument

src/screens/Profile/Profile.js
```diff
@@ -54,9 +54,11 @@ import { delay } from 'utils/common';
 import ProfileSettingsItem from './ProfileSettingsItem';
 import EditProfile from './EditProfile';
 import SettingsModalTitle from './SettingsModalTitle';
+import ReferralCodeModal from './ReferralCodeModal';
 
 // sections
 import AppearanceSettingsSection from './AppearanceSettingsSection';
+import { isProdEnv } from '../../utils/shim';
```
jsidorenko:
```suggestion
import { isProdEnv } from 'utils/shim';
```

src/screens/Profile/Profile.js
```diff
@@ -44,8 +45,7 @@ import {
   changeUseBiometricsAction,
   updateAppSettingsAction,
 } from 'actions/appSettingsActions';
-import { updateUserAction } from 'actions/userActions';
-import { repairStorageAction } from 'actions/appActions';
```
jsidorenko: Are you sure we don't need this? I can still see the reference in that file

src/screens/Profile/Profile.js
```diff
@@ -565,10 +633,11 @@ const mapStateToProps = ({
 
 const mapDispatchToProps = (dispatch: Function) => ({
   saveBaseFiatCurrency: (currency) => dispatch(saveBaseFiatCurrencyAction(currency)),
-  repairStorage: () => dispatch(repairStorageAction()),
```
jsidorenko: Same as above

src/screens/Profile/ReferralCodeModal.js
```diff
@@ -0,0 +1,43 @@
...
+`;
+
+const shareOnSocialMedia = (sharerUrl: string, closeModal) => () => {
+  const options = {
+    message: 'Get your 25 PLR',
+    url: sharerUrl,
+  };
+  Share.open(options)
+    .then(closeModal)
+    .catch((() => {}));
```
jsidorenko:
```suggestion
    .catch(() => {});
```

src/screens/Profile/Profile.js
```diff
@@ -322,6 +348,35 @@ class Profile extends React.Component<Props, State> {
             />
           </Wrapper>
         </SlideModal>
+        <SlideModal
+          isVisible={this.state.visibleModal === 'referralCode'}
+          title="Referral code"
+          onModalHide={this.toggleSlideModalOpen}
+        >
+          <ReferralCodeModal username={user.username} closeModal={this.toggleSlideModalOpen} />
```
jsidorenko:
```suggestion
          <ReferralCodeModal username={user.username} onModalClose={this.toggleSlideModalOpen} />
```

src/utils/shim.js
```diff
@@ -21,6 +41,7 @@ if (typeof localStorage !== 'undefined') {
   localStorage.debug = isDev ? '*' : ''
 }
 
+export const isProdEnv = NETWORK_PROVIDER === 'homestead'
```
jsidorenko: shim.js file is auto-generated, once you install any npm package this function will be deleted

src/utils/validators.js
```diff
@@ -72,3 +72,8 @@ export function isValidCityName(cityName: string) {
   const re = /^([a-zA-Z\u0080-\u024F]+(?:. |-| |'))*[a-zA-Z\u0080-\u024F]*$/u;
   return re.test(cityName);
 }
+
+export function isValidUKPhone(phone: string) {
+  const re = /^(\+44\d{10}\b)/;
```
jsidorenko: Shall we support numbers those start with 0 instead of +44? In the UK it's a very common format

src/screens/Profile/Profile.js
```diff
@@ -191,8 +206,20 @@ class Profile extends React.Component<Props, State> {
...
     Keyboard.dismiss();
-    const { updateUser, user } = this.props;
-    updateUser(user.walletId, field, () => this.toggleSlideModalOpen(null));
+    const {
+      updateUser,
+      user,
+      navigation,
+      createOneTimePassword,
+    } = this.props;
+    const createOTP = () => {
```
jsidorenko: As `handleUserFieldUpdate` is a generic function, you will call the `createOneTimePassword` even if the user changes his first name

src/actions/userActions.js
PeterKow: - [ ] remove brackets

src/navigation/appNavigation.js
PeterKow: aaa that's why you have `package.json`
- [ ] remove OTP

src/screens/OTP/OTP.js
PeterKow: - [ ] remove `this.props.`

src/screens/OTP/OTP.js
PeterKow: yeah I can see we have more `Objects` than types, therefore in `verifyPhoneAction` a function, I split it to `{ walletId, phone, oneTimePassword }`

Also, all SDK is untyped, but node-sdk is fully typed as it's TS, a bit of waste. We can use some generator to get all types in `Flow`

src/screens/Profile/Profile.js
PeterKow: nice 👍 catch!, my editor picked it as unused, sad that flow didn't pick this up
- [ ] revoken repairStorage

src/utils/shim.js
PeterKow: 😢 any other suggestion for this then? @jsidorenko

src/utils/validators.js
PeterKow: we still need verification that it's UK number, so probably we need country picker, but this is for next release. As it needs to be changed, I would leave it with +44 for now, had a quick chat with Dmirty and he is ok.

src/screens/Profile/Profile.js
jsidorenko: It's because you left it in Props definition, so Flow thinks it's still here. But yeah, at least linter should warn about that

src/utils/shim.js
PeterKow: ok moved to `utils/environment` for now

src/screens/OTP/OTP.js
jsidorenko: We need to start to write types and use less `Object` :) I put that kind of comment only for this method as we should avoid having magic function params those are objects and are untyped

src/utils/shim.js
jsidorenko: just put in any other file from utils folder

src/utils/validators.js
jsidorenko: 👍

src/screens/Profile/Profile.js
PeterKow: You are totally right, I will move it to `handleUserPhoneFieldUpdate` 👍

src/screens/OTP/OTP.js
PeterKow: added

export type VerificationPhoneAction = {
  wallet: string,
  phone: string,
  oneTimePassword: string,
}```

PeterKow:
@jsidorenko thanks, good comments all resolved 👍

src/components/ProfileForm/profileFormDefs.js
```diff
@@ -95,8 +99,18 @@ CityStructDef.getValidationErrorMessage = (city): string => {
   return 'Please specify your city';
 };
 
+PhoneStructDef.getValidationErrorMessage = (phone): string => {
+  if (phone) {
+    if (!isValidUKPhone(phone)) {
+      return 'Please enter a valid UK phone number with +44';
```
vyvil: Why only UK numbers?

PeterKow: Hope this will clarify, here -> https://github.com/pillarwallet/pillarwallet/pull/1085#discussion_r278007537

PR #1091, opened 2019-04-24
fix/connection Generation WalletId access Fallback

alazarevski:
Fix walletId not accessible from State case.


src/actions/oldInvitationsActions.js
```diff
@@ -142,7 +142,9 @@ export const fetchOldInviteNotificationsAction = () => {
     });
 
     await dispatch(updateConnectionsAction());
-    dispatch(getExistingChatsAction());
+    await dispatch(getExistingChatsAction());
```
poocart: `return dispatch(getExistingChatsAction())` will await for promise to complete

src/actions/oldInvitationsActions.js
```diff
@@ -142,7 +142,9 @@ export const fetchOldInviteNotificationsAction = () => {
...
+
+    return Promise.resolve(true);
```
jsidorenko: why do we need this? (just curious)

alazarevski: during debug, it ran async even though await was called before the dispatch

src/actions/oldInvitationsActions.js
alazarevski: Changed and tested in 2a6fa98

PR #1096, opened 2019-04-25
Fix/Prepend connectionKeys on error

alazarevski:
- Prepend connectionKeys on error
- Better import/update flow
- Switch back to decrypting status after generating keys (or on error)
- Tests update

### DIFF
--- a/src/actions/connectionKeyPairActions.js
+++ b/src/actions/connectionKeyPairActions.js
@@ -99,8 +99,7 @@ export const mapIdentityKeysAction = (connectionPreKeyCount: number, theWalletId
         } else {
           const { sourceIdentityKey, targetIdentityKey } = conn;
           const errorConnKey = currentConnectionKeyPairs.find((connKeyPair) => {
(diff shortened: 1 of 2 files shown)


src/actions/connectionKeyPairActions.js
```diff
@@ -174,30 +217,40 @@ export const updateConnectionKeyPairs = (mnemonic: string, privateKey: string, w
...
+      await dispatch({
+        type: UPDATE_CONNECTION_KEY_PAIRS,
+        payload: resultConnectionKeys,
+      });
+      await dispatch(saveDbAction('connectionKeyPairs', { connectionKeyPairs: resultConnectionKeys }, true));
+    } catch (e) {
+      await dispatch({
+        type: UPDATE_WALLET_STATE,
+        payload: DECRYPTING,
+      });
```
grilix: Just to make sure, is this the correct event we want to dispatch? I'd have expected it to be an error 🤔

alazarevski: Sentry is declared after the call so in a future improvement it can be used here.

PR #1097, opened 2019-04-26
Added confirm screen for claiming token

PeterKow:
- confirm claim screen
- next is to update SDK with API
claim screen

image

claiming


src/actions/referralsActions.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export const claimTokensAction = (props: ClaimTokenAction, callback?: Function) => {
+  return async (dispatch: Function, getState: Function, api: Object) => {
+    const response = await api.claimTokens(props);
+    const { responseStatus } = response;
+
+    if (responseStatus === 200 && callback) {
+      callback({ error: false });
+    } else {
```
grilix: This is a bit weird to me; For example, if we don't receive a callback, the `else` block is gonna be reached even if `responseStatus === 200`, which will send the user a "Please try again later", but since the API returned success, the token is already claimed.

What do we expect to happen if the response is success but we didn't receive a callback?

Maybe something like... ?

```js
    if (responseStatus === 200) {
      if (callback) {
        callback({ error: false });
      }
    } else {
    }
```

Not really sure.

src/screens/Referral/ConfirmClaimScreen.js
```diff
@@ -0,0 +1,140 @@
...
+        this.setState({ isError: true, isFetching: false });
+      } else {
+        navigation.goBack();
+      }
+    });
+  };
+
+  render() {
+    const { navigation } = this.props;
+    const code = navigation.getParam('code') ? navigation.getParam('code') : 'No code';
```
vyvil: Here as well :) You can even pass `code` as a prop to `this.handleClaim` and save one look up :)

src/screens/Referral/ConfirmClaimScreen.js
```diff
@@ -0,0 +1,140 @@
...
+    const { isFetching, isError } = this.state;
+    return (
+      <Container>
+        <Header gray title="claim tokens" onBack={() => navigation.goBack(null)} />
+        <MainWrapper>
+          <TextCode>From {code}</TextCode>
+          { isError ? <ErrorClaim /> : null }
```
vyvil: You can simplify this and following ones by changing it like this:
`{isError && <ErrorClaim />}`
(if variable is boolean. If not, it's better to turn it into one (by adding !!), otherwise Android will throw errors of text not being placed inside of `<Text> `component.

src/screens/Referral/ConfirmClaimScreen.js
```diff
@@ -0,0 +1,140 @@
...
+import AssetPattern from 'components/AssetPattern';
+import { baseColors, fontSizes } from 'utils/variables';
+import { claimTokensAction } from 'actions/referralsActions';
+import type { ClaimTokenAction } from 'actions/referralsActions';
+import ProcessingClaim from './ProcessingClaim';
+import ErrorClaim from './ErrorClaim';
+
+const Center = styled.View`
+  alignItems: stretch;
+  margin-top: 30px
```
vyvil: Missed couple of `;` :)
(here and in the line below)

src/screens/Referral/ConfirmClaimScreen.js
```diff
@@ -0,0 +1,140 @@
...
+const TokenValue = styled(BoldText)`
+  font-size: ${fontSizes.giant}px;
+  text-align: center;
+`;
+
+const TextCode = styled(LightText)`
+  font-size: ${fontSizes.medium}px;
+  color: ${baseColors.coolGrey};
+  text-align: center;
+  margin-bottom: 30px
```
vyvil: Missed `;`

src/screens/Referral/ErrorClaim.js
```diff
@@ -0,0 +1,46 @@
...
+import { baseColors, fontSizes } from 'utils/variables';
+import { BoldText, LightText } from 'components/Typography';
+import Animation from 'components/Animation';
+
+const animationFailure = require('assets/animations/transactionFailureAnimation.json');
+
+const Status = styled(BoldText)`
+  font-size: ${fontSizes.giant}px;
+  text-align: center;
+  margin-top: 35px
```
vyvil: `;` :(
(here and in the line below)

src/screens/Referral/ErrorClaim.js
```diff
@@ -0,0 +1,46 @@
...
+`;
+
+const Center = styled.View`
+  align-items: center; 
+  flexDirection: column;
+  justify-content: center;
+  margin-horizontal: 20px;
+`;
+
+const errorMessage = "We can't verify your code at this time.";
```
vyvil: Curious why you decided to put this copy as a variable while adding others as strings in the code? :)
I guess it would be better to unify it :)

src/screens/Referral/ProcessingClaim.js
```diff
@@ -0,0 +1,44 @@
...
+import * as React from 'react';
+import styled from 'styled-components/native';
+import { baseColors, fontSizes } from 'utils/variables';
+import { BoldText, LightText } from 'components/Typography';
+import Spinner from 'components/Spinner';
+
+const Status = styled(BoldText)`
+  font-size: ${fontSizes.giant}px;
+  text-align: center;
+  margin-top: 35px
```
vyvil: `;` Here as well :(
You are reusing this styling a lot. Maybe you could create a dumb component and import it to all related screens rather than repeating it? :)

src/screens/Referral/ProcessingClaim.js
```diff
@@ -0,0 +1,44 @@
...
+  margin-bottom: 25px
+`;
+
+const Text = styled(LightText)`
+  font-size: ${fontSizes.small}px;
+  color: ${baseColors.coolGrey};
+  text-align: center;
+
```
vyvil: Extra line :)

src/services/api.js
```diff
@@ -212,6 +213,30 @@ SDKWrapper.prototype.verifyPhone = function (user: Object) {
...
+    .then(() => ({ responseStatus: 200, walletId, code }))
+    .catch((error) => {
+      const {
+        response: {
+          status,
+          data: { message } = {},
+        },
+      } = error;
+      Sentry.captureException({
+        error: 'Can\'t claim code',
```
vyvil: Maybe this message could be a bit more explicit to make life easier. 
Even providing info that it's the code for claiming referral tokens that fails, would help :)

src/screens/Referral/ConfirmClaimScreen.js
PeterKow: nice! 👍 thanks

src/actions/referralsActions.js
PeterKow: good catch, updated it and another place from where I copied it :D 
```
if (responseStatus === 200) {
      if (callback) callback({ error: false });
```

src/screens/Referral/ConfirmClaimScreen.js
PeterKow: 👍  updated for all 3

src/screens/Referral/ConfirmClaimScreen.js
PeterKow: yeah, do we need it? I wasn't sure?

vyvil: Yeah :)

src/screens/Referral/ErrorClaim.js
PeterKow: because tslin is complaining when RN is satisfied :D 
image
and when it's escaped tslint will not like it

src/services/api.js
PeterKow: hmm then we pass more details from sever anyway like `message` to Sentry but changed to `Can\'t claim referral code'`

src/screens/Referral/ProcessingClaim.js
PeterKow: hmm it's just 2 components 🤔

src/screens/Referral/ConfirmClaimScreen.js
PeterKow: do then this style is not applied if there is no `;` ? Just wondering why nothing is faling

vyvil: It does - styled components now has no-semicolon support. It's mainly for formatting

src/screens/Referral/ProcessingClaim.js
vyvil: Hm, miscalculated - sorry. So then it's okay :)

src/screens/Referral/ConfirmClaimScreen.js
PeterKow: ok, I was just scared that all my styles don't work... I wish prettier could fix it for me... - done ✅

src/services/api.js
vyvil: 👍

PeterKow:
Added deeplink handling: 
deep error
deep

PR #1107, opened 2019-05-01
improvement/Connection Identity Key Loss Prevention

alazarevski:
Connection loss prevention from api errors

### DIFF
--- a/src/actions/authActions.js
+++ b/src/actions/authActions.js
@@ -53,7 +53,6 @@ import { updateOAuthTokensCB, onOAuthTokensFailedCB } from 'utils/oAuth';
 import { setupSentryAction } from 'actions/appActions';
 import { signalInitAction } from 'actions/signalClientActions';
 import { updateConnectionKeyPairs } from 'actions/connectionKeyPairActions';
-import { updateConnectionsAction } from 'actions/connectionsActions';
 import { saveDbAction } from './dbActions';
 
 const Crashlytics = firebase.crashlytics();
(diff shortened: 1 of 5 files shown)


src/actions/authActions.js
```diff
@@ -120,12 +119,9 @@ export const loginAction = (pin: string, touchID?: boolean = false, onLoginSucce
...
         await dispatch(signalInitAction({ ...signalCredentials, ...OAuthTokensObject }));
         user = merge({}, user, userInfo);
-        if (generateNewConnKeys) {
-          await dispatch(updateConnectionKeyPairs(wallet.mnemonic, wallet.privateKey, user.walletId));
-        } else {
-          dispatch(updateConnectionsAction(user.walletId));
-        }
         dispatch(saveDbAction('user', { user }, true));
+        await
+        dispatch(updateConnectionKeyPairs(wallet.mnemonic, wallet.privateKey, user.walletId, generateNewConnKeys));
```
jsidorenko: During the biometrics login, you won't get a mnemonic and private key

jsidorenko: It's better to wrap this call with `if(generateNewConnKeys)..` or you need to change the method's signature and make the first and second params optional

alazarevski: It is fine as long as the generation flow is wrapped in a condition with the last bool parameter 'generateNewConnKeys'

alazarevski: Added changes in f0ac452

PR #1110, opened 2019-05-02
fix/require PIN on Token Generation When State Changed

alazarevski:
Check if key need to be generated on login w/o biometrics

### DIFF
--- a/src/utils/keyPairGenerator.js
+++ b/src/utils/keyPairGenerator.js
@@ -68,10 +68,10 @@ const threadJobWorkerSeed = (
 ) => {
   return () => {
     return new Promise((resolve, reject) => {
-      let count = 20;
-      let lastCount = connectionsCount + (threadIndex * 20);
+      let count = 10;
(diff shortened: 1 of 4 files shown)


src/utils/keyPairGenerator.js
```diff
@@ -68,10 +68,10 @@ const threadJobWorkerSeed = (
 ) => {
   return () => {
     return new Promise((resolve, reject) => {
-      let count = 20;
-      let lastCount = connectionsCount + (threadIndex * 20);
+      let count = 10;
+      let lastCount = lastConnectionKeyIndex + (threadIndex * count);
       if (lastConnectionKeyIndex < 0) {
-        count = Math.ceil((100 + connectionsCount) / 5);
+        count = Math.ceil((50 + connectionsCount) / 5);
```
jsidorenko: It would be good to extract those magic numbers (10, 20, 50, 5) into some constants

PR #1113, opened 2019-05-03
Fix / Updating UI color and TextInput for iOS, Fix fatal crash on Lottie for Android

alazarevski:


src/screens/SendToken/SendTokenConfirm.js
```diff
@@ -152,6 +155,22 @@ class SendTokenContacts extends React.Component<Props, State> {
               }
             />
             }
+            {(!!recipientUsername && Platform.OS === 'ios') &&
+            <MultilineTextInput
```
jsidorenko: Is it possible to join `TextInput` and `MultilineTextInput ` components into one and put a platform condition inside? Then it will be easier to re-use

src/components/TextInput/TextInput.js
```diff
@@ -206,6 +208,20 @@ class TextInput extends React.Component<Props, State> {
     });
   };
 
+  handleMultilineFocus = () => {
+    if (!this.state.isFocused) {
+      this.rnInput.current.focus();
+      this.setState({
+        isFocused: false,
```
grilix: Just checking, isn't `state.isFocused` already `false`? Maybe this should set it to `true`?

src/components/TextInput/TextInput.js
```diff
@@ -206,6 +208,21 @@ class TextInput extends React.Component<Props, State> {
     });
   };
 
+  handleRNFocus = () => {
+    setTimeout(async () => {
+      this.multilineInputField._root.focus();
+      await this.setState({
```
jsidorenko: As I remember `setState()` doesn't return anything, so there is no need to write `await` before it

src/components/TextInput/TextInput.js
```diff
@@ -248,7 +266,7 @@ class TextInput extends React.Component<Props, State> {
             onChange={this.handleChange}
             onBlur={this.handleBlur}
             onEndEditing={() => this.handleBlur}
```
jsidorenko: This part looks weird, maybe we should return a function reference like on the line above, or we should actually call the handleBlur function (`onEndEditing={() => this.handleBlur()}`)
Could you check it pls?

src/components/TextInput/TextInput.js
```diff
@@ -258,10 +276,15 @@ class TextInput extends React.Component<Props, State> {
               paddingRight: (inputProps.multiline ? 58 : 14) + additionalRightPadding,
               paddingTop: inputProps.multiline ? 10 : 0,
               textAlignVertical: inputProps.multiline ? 'top' : 'center',
-              marginBottom: 10,
             }}
             onLayout={onLayout}
           />
+          <RNInput
```
jsidorenko: I'm a little bit confused now - will it render two input fields now (one below another)?

alazarevski: This hack focuses on a transparent input first which is a single line input and on ios it brings the keyboard avoiding scrollview up above the device keyboard. After the animation timing it focuses back on the multiline input.

PR #1125, opened 2019-05-10
Merge "Me" Screen into Feature/WalletConnect

pedrouid:


src/navigation/appNavigation.js
```diff
@@ -202,6 +206,13 @@ const assetsFlow = createStackNavigator(
 
 assetsFlow.navigationOptions = hideTabNavigatorOnChildView;
 
+// SELF FLOW
```
jsidorenko: Shall we rename it to `ME FLOW`? Only because we started to use that term everywhere

src/screens/Home/Home.js
```diff
@@ -337,22 +332,6 @@ class HomeScreen extends React.Component<Props, State> {
     return !isEq;
   }
 
-  validateWalletConnectQRCode = (uri: string) => {
```
jsidorenko: Are you sure all the deletions (according to this git diff) in this file should be deleted?

grilix: These are being moved into the `MeScreen`

src/navigation/appNavigation.js
grilix: Updated

PR #1130, opened 2019-05-14
Fix/sliding modal tab overlap

alazarevski:
- Fixed tab overlap in the Sliding modal
- 'More' icon removed from non-contacts details
- Fixed stuck spinner after wallet locks
- Added icons for muted/blocked connections


vyvil:
@alazarevski, the part removing "more" icon for non connections looks good for me :)
Although, I guess it would be better for me to sustain from approving this PR, given the fact that there are lots of changes made by me. Maybe @grilix could take a look? :)

PR #1132, opened 2019-05-15
fix/update-Sentry-Tokens-DSN

alazarevski:
Remove auth token from repo and change sentry dsn


antonioned:
@alazarevski I added changes to the build jobs to include the sentry token variable in the .env file.
Also change the naming of the hockeyapp jobs to be more specific. Please take a look before merging.

PR #1149, opened 2019-05-23
Fix/ghost invitations notifications

alazarevski:
Take all notifications from backend as invitations, keep TYPE_SENT in local storage only


src/actions/oldInvitationsActions.js
```diff
@@ -99,7 +99,11 @@ export const fetchOldInviteNotificationsAction = (theWalletId?: string = '') =>
       ...groupedNotifications.connectionRejectedEvent,
     ].map(({ id: userId }) => userId);
 
-    const updatedInvitations = uniqBy(latestEventPerId.concat(invitations), 'id')
+    const sentInvitations = invitations.filter((invi) => {
+      return invi.type === TYPE_SENT;
```
poocart: how about `const sentInvitations = invitations.filter(invi => invi.type === TYPE_SENT)`?

alazarevski: i'll add sentry logging too

PR #1151, opened 2019-05-24
Feature/firebase feature flags

poocart:
**Firebase Remote Config based feature flags**

1. Feature flags can be used for various purposes such as specific device debugging, canonical releases, A/B testing, etc.
2. In order to start using these flags we must add flag key to initial flags constants in the front-end and then add it to Firebase Remote Config to make it manageable. Publishing feature flag changes has logged history on Firebase so this can be reviewed.
3. Flags are retrieved and stored to reducer state, then cached for 1 hour in production and retrieved on each app open. Flags are not cached on development environment.
4. Our used flag format is `FEATURE_NAME_ENABLED: boolean`. 
5. Recommendations for implementing flags in the front-end code: https://martinfowler.com/articles/feature-toggles.html.


src/services/featureFlag.js
```diff
@@ -0,0 +1,37 @@
...
+export async function getRemoteFeatureFlags() {
+  const isDev = isTest || __DEV__;
+  if (isDev) firebase.config().enableDeveloperMode();
+  const cacheFor = isDev ? 0 : 3600; // 1 hour cache, 0 for non-cached results
+  const firebaseConfig = firebase.config();
+  await firebaseConfig.fetch(cacheFor).catch(() => null); // retrieve cached or fetch new if cache expired
+  await firebaseConfig.activateFetched();
+  const featureFlagKeys = Object.keys(INITIAL_FEATURE_FLAGS || {});
+  const fetchedFlags = await firebaseConfig.getValues(featureFlagKeys).catch(() => {});
+  return Object.keys(fetchedFlags).reduce((flags, flagKey) => {
```
jsidorenko: How about:
```
return Object.keys(fetchedFlags).reduce((flags, flagKey) => {
    return {
      ...flags,
      [flagKey]: !!fetchedFlags[flagKey].val(),
    };
  }, {});
```

src/actions/featureFlagsActions.js
```diff
@@ -0,0 +1,32 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import { getRemoteFeatureFlags } from 'services/featureFlag';
```
jsidorenko: I would keep the service name consistent, if you use `featureFlags` everywhere, then it's better to rename the service's file accordingly

src/services/featureFlag.js
```diff
@@ -0,0 +1,37 @@
...
+import { isTest } from 'utils/environment';
+import { INITIAL_FEATURE_FLAGS } from 'constants/featureFlagsConstants';
+
+export async function getRemoteFeatureFlags() {
+  const isDev = isTest || __DEV__;
+  if (isDev) firebase.config().enableDeveloperMode();
+  const cacheFor = isDev ? 0 : 3600; // 1 hour cache, 0 for non-cached results
+  const firebaseConfig = firebase.config();
+  await firebaseConfig.fetch(cacheFor).catch(() => null); // retrieve cached or fetch new if cache expired
+  await firebaseConfig.activateFetched();
```
jsidorenko:
```suggestion
  await firebaseConfig.activateFetched().catch(() => null);
```

src/actions/featureFlagsActions.js
poocart: 👍

src/services/featureFlag.js
poocart: 👍

src/services/featureFlag.js
poocart: 👍

PR #1152, opened 2019-05-24
Fix/166113222 update sliding modal

vyvil:
This PR includes changes on BottomSheet component and couple of related components based on Dmitry's and Drew's feedback.

Here's the list of changes applied to the components mentioned:
1. Add handlebars to indicate drag functionality;
2. Add an option to close the modal by tapping anywhere outside it;
3. Add backdrop;
4. Adjust listeners so tapping on a tab when modal is collapsed would result in opening modal AND changing to corresponding tab;
5. When CHAT modal is collapsed only last message should be visible. If last message is too long, it should be trimmed down to 2 lines of text. [Collapsed modal height should be "smart" and adapt to content inside];
6. Vertically center empty state message;
7. Adjust swiping gestures to match prototype;
8. Tapping on input field should open modal AND show keyboard;
9. Opening Chat should reveal the keyboard automatically;

And here's the card: [166113222](https://www.pivotaltracker.com/story/show/166113222)


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -392,20 +431,24 @@ class ActivityFeed extends React.Component<Props, State> {
       return true;
     });
 
-    const processedHistory = additionalFiltering ? additionalFiltering(filteredHistory) : filteredHistory;
+    this.feedData = additionalFiltering ? additionalFiltering(filteredHistory) : filteredHistory;
```
grilix: Mutating the component status on the `render` feels weird. Not sure what would be a good solution, but, isn't there a chance we can maybe move this whole part to its parent and pass the result as `props`? Probably the parent can filter the items itself, which would allow it to calculate the height, instead of having the flow go back and forth.

vyvil: Yeah, we could. Although, this component is used across multiple screens while holding almost the same information so it's connected to this component and are handled inside of it.

PR #1154, opened 2019-05-27
Allow scanning deep links on the Me screen

grilix:
This adds the support for handling deep links when scanning a QR code from the Add Session screen.

### DIFF
--- a/src/screens/Me/Me.js
+++ b/src/screens/Me/Me.js
@@ -28,6 +28,7 @@ import QRCodeScanner from 'components/QRCodeScanner';
 import CircleButton from 'components/CircleButton';
 import SettingsListItem from 'components/ListItem/SettingsItem';
 import { onWalletConnectSessionRequest } from 'actions/walletConnectActions';
+import { executeDeepLinkAction } from 'actions/deepLinkActions';
 import * as styled from './styles';
 
 const iconReceive = require('assets/icons/icon_receive.png');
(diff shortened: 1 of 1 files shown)


src/screens/Me/Me.js
```diff
@@ -51,18 +53,19 @@ class MeScreen extends React.Component<Props, State> {
   };
 
   validateWalletConnectQRCode = (uri: string) => {
```
poocart: Since now it validates both Wallet Connect and Pillarwallet deep links maybe we can rename `validateWalletConnectQRCode` as for example `validateScannedQRCode`, what do you think?

src/screens/Me/Me.js
```diff
@@ -51,18 +53,19 @@ class MeScreen extends React.Component<Props, State> {
...
   };
 
   toggleQRScanner = () => this.setState({ isScanning: !this.state.isScanning });
 
   handleQRScannerClose = () => this.setState({ isScanning: false });
 
   handleQRRead = (uri: string) => {
-    this.props.onWalletConnectSessionRequest(uri);
+    if (uri.startsWith('wc:')) {
+      this.props.onWalletConnectSessionRequest(uri);
```
poocart: Maybe destruct from props and then excecute the destructed fns?

grilix: Done, had to rename this one since it would shadow the imported function.

src/screens/Me/Me.js
grilix: Sounds good, changed.

PR #1157, opened 2019-05-30
Add fabric events

grilix:
This adds the following events:

- Wallet created
- Wallet imported
- User created
- Backup wallet phrase
- Token assets added
- Transaction sent from people tab
- Transaction sent from assets tab


src/actions/userActions.js
```diff
@@ -33,11 +34,16 @@ export const updateUserAction = (walletId: string, field: Object, callback?: Fun
         type: UPDATE_USER,
         payload: { user: updatedUser, state: REGISTERED },
       });
+      Answers.logCustom('Create User');
```
vyvil: `updateUserAction` is used to update user's profile information.
To track new created users `finishRegistration` or `registerOnBackendAction` (could be found in `onboardingActions.js`) might be more relevant :)
Or the message itself should be changed into a more suitable one. Like 'User info updated'.

src/actions/walletActions.js
```diff
@@ -160,6 +153,8 @@ export const generateWalletMnemonicAction = (mnemonicPhrase?: string) => {
         wordsToValidate,
       },
     });
+
+    Answers.logCustom('Create Wallet');
```
vyvil: Hm... In the list above I see that we should track "Wallet created" event.
On this step app just generates mnemonic. But wallet does not get registered anywhere yet.
Maybe to track wallets that have been created (not intended to be created) `getTokenWalletAndRegister` (could be found in `onboardingActions.js`) would be more accurate.

src/screens/SendToken/SendTokenPinConfirmScreen.js
```diff
@@ -58,9 +62,10 @@ class SendTokenPinConfirmScreen extends React.Component<Props, State> {
   handleTransaction = (pin: string, wallet: Object) => {
     const { sendAsset } = this.props;
     const { transactionPayload } = this.state;
-    this.setState({
-      isChecking: true,
-    }, () => sendAsset(transactionPayload, wallet, this.handleNavigationToTransactionState));
+    this.setState({ isChecking: true }, () => {
+      Answers.logCustom('Send Transaction', { Source: this.source });
```
vyvil: Also, for this - should we track transaction attempts or track only successful transactions?
At this point transactions can fail.

src/actions/userActions.js
grilix: You're right, I've changed this one to track the profile update, and added the create user on `registerOnBackendAction`.

src/actions/walletActions.js
grilix: Yeah, that's a better option, thanks! Updated.

src/screens/SendToken/SendTokenPinConfirmScreen.js
grilix: We wanted to have on one side the `Transaction sent` event, and in the other side the `Transaction success`/`Transaction failed`(I'm adding those in another PR), but now that you mention, I'm not sure if that makes sense or maybe we could use another name for this event. I'm gonna double check that.

Assuming we want to have the send and success/fail in separate events, do you think this would be the right place for this one?

vyvil: Yep, If you could also want to track transaction attempts - this method would suit you well :)

grilix: Thanks, then we're gonna have this event as a transaction start, with the source information. And then have a fail/success event.

PR #1162, opened 2019-06-07
Release candidate 07.06.2018

vyvil:
Includes:
1. Permissions screen for Android and reduced permissions list;
2. Wallet connect feature and initial part of Me section;
3. Updates on sliding modal (adjusted responses);
4. 64bit apk build support for Android;
5. Feature flags;


.env
```diff
@@ -14,4 +14,4 @@ OPEN_SEA_API_KEY=
 BADGES_CONTRACT_ADDRESS=0x79215cdE52B4E8e4d46Dd81f45e38b88d2290815
 INFURA_PROJECT_ID=
 SENTRY_AUTH_TOKEN=
-DEFAULT_PIN=
+DEFAULT_PIN=123456
```
jsidorenko: WTF? :)

vyvil: I guess Gonzalo mentioned this before.
If I'm not mistaken, it's a quick way to login on DEV mode - it enter this pin for you.
@grilix, am I right?

src/components/Header/Header.js
```diff
@@ -46,23 +46,34 @@ type Props = {
...
+`;
+
+const InnerWrapper = styled.View`
   justify-content: flex-end;
   align-items: flex-end;
   flex-direction: row;
   margin-top: ${spacing.rhythm};
-  margin-bottom: ${props => props.flexStart ? 'auto' : 0};
-  z-index: 10;
+  margin-bottom: ${props => props.flexStart ? 'auto' : 4};
```
jsidorenko: Shouldn't it be '4px'?

vyvil: Yeah, it would be safer to add unit. I'll fix this

.env
grilix: You are correct. It would be nice to have a way of adding custom environment variables on development without having to change this file, since it's a bit annoying having it as changed all the time 😞

PR #1174, opened 2019-06-11
Fix/wallet connect transaction fixes

poocart:
This PR includes:
- Added check if it's enough ether when confirming and display warning if false
- Added gas price and gas limit to all ERC20 transactions
- Added check if Wallet Connect ERC20 transaction request token is supported by Pillar Wallet and display warning if false
- Fixed incoming Wallet Connect transaction amount for ETH transactions
- Added default gas limit for Wallet Connect incoming request payload
- Fixed Wallet Connect ERC20 transaction approval and send


src/screens/WalletConnect/WalletConnectCallRequest.js
```diff
@@ -95,10 +109,18 @@ class WalletConnectCallRequestScreen extends React.Component<Props, State> {
       if (matchingAssets && matchingAssets.length) {
         asset = matchingAssets[0]; // eslint-disable-line
         symbol = asset.symbol; // eslint-disable-line
-        amount = utils
-          .bigNumberify(data.substring(73))
-          .mul(utils.bigNumberify('10').pow(asset.decimals));
+        const iface = new Interface(ERC20_CONTRACT_ABI);
```
jsidorenko: Pls check how all this work with tokens those have a non-standard amount of decimals (not 18)

src/screens/WalletConnect/WalletConnectCallRequest.js
```diff
@@ -316,13 +359,14 @@ class WalletConnectCallRequestScreen extends React.Component<Props, State> {
...
 const mapStateToProps = ({
-  assets: { supportedAssets },
+  assets: { supportedAssets, balances },
   contacts: { data: contacts },
   session: { data: session },
 }) => ({
   contacts,
   session,
   supportedAssets,
+  balances,
```
jsidorenko: When we merge this into smart-wallet branch, we need to use a selector to get the balance

poocart: 👍

src/screens/WalletConnect/WalletConnectCallRequest.js
grilix: That's a good catch. Do we have that information on the assets list?

grilix: Actually, the contract doesn't seem to include the decimals and then we're doing `amount = parseFloat(utils.formatUnits(methodValue, asset.decimals));`. Is there any example you can come up with where we could have a problem regarding the decimal points?

jsidorenko: Everything should be fine here, I just missed that part during review: `amount = parseFloat(utils.formatUnits(methodValue, asset.decimals));`

src/services/assets.js
```diff
@@ -62,21 +64,31 @@ function contractHasMethod(contractCode, encodedMethodName) {
...
+    gasLimit,
+    gasPrice,
   } = options;
   wallet.provider = getEthereumProvider(NETWORK_PROVIDER);
   const contract = new Contract(contractAddress, ERC20_CONTRACT_ABI, wallet);
-  if (decimals > 0) {
-    return contract.transfer(to, utils.parseUnits(amount.toString(), decimals), { nonce });
-  }
-  return contract.transfer(to, utils.bigNumberify(amount.toString()), { nonce });
+  const decimals = defaultDecimals > 0
```
alazarevski: for naming purposes, this should be parsedAmount right ?

src/screens/WalletConnect/WalletConnectCallRequest.js
poocart: Changed this to string as last decimals are being rounded when converted to JavaScript number.

src/services/assets.js
poocart: Thanks, changed it to `contractAmount`.

PR #1190, opened 2019-06-17
Add firebase analytics tracking for user properties

grilix:
Store `assets_layout` and `currency` as user properties on Firebase

### DIFF
--- a/src/actions/appSettingsActions.js
+++ b/src/actions/appSettingsActions.js
@@ -18,29 +18,40 @@
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 import { UPDATE_APP_SETTINGS } from 'constants/appSettingsConstants';
+import firebase from 'react-native-firebase';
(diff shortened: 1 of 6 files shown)


src/actions/appSettingsActions.js
```diff
@@ -18,29 +18,40 @@
...
     dispatch(saveDbAction('app_settings', { appSettings: settings }));
-    dispatch({
-      type: UPDATE_APP_SETTINGS,
-      payload: settings,
-    });
+    dispatch({ type: UPDATE_APP_SETTINGS, payload: settings });
+  };
+};
+
+export const updateAssetsLayoutAction = (layoutId: any) => {
```
jsidorenko: could you type it better?

grilix: Done.

grilix:
We're cancelling the migration to Firebase due to it's aggressive data collection. We're gonna stick to Fabric for now.

It will drop #1190, and require updating #1196.

PR #1192, opened 2019-06-18
now it’s possible to send a request back after rejecting it

andreysaleba:


src/actions/invitationsActions.js
```diff
@@ -237,9 +237,9 @@ export const rejectInvitationAction = (invitation: Object) => {
     } = getState();
 
     const rejectedInvitation = await api.rejectInvitation(
-      invitation.id,
-      invitation.sourceIdentityKey,
-      invitation.targetIdentityKey,
+      invitations[0].id,
+      invitations[0].sourceIdentityKey,
+      invitations[0].targetIdentityKey,
```
grilix: This sounds weird to me; we're gonna reject the first invitation? What happens if there are multiple invitations? Could you describe a little bit what the issue is and how this fixes it?

andreysaleba:
This bug on backend as I understand. Because now it works without those changes

PR #1199, opened 2019-06-20
fix/166546060_remove-sentry-info-restoreAccess-error

alazarevski:
Remove sentry message on an empty access token for a connection

### DIFF
--- a/src/actions/oldInvitationsActions.js
+++ b/src/actions/oldInvitationsActions.js
@@ -55,7 +55,6 @@ export const fetchOldInviteNotificationsAction = (theWalletId?: string = '') => 
     } = getState();
 
     if (accessTokens === undefined || !accessTokens.length) {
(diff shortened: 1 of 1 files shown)


src/actions/oldInvitationsActions.js
```diff
@@ -55,7 +55,6 @@ export const fetchOldInviteNotificationsAction = (theWalletId?: string = '') =>
     } = getState();
 
     if (accessTokens === undefined || !accessTokens.length) {
-      Sentry.captureMessage('Empty connection access tokens, dispatching restoreAccessTokensAction', { level: 'info' });
```
jsidorenko: Why do you think we don't need this?

alazarevski: this will almost always trigger because accessTokens are not used anymore, the condition is needed for old non-migrated connections yet

jsidorenko: 👍 thanks

PR #1200, opened 2019-06-20
Add Firebase notice on Permissions

grilix:
We need to let users know we're gonna track the activity using Firebase.


grilix:
We're not gonna collect user events with Firebase.

PR #1203, opened 2019-06-24
Send more events to Fabric

grilix:
## Current events

Action|Event name|Parameters
----|----|----
Opted out tracking|tracking_opted_out|
Opted back in tracking|tracking_opted_in|
Change assets layout|assets_layout_changed|
Add asset token|asset_token_added|symbol
Remove contact|contact_disconnected|
Mute contact|contact_muted|
Unmute contact|contact_unmuted|
Block contact|contact_blocked|
Unblock contact|contact_unblocked|
Add contact|connection_requested|
Accept connection|connection_accepted|
Reject connection|connection_rejected|
Cancel request|connection_cancelled|
Create wallet|wallet_created|
Create user|user_created|
Update profile|user_profile_updated|
Create onetime password|one_time_password_created|
Verify phone|phone_verified|
Update avatar|avatar_updated|
Import wallet|wallet_imported|method
Backup phrase|phrase_backed_up|
Kill Wallet Connect session|walletconnect_session_killed|
Disconnect Wallet Connect|walletconnect_disconnected|
Reject Wallet Connect request|walletconnect_rejected|
Request Wallet Connect|walletconnect_requested|
Wallet Connect request timed out|walletconnect_timed_out|
Wallet Connect session approved|walletconnect_connected|
First chat message to contact|chat_started|
Send transaction|transaction_sent|

## Opting out

The user can opt out tracking with a checkbox accessible via settings:

analytics-1-2

analytics-2-2 png


src/screens/Asset/Asset.js
```diff
@@ -189,9 +191,12 @@ class AssetScreen extends React.Component<Props, State> {
   };
 
   componentDidMount() {
-    const { fetchTransactionsHistory, navigation } = this.props;
+    const { fetchTransactionsHistory, navigation, logScreenView } = this.props;
     const { assetData, resetHideRemoval } = navigation.state.params;
     fetchTransactionsHistory(assetData.token);
+
+    logScreenView('asset', 'Asset');
```
vyvil: We could log a specific asset name here, not sure if it's needed, though.

src/screens/Assets/Assets.js
```diff
@@ -189,20 +190,22 @@ class AssetsScreen extends React.Component<Props, State> {
     const {
       fetchInitialAssets,
       assets,
+      navigation,
+      logScreenView,
     } = this.props;
 
-    Answers.logContentView('Assets screen');
+    logScreenView('assets', 'Assets');
```
vyvil: On Contact screen you are passing activeTab as a first parameter (view). This screen also has tabs (assets and collectibles), so I'm wondering maybe this data should be logged for this screen as well.

src/screens/Contact/Contact.js
```diff
@@ -283,6 +285,10 @@ class Contact extends React.Component<Props, State> {
   };
 
   setActiveTab = (activeTab) => {
+    const { logScreenView } = this.props;
+
+    logScreenView(activeTab, 'Contact');
```
vyvil: Just to check: This will be logged only when user presses on a tab. It won't log screen on its first render.

src/screens/Contact/ContactDetails.js
```diff
@@ -31,6 +31,7 @@ import {
   blockContactAction,
 } from 'actions/contactsActions';
 import { fetchContactTransactionsAction } from 'actions/historyActions';
+import { logScreenViewAction } from 'actions/analyticsActions';
```
vyvil: This screen is no longer in use - we could remove it

src/screens/Home/Home.js
```diff
@@ -341,9 +342,9 @@ class HomeScreen extends React.Component<Props, State> {
   };
 
   componentDidMount() {
-    const { fetchTransactionsHistory } = this.props;
+    const { fetchTransactionsHistory, logScreenView } = this.props;
 
-    Answers.logContentView('Home screen');
+    logScreenView('home', 'Home');
```
vyvil: Just to check: this screen also has tabs

src/screens/Profile/Profile.js
```diff
@@ -330,12 +365,37 @@ class Profile extends React.Component<Props, State> {
...
+        <SlideModal
+          isVisible={showTrackingModal}
+          fullScreen
+          showHeader
+          onModalHide={this.toggleTrackingModal}
+          backgroundColor={baseColors.lightGray}
+          avoidKeyboard
+        >
+          <Wrapper regularPadding flex={1}>
+            <SettingsModalTitle extraHorizontalSpacing>
```
vyvil: I believe we do not need extra horizontal spacing here - header gets shifted to right too much compared to other similar headers on this screen's modals. Unless its been decided otherwise :)

grilix: You're right, it looks weird. I've updated it and it looks a bit better now:

tracking-3

grilix:
@vyvil thanks for the comments, I've updated the PR. I've also added some initial tests, just to add a little bit of coverage.

src/screens/Contact/ContactDetails.js
vyvil: Thanks! :)

src/screens/Profile/Profile.js
vyvil: Great! :)

jsidorenko:
I agree we should tell the user that we don't gather any sensitive/private information, and all the data we're receiving is anonymized (if so)

grilix:
I agree, we've added the "opt out" checkbox there as a quick fix since that's something we're required to have, but we never defined exactly where it should be, how it should like like or what it should say. That checkbox was more important for Firebase, which collects a lot of user data and it would avoid all of that data collection (including the automatic collection done by the library).

In other hand, we already state on the privacy policy (that the user is required to read, and is then available on the settings screen) that we collect anonymous data for improving user experience.

The "Tracking" word is really ambiguous, but I have no idea of what else to put there. I've asked for the copy for these things and haven't received a reply. I'm open to any suggestion :grimacing:

PR #1204, opened 2019-06-25
modal hide listeners for forum login

poocart:
https://www.pivotaltracker.com/story/show/166871723

This issue was actually something from `react-native-modal` side, but it was happening only on iOS and it started when we merged QR scanner Camera modal to Home screen.

The problem was that modals can't be shown if one is not fully closed. In this case it was happening when QR scanner Camera modal is not yet fully closed and confirm login modal is already set to be visible. Same issue appeared for add email modal so this also includes fix for it More about the issue – https://github.com/react-native-community/react-native-modal#i-cant-show-multiple-modals-one-after-another.


jsidorenko:
You rock ;)

PR #1206, opened 2019-06-25
Release Candidate 25.06.2019

poocart:
This includes:
- Forum login issue on iOS (nothing was happening after scanning login QR code).
- Smart Wallet feature which is off by default and can be controlled by feature flag using Firebase Remote Config.
- Exchange feature which is off by default and can be controlled by feature flag using Firebase Remote Config.
- Corrupted database storage fix.


alazarevski:
would add #1207

vyvil:
> would add #1207

Added :)

PR #1212, opened 2019-06-26
Feature/show badges on assets screen

vyvil:
This PR includes:
1. User badges list moved to Assets screen (to Collectibles tab);
2. Added Pillar SDK method to fetch badges owned by contact (increased SDK version to 1.30.1558);
3. Added UI components to show badges owned by user contacts on Contact screen (hidden till we allow users to manage badges visibility).

More info on this task and it's re-scope could be found [here](https://www.pivotaltracker.com/story/show/166697394).


src/screens/Contact/Contact.js
```diff
@@ -446,30 +510,58 @@ class Contact extends React.Component<Props, State> {
...
+                  />
+                }
+                onPress={this.toggleBadgesSection}
+                open={isBadgesSectionOpen}
+                onAnimationEnd={
+                  isBadgesSectionOpen
+                  ? () => { this.scroll.scrollToEnd(); }
+                  : () => {}
+                }
+              /> */}
```
grilix: Do we want to keep this? There are some other commented blocks that I'd remove if they are not required.

vyvil: I'd leave it - it's, as I've wrote in the description, is commented out till we allow users to manage badge visibility

PR #1221, opened 2019-06-28
Refactor token price calculation

grilix:
Pivotal cad: https://www.pivotaltracker.com/n/projects/2223806/stories/164056062

We want to calculate the token values using ethereum value as reference, instead of relying on token price directly. This changes the `getRate` to do the calculation taking that into account.

### DIFF
--- a/src/utils/assets.js
+++ b/src/utils/assets.js
@@ -34,7 +34,32 @@ export function getBalance(balances: Balances = {}, asset: string = ''): number 
 }
 
 export function getRate(rates: Rates = {}, token: string, fiatCurrency: string): number {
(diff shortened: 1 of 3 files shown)


src/utils/assets.js
```diff
@@ -34,7 +34,32 @@ export function getBalance(balances: Balances = {}, asset: string = ''): number
 }
 
 export function getRate(rates: Rates = {}, token: string, fiatCurrency: string): number {
-  return rates[token] && rates[token][fiatCurrency] ? Number(rates[token][fiatCurrency]) : 0;
+  const rate = rates[token];
```
jsidorenko:
```suggestion
  const tokenRates = rates[token];
```

PR #1222, opened 2019-07-01
Fix/166952203 messages after push

vyvil:
This PR includes:
1. Updated conditions when to fetch new messages (to fetch them when chat bottom sheet is forced opened);
2. Updated conditions when to show new message toast.


src/actions/notificationsActions.js
```diff
@@ -412,12 +416,18 @@ export const startListeningChatWebSocketAction = () => {
               type: SET_UNREAD_CHAT_NOTIFICATIONS_STATUS,
               payload: true,
             });
-            if (!!navParams.contact && !!navParams.contact.username
-              && navParams.contact.username === senderUsername && navParams.chatTabOpen) {
+            if (get(navParams, 'contact.username', '') === senderUsername && navParams.chatTabOpen) {
```
jsidorenko: what if we join that block with the block below?
`const contact = get(navParams, 'contact.username', '') || get(navParams, 'username', '');`

grilix: Not sure if I follow, @jsidorenko, but it seems like joining those will add some complexity, since we use `contact = navParams.contact` or `contact = contacts.find(...)`; I think it's not really that straighforward :thinking:

jsidorenko: You have 2 very similar blocks, that only difference is how you get/extract the username: `get(navParams, 'contact.username', '')` vs. `get(navParams, 'username', '')`
Probably we could rewrite this in this way:
```javascript
const contactUsername = get(navParams, 'contact.username', '') || get(navParams, 'username', '');
            if (contactUsername === senderUsername && !!navParams.chatTabOpen) {
              const contact = contacts.find(c => c.username === contactUsername) || {};
              dispatch(getChatByContactAction(contact.username, contact.id, contact.profileImage));
              return;
            }
```

grilix: Yes, the other difference is how we extract the contact, but we could do `contact = contacts.find(...)` for both cases :+1:

vyvil: Yeah, this is mainly the reason why I did not combine them in the first place. I see what I can do about it :)

vyvil: Merged into one like this: 
```suggestion
const senderNameInNavParams = get(navParams, 'contact.username', '') || (get(navParams, 'username', ''));
const relatedContact = get(navParams, 'contact', null) || contacts.find(c => c.username === navParams.username) || {};

if (senderNameInNavParams === senderUsername && !!navParams.chatTabOpen && !!Object.keys(relatedContact).length) {
 const { username, id, profileImage } = relatedContact;
 dispatch(getChatByContactAction(username, id, profileImage));
 return;
}
```
What do you guys think?

jsidorenko: Looks good!
I would also change `contacts.find(c => c.username === navParams.username)` to `contacts.find(c => c.username === senderNameInNavParams)`

vyvil: Ahh, missed that one. Good catch. Thanks! :) I'll update this and will merge that in

PR #1225, opened 2019-07-01
Update imports on ContactInfo tests

grilix:
Fix ContactInfo tests on case sensitive filesystems.

Error message:

test-error

### DIFF
--- a/src/actions/__tests__/authActions.test.js
+++ b/src/actions/__tests__/authActions.test.js
@@ -98,7 +98,7 @@ describe('Wallet actions', () => {
       });
   });
 
-  it('Should expect a different set of actions for registered users.', () => {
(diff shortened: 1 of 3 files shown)


src/actions/__tests__/authActions.test.js
```diff
@@ -98,7 +98,7 @@ describe('Wallet actions', () => {
       });
   });
 
-  it('Should expect a different set of actions for registered users.', () => {
+  it('Should expect a different set of actions for registered users.', async () => {
```
grilix: This one fails randomly and I'm not able to consistently reproduce it. I've added the `async` to let the tests know this will not be run instantly, but not sure about that. Let me know what you people think.

PR #1227, opened 2019-07-02
Fix/exchange fixes and improvements

poocart:
PR includes these exchange fixes/improvements:
- Fix for fiat display amount.
- Formatting large decimals on available and rate numbers.
- Added set token allowance flow.
- New Exchange settings screen.
- Updated Exchange confirm screen.
- Other UI fixes/improvements.


src/screens/Exchange/ExchangeInfo.js
```diff
@@ -0,0 +1,266 @@
...
+      return 'Uniswap';
+    case PROVIDER_ZEROX:
+      return '0x';
+    case PROVIDER_CHANGELLY:
+      return 'Changelly';
+    default:
+      return 'Unknown';
+  }
+};
+const dummyExchanges = [
```
jsidorenko: Shall we rename this?

vyvil: It's just a placeholder info for styling. I've already removed in my following commit

src/actions/historyActions.js
```diff
@@ -37,16 +37,31 @@ import { checkForMissedAssetsAction } from './assetsActions';
 import { saveDbAction } from './dbActions';
 import { getExistingTxNotesAction } from './txNoteActions';
 import { checkAssetTransferTransactionsAction } from './smartWalletActions';
+import { checkEnableExchangeAllowanceTransactionsAction } from './exchangeActions';
 import { ETH } from '../constants/assetsConstants';
 
 const TRANSACTIONS_HISTORY_STEP = 10;
 
+export const afterHistoryUpdatedAction = () => {
```
grilix: This looks like some utility more than an action. Just as a suggestion, I wouldn't export it and probably wouldn't even call it "action", but not a big deal.

src/screens/Exchange/ExchangeConfirm.js
```diff
@@ -115,6 +118,16 @@ class ExchangeConfirmScreen extends React.Component<Props, State> {
     transactionSpeed: NORMAL,
   };
 
+  componentDidUpdate() {
+    const {
+      executingExchangeTransaction,
+      navigation,
+    } = this.props;
+    if (!executingExchangeTransaction) {
+      navigation.goBack();
```
grilix: How could we end up on this screen if we're not supposed to be here? Can we handle this somewhere else?

poocart: This is because we're navigation to asset send flow and we want to dismiss this screen once back from that flow and transaction succeeded (we dispatch exchange transaction reset on succeed).

src/actions/historyActions.js
poocart: We want to use latest and updated getState in this action after dispatch, removed the export part.

src/screens/Exchange/ExchangeStatus.js
```diff
@@ -0,0 +1,67 @@
...
+  margin-top: 2px;
+`;
+
+const IconHolder = styled.View`
+  position: relative;
+`;
+
+const animationSource = require('assets/animations/livePulsatingAnimation.json');
+
+export const ExchangeStatus = () => {
```
grilix: Are we using this?

vyvil: Yes.
Here: https://github.com/pillarwallet/pillarwallet/pull/1227/files/e5eaf0bf9d8c592cecda95aa2d44a48541adbc7f#diff-5eb6ff88b9f7b09c6c5388ef4cfb5818R925
and here: https://github.com/pillarwallet/pillarwallet/pull/1227/files/e5eaf0bf9d8c592cecda95aa2d44a48541adbc7f#diff-5eb6ff88b9f7b09c6c5388ef4cfb5818R931
(both in Exchange.js)

grilix: Oh, totally missed that. I was wondering about the hardcoded "ACTIVE" status here, but it seems like we are already handling that.

PR #1228, opened 2019-07-02
[WIP] faceid_permission tried to add prompt for faceid auth permission

andreysaleba:


src/screens/Profile/Profile.js
```diff
@@ -215,6 +215,25 @@ class Profile extends React.Component<Props, State> {
...
+      return TouchID.isSupported()
+        .then(biometryType => {
+          if (biometryType === 'FaceID') {
+            FaceId.request();
+            this.setBiometricState(value);
+          } else if (biometryType === 'TouchID') {
+            this.setBiometricState(value);
+          } else if (biometryType === true) {
+            this.setBiometricState(value);
+          }
```
grilix: If `biometryType` is none of the listed types, `this.setBiometricState(value);` will never be called, is that right? I'm curious because if it fails, it will be called.

grilix:
This is not complete and we're not working on this right now. Let's close this PR for now.

PR #1236, opened 2019-07-08
fix/update_exchange_search

vyvil:
Tiny PR to update selector option search to find assets by the name.


poocart:
Maybe it's also worth adding asset name to asset list card?

vyvil:
> Maybe it's also worth adding asset name to asset list card?

yup, on it. Thanks! :)

PR #1249, opened 2019-07-09
Add yarn clean command

grilix:
Add a shortcut command for cleaning up the environment.


package.json
```diff
@@ -47,6 +47,7 @@
     "run-s": "^0.0.0"
   },
   "scripts": {
+    "clean": "rm -rf node_modules ios/build ios/Pods",
```
poocart: please also remove ios/Podfile.lock

jsidorenko: What about windows users?

grilix: @jsidorenko not sure of what we can do about Windows; we already use unix commands on this same file.

We could have two different entries, like `clean`, `clean-windows` (or something like that); or we could have a script that detects the system and do as appropriate, but I don't know if any of those is really worth it.

What do you think? Any suggestion?

jsidorenko: @grilix check that library: https://www.npmjs.com/package/rimraf

grilix: Updated

jsidorenko: 👍 And where are the changes of yarn.lock?

grilix: @jsidorenko no changes on `yarn.lock`, we already have it: https://github.com/pillarwallet/pillarwallet/blob/7e725f7785cdbabd59a1574ee907432ccf2ff437/yarn.lock#L11453

jsidorenko: Cool, we can merge it then

PR #1256, opened 2019-07-10
Fix icons on activity list

grilix:
- Show collectible icons
- Show asset icons when address is not on contacts

## Activity list

activity-images

### DIFF
--- a/src/actions/collectiblesActions.js
+++ b/src/actions/collectiblesActions.js
@@ -46,19 +46,23 @@ export const fetchCollectiblesAction = () => {
         asset_contract: assetContract,
         name,
         description,
-        image_preview_url: image,
+        image_url: fullImage,
+        image_preview_url: preview,
         image_thumbnail_url: thumbnail,
       } = collectible;
       const { name: category, address: contractAddress } = assetContract;
       const collectibleName = name || `${category} ${id}`;
 
+      const image = (/\.(png|gif)$/i).test(fullImage) ? fullImage : '';
+      const previewImage = (/\.(png|gif)$/i).test(preview) ? preview : '';
+
       return {
         id,
         category,
+        image: image || previewImage,
         name: collectibleName,
         description,
-        thumbnail,
-        icon: (/\.(png|gif)$/i).test(image) ? image : '',
+        icon: thumbnail || previewImage || image,
         contractAddress,
         assetContract: category,
         tokenType: COLLECTIBLES,
--- a/src/components/ActivityFeed/ActivityFeed.js
+++ b/src/components/ActivityFeed/ActivityFeed.js
@@ -50,6 +50,7 @@ import { COLLECTIBLE_TRANSACTION } from 'constants/collectiblesConstants';
 import { TRANSACTION_EVENT, CONNECTION_EVENT } from 'constants/historyConstants';
 import { CONTACT } from 'constants/navigationConstants';
 import { CHAT } from 'constants/chatConstants';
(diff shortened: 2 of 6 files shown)


src/actions/collectiblesActions.js
```diff
@@ -46,19 +46,22 @@ export const fetchCollectiblesAction = () => {
...
         description,
-        image_preview_url: image,
+        image_url: fullImage,
+        image_preview_url: preview,
         image_thumbnail_url: thumbnail,
       } = collectible;
       const { name: category, address: contractAddress } = assetContract;
       const collectibleName = name || `${category} ${id}`;
 
+      const image = (/\.(png|gif)$/i).test(fullImage) ? fullImage : '';
```
jsidorenko: Shall we test `preview` for extensions as well? The problem here is that some images are stored on ipfs and we have no way to display them

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -223,16 +227,21 @@ class ActivityFeed extends React.Component<Props, State> {
         directionSymbol = '';
       }
 
+      const fullIconUrl = iconUrl ? `${SDK_PROVIDER}/${iconUrl}?size=3` : '';
+
       const contact = contacts
         .find(({ ethAddress }) => address.toUpperCase() === ethAddress.toUpperCase()) || {};
+      const inContacts = Object.keys(contact).length !== 0;
```
jsidorenko:
```suggestion
      const isContact = Object.keys(contact).length !== 0;
```

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -223,16 +227,21 @@ class ActivityFeed extends React.Component<Props, State> {
...
         <ListItemWithImage
           onPress={() => this.selectEvent({ ...notification, value, contact }, type, notification.status)}
           label={nameOrAddress}
-          avatarUrl={contact.profileImage}
-          navigateToProfile={Object.keys(contact).length !== 0 ? navigateToContact : null}
-          iconName={Object.keys(contact).length === 0 || showArrowsOnly ? directionIcon : null}
+          avatarUrl={itemImage}
+          navigateToProfile={inContacts ? navigateToContact : null}
+          iconName={(showArrowsOnly || !itemImage) ? directionIcon : null}
+          imageAddonIconName={(itemImage && !showArrowsOnly) ? directionIcon : undefined}
```
jsidorenko: Shouldn't it be `null` and not `undefined`?

src/actions/collectiblesActions.js
grilix: Oh, okay, we can test that as well. Will update

src/components/ActivityFeed/ActivityFeed.js
grilix: With `null` it triggers a warning for incompatible types. I just copied how we do it on other places. We could update the `ListItemWithImage` component to allow `imageAddonIconName=null`, but  I think it's not on the scope of this fix. What do you think?

jsidorenko: No, it's fine with undefined :)

src/components/ActivityFeed/ActivityFeed.js
grilix: Updated

src/actions/collectiblesActions.js
grilix: Updated

PR #1258, opened 2019-07-12
Feature/set exchange min max button

poocart:
This includes:
- Removal of offers filter for min and max amounts.
- Added button on offer card to set min or max amount if the current amount is below min or above max amount.
- Fix for availability display when min is not set and max is higher than 0.01.

### DIFF
--- a/src/screens/Exchange/Exchange.js
+++ b/src/screens/Exchange/Exchange.js
@@ -189,8 +189,12 @@ const getAvailable = (_min, _max, rate) => {
   let max = (new BigNumber(rate)).multipliedBy(_max);
   if ((min.gte(0) && min.lt(0.01)) || (max.gte(0) && max.lt(0.01))) {
     if (max.isZero()) return '>0.01';
+    const maxAvailable = max.lt(0.01)
+      ? '<0.01'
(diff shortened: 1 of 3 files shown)


src/screens/Exchange/Exchange.js
```diff
@@ -533,6 +552,23 @@ class ExchangeScreen extends React.Component<Props, State> {
     });
   };
 
+  setFromAmount = amount => {
+    this.props.resetOffers();
+    this.setState({
```
jsidorenko: it's not safe to use this.state.value, I would consider converting into this way:
```
this.setState(prevState) => {
  return {
value: {
        ...prevState.value,
        fromInput: {
          ...prevState.value.fromInput,
          input: amount,
        },
      }};
}, () => { ...
```

poocart: Thanks! 👍 But then if `prevState` is used then state updated callback is not acceptable, these values in state are always present so I think it's safe to spread state like this and pass callback as we need to validate the input right after state was updated. I will look for another option how to pass validation and perform search, but it can be we will need to leave it this way.

poocart: Thanks, updated 👍

PR #1263, opened 2019-07-17
Show collectible background on collectible list

grilix:
I've seen some collectibles have a background color which can be used on the collectible details and listings. Some services use that background and I think it would be a step forward into having a more familiar place for holding our collectibles.

This is a working approach for the collectibles listing, but I think it would be nice to have it on the collectible details page as well. Any help/suggestion on that would be appreciated.

> **NOTE**: Not yet tested on iOS.

## Previous

collectibles-1

## Updated

collectibles-2


src/components/AssetCard/AssetCardMinimized.js
```diff
@@ -272,21 +273,31 @@ class AssetCardMinimized extends React.Component<Props, State> {
       balanceInFiat,
       isCollectible,
       name,
+      backgroundColor,
     } = this.props;
+    const background = backgroundColor ? `#${backgroundColor}` : null;
```
poocart: Not important, but this is nicer:
```
const background = backgroundColor && `#${backgroundColor};
```

src/screens/Assets/CollectiblesList.js
```diff
@@ -127,6 +127,7 @@ class CollectiblesList extends React.Component<Props> {
     const { navigation } = this.props;
     return (
       <BadgeTouchableItem
+        background={`#${item.backgroundColor}`}
```
poocart: Correct me if I'm wrong, but `backgroundColor` here is now always present as you also check for the same in in `AssetCardMinimized.js`?

PR #1266, opened 2019-07-18
Add WIDGET_SIGNATURE env var. to circle

antonioned:


antonioned:
@alazarevski this is set up the same way we handle OPEN_SEA_API_KEY for example... so if we use the same OPEN_SEA value for prod and staging builds at the moment, same will apply for WIDGET_SIGNATURE env var.

antonioned:
let me add those changes @alazarevski

antonioned:
@alazarevski done

PR #1267, opened 2019-07-18
Feature/Fiat Exchange with SendWyre and MoonPay Integration

alazarevski:
WebView Widgets Integration
Added exchange for USD, EUR, GBP to crypto buying options


src/screens/Home/Home.js
```diff
@@ -756,6 +756,12 @@ class HomeScreen extends React.Component<Props, State> {
                 onPress={() => navigation.navigate(MANAGE_DETAILS_SESSIONS)}
                 wrapperPaddingHorizontal={spacing.mediumLarge}
               />
+              {/* <SettingsListItem */}
```
poocart: Probably was used for debug purpose? Maybe not needed anymore?

alazarevski: Updated in d685df4

src/actions/exchangeActions.js
```diff
@@ -167,6 +167,19 @@ export const searchOffersAction = (fromAssetCode: string, toAssetCode: string, f
     );
     // we're requesting although it will start delivering when connection is established
     const { error } = await exchangeService.requestOffers(fromAssetCode, toAssetCode);
+
+    await api.fetchMoonPayOffers(fromAssetCode, toAssetCode, fromAmount).then((offer) => {
```
jsidorenko: Combining of .then() with await is not a good practise (only if you don't transform the data).
I would use the same approach as everywhere: `const moonPayOffers = await ...`

jsidorenko: The exception would be if you don't need to actually need to wait for the response in this flow, then consider removing `await`

alazarevski: This caught my attention too, wouldn't mind even to make it truly async as it does not depend on the other results anyway. Would it be ok like that ?

src/screens/Exchange/Exchange.js
```diff
@@ -716,13 +778,36 @@ class ExchangeScreen extends React.Component<Props, State> {
   generateAssetsOptions = (assets) => {
     const { balances, paymentNetworkBalances } = this.props;
     const assetsList = Object.keys(assets).map((key: string) => assets[key]);
-    const nonEmptyAssets = assetsList.filter((asset: any) => {
-      return getBalance(balances, asset.symbol) !== 0 || asset.symbol === ETH;
+    const fiatCurrencies = [
```
jsidorenko: Maybe it's better to extract this to some config file?

src/screens/Exchange/Exchange.js
```diff
@@ -716,13 +778,36 @@ class ExchangeScreen extends React.Component<Props, State> {
...
+      {
+        symbol: EUR,
+        decimals: 2,
+        iconUrl: 'asset/images/fiat/eurColor.png',
+        iconMonoUrl: 'asset/images/fiat/eurColor.png',
+      },
+    ];
+    const cryptoFiatAssets = assetsList.concat(fiatCurrencies);
+    const nonEmptyAssets = cryptoFiatAssets.filter((asset: any) => {
+      return getBalance(balances, asset.symbol) !== 0 || [ETH, USD, defaultFiatCurrency, EUR].includes(asset.symbol);
```
jsidorenko: I would extract this `[ETH, USD, defaultFiatCurrency, EUR].includes(asset.symbol)` as well, you're using it in a lot of places

src/screens/FiatExchange/FiatCrypto.js
```diff
@@ -0,0 +1,216 @@
...
+            options: { $set: optionsFrom },
+          },
+        },
+      },
+    });
+
+    this.setState({ options: newOptions });
+  };
+
+  componentDidMount = () => {
```
jsidorenko: Minor: pls move this method above the `handleFormChange` as life-cycle methods normally go first

src/screens/FiatExchange/FiatExchange.js
```diff
@@ -0,0 +1,137 @@
...
+        provider,
+        fromAssetCode: sourceCurrency,
+        toAssetCode: destCurrency,
+        amount: sourceAmount,
+      },
+    } = navigation.state.params;
+
+    const { email = '' } = user;
+
+    const destAddress = getActiveAccountAddress(accounts);
```
jsidorenko: Maybe it's better to add all those preparation bellow to componentDidMount()?
I mean `const destAddress`, `const secretKey`, `const wyreTemplate`

src/services/api.js
```diff
@@ -743,3 +746,63 @@ SDKWrapper.prototype.importedErc20TransactionHistory = function (walletAddress:
...
     .catch(() => []);
 };
+
+SDKWrapper.prototype.fetchMoonPayOffers = function (fromAsset: string, toAsset: string, amount: number) {
+  const url = `${MOONPAY_API_URL}/v2/currencies/${toAsset.toLowerCase()}/quote/?apiKey=${MOONPAY_KEY}`
+  + `&baseCurrencyAmount=${amount}&baseCurrencyCode=${fromAsset.toLowerCase()}`;
+
+  return Promise.resolve()
+    .then(() => fetch(url))
+    .then(async (data) => {
```
jsidorenko:
```
.then(data => data.json())
.then(data => ....
```

src/services/api.js
```diff
@@ -743,3 +746,63 @@ SDKWrapper.prototype.importedErc20TransactionHistory = function (walletAddress:
...
+        };
+      }
+      return { error: true };
+    })
+    .catch(() => ({ error: true }));
+};
+
+SDKWrapper.prototype.fetchSendWyreOffers = function (fromAsset: string, toAsset: string, amount: number) {
+  return Promise.resolve()
+    .then(async () => {
```
jsidorenko: Do you really need async here?

src/services/api.js
```diff
@@ -743,3 +746,63 @@ SDKWrapper.prototype.importedErc20TransactionHistory = function (walletAddress:
...
+      return fetch(`${SENDWYRE_API_URL}/v3/rates?as=MULTIPLIER`);
+    })
+    .then(async (resp) => {
+      const data = await resp.json();
```
jsidorenko: Same as above

src/screens/Exchange/Exchange.js
```diff
@@ -691,12 +730,34 @@ class ExchangeScreen extends React.Component<Props, State> {
             </CardInnerRow>
           </CardRow>
           <CardRow>
+            {!!isFiat &&
+            <CardColumn style={{ flex: 1 }}>
+              <CardText label>Fees Total</CardText>
+              <View style={{ flexDirection: 'row' }}>
+                <CardText>{feeAmount !== '' ? `${feeAmount + extraFeeAmount} ${fromAssetCode}` : 'Check'}</CardText>
```
poocart: if no fees then what "Check" in fees means?

poocart: got fee 5.39000........1 from eur to eth on MoonPay, maybe we can format it using `formatAmountDisplay`?

alazarevski: Couldn't get the sendwyre api to work and give a quote, just the rates between the pairs. Check on Payment there is a screen in the widget of the price breakdown. Will have to speak with UI how to handle it better.

alazarevski: For fiat currency we can go to 2 decimals, it would be enough

src/screens/Exchange/Exchange.js
```diff
@@ -660,10 +691,18 @@ class ExchangeScreen extends React.Component<Props, State> {
           onPress={() => this.onOfferPress(offer)}
```
poocart: shouldn't it check which type is clicked and then trigger `onFiatOfferPress` if it's fiat?

src/screens/Exchange/Exchange.js
```diff
@@ -691,12 +730,34 @@ class ExchangeScreen extends React.Component<Props, State> {
...
               <CardText label>Available</CardText>
               <View style={{ flexDirection: 'row' }}>
                 <CardText>{available}</CardText>
               </View>
             </CardColumn>
+            }
+            {!!isFiat &&
+            <CardColumn>
+              <Button
+                title={isTakeOfferPressed ? '' : `${quoteCurrencyAmount} ${toAssetCode}`}
```
poocart: format the value as there might be many decimals? use the formatting methods for other exchange cards

poocart: you can use `formatAmountDisplay`

poocart:
please check `componentDidMount` and `setInitialSelection`, it will fail if fiat was selected, app gone to background and presented again after PIN.

src/actions/exchangeActions.js
alazarevski: updated in 571513a

src/services/api.js
alazarevski: 571513a

src/services/api.js
alazarevski: 571513a

src/services/api.js
alazarevski: 571513a

src/screens/Exchange/Exchange.js
poocart: So maybe then if it's unable to get fees then don't show this part?

poocart: needs formatting, but agree on 2 decimals only 👍

src/screens/Exchange/Exchange.js
alazarevski: 0291e7e, except this line because it had ETH check on the previous version

src/screens/Exchange/Exchange.js
alazarevski: 0291e7e

src/screens/Exchange/Exchange.js
```diff
@@ -650,6 +679,8 @@ class ExchangeScreen extends React.Component<Props, State> {
       || !allowanceSet
       || (isShapeShift && !shapeshiftAccessToken);
 
+    const isFiat = offerProvider === PROVIDER_MOONPAY || offerProvider === PROVIDER_SENDWYRE;
```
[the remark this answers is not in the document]
alazarevski: 0291e7e

src/screens/Exchange/Exchange.js
alazarevski: I'd let it stay just because it will imply fees and the user should have the heads up for that

src/screens/FiatExchange/FiatExchange.js
alazarevski: d154ef1

src/screens/FiatExchange/FiatCrypto.js
alazarevski: d154ef1

src/screens/Exchange/Exchange.js
alazarevski: 927faae

src/screens/Exchange/Exchange.js
alazarevski: 927faae

src/screens/Exchange/Exchange.js
alazarevski: 927faae

alazarevski:
> please check `componentDidMount` and `setInitialSelection`, it will fail if:
> 
> 1. fiat was selected
> 2. app gone to background
> 3. app presented again after PIN.

2079c72 fixes it

src/utils/exchange.js
```diff
@@ -55,7 +66,34 @@ export const getProviderLogo = (provider?: string) => {
...
+    case PROVIDER_MOONPAY:
+      return moonPayLogo;
+    case PROVIDER_SENDWYRE:
+      return sendWyreLogo;
     default:
       return '';
   }
 };
+
+export const checkFiatProvider = (provider: string) => {
```
grilix: The word "check" is ambiguous, I'd rather name this something like `isFiatProvider`, for example.

src/utils/exchange.js
```diff
@@ -55,7 +66,34 @@ export const getProviderLogo = (provider?: string) => {
...
+  switch (provider) {
+    case PROVIDER_MOONPAY:
+    case PROVIDER_SENDWYRE:
+      return true;
+    default:
+      return false;
+  }
+};
+
+export const checkFiatCurrency = (symbol: string) => {
```
grilix: Same here, I'd avoid "check" on names, we could use something like `isFiatCurrency` instead.

src/screens/Exchange/Exchange.js
```diff
@@ -455,8 +463,11 @@ class ExchangeScreen extends React.Component<Props, State> {
 
   setInitialSelection = (fromAssetCode: string, toAssetCode?: string, fromAmount?: number) => {
     const { assets, supportedAssets } = this.props;
+    const fromAsset = checkFiatCurrency(fromAssetCode) ?
+      fiatCurrencies.find(currency => currency.symbol === fromAssetCode) :
+      assets[fromAssetCode];
```
grilix: Small detail: Since we search for the symbol on the `fiatCurrencies` as a way of knowing if a symbol is fiat, I'd avoid doing it twice:

```js
const fromAsset =  fiatCurrencies.find(
  currency => currency.symbol === fromAssetCode
) || assets[fromAssetCode];
```

src/utils/exchange.js
poocart: Updated in last commit 👍

src/utils/exchange.js
poocart: Updated in last commit 👍

src/screens/Exchange/Exchange.js
poocart: Updated in last commit 👍

PR #1271, opened 2019-07-23
Feature/transaction gas estimate calc improvements

poocart:
https://www.pivotaltracker.com/story/show/167369941

Includes:
- Updated gas limit calculation method to rely on `ether.js` gas estimate method. This affects regular, collectible and Wallet Connect transactions.
- Wallet Connect call request screen back button handler fix.
- Wallet Connect call request screen UI content wrapper background fix.
- Exchange confirm screen update in regards of updated gas calculation method (`ethers.js` approach was already there before this PR).
- Regular transaction speed select option is hidden until you enter any amount – this is because we cannot calculate proper gas limit if there is no amount and therefore cannot show speed select with estimate prices for each speed.

Missing:
- Smart Wallet gas estimate calculations not in this PR as it will be separate task and will be also dependent on Smart Wallet SDK methods.

**P. S. This also might cover other Pivotal tickets that are related to long transactions or rejected transactions due unknown reasons.**


src/screens/SendToken/SendTokenAmount.js
```diff
@@ -175,23 +178,24 @@ class SendTokenAmount extends React.Component<Props, State> {
     });
   };
 
-  handleChange = (value: Object) => {
-    this.setState({ value });
+  handleChange = async (value: Object) => {
+    const gasLimit = await this.getGasLimit();
```
jsidorenko: We need to debounce this

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -358,12 +394,14 @@ const mapStateToProps = ({
   rates: { data: rates },
   history: { gasInfo },
   appSettings: { data: { baseFiatCurrency, transactionSpeed } },
+  wallet: { data: { address: fromAddress } },
```
jsidorenko: We have active account address selector for that

src/screens/WalletConnect/WalletConnectCallRequest.js
```diff
@@ -416,11 +441,13 @@ const mapStateToProps = ({
   contacts: { data: contacts },
   session: { data: session },
   history: { gasInfo },
+  wallet: { data: { address: fromAddress } },
```
jsidorenko: Consider using selector here

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -151,7 +154,10 @@ class SendTokenAmount extends React.Component<Props, State> {
     this.state = {
       value: null,
       showModal: false,
+      gasLimit: 0,
     };
+
+    this.handleChange = debounce(this.handleChange, 500);
```
jsidorenko: You need to debounce the `getGasLimit()` method, not the handleChange

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -175,23 +181,24 @@ class SendTokenAmount extends React.Component<Props, State> {
     });
   };
 
-  handleChange = (value: Object) => {
-    this.setState({ value });
+  handleChange = async (value: Object) => {
+    const gasLimit = await this.getGasLimit();
```
jsidorenko: Consider this approach:
```
handleChange = (value: Object) => {
  this.setState({ value });
  this.getGasLimit().then(gasLimit => this.setState({ gasLimit }));
};
```

jsidorenko: You need to call `this.setState({ value });` immediately in order to prevent having a lag once you type

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -175,23 +181,24 @@ class SendTokenAmount extends React.Component<Props, State> {
...
   handleFormSubmit = () => {
     this.formSubmitted = true;
     const txFeeInWei = this.getTxFeeInWei();
     const value = this._form.getValue();
     const { navigation } = this.props;
+    const { gasLimit } = this.state;
     const transactionSpeed = this.getTxSpeed();
-    const gasPrice = txFeeInWei.div(GAS_LIMIT).toNumber();
-
+    const gasPrice = txFeeInWei.div(gasLimit).toNumber();
```
jsidorenko: If the user is offline or there're some problems fetching the gas - the result will be `Infinity` as you're trying to divide by 0

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -207,26 +214,58 @@ class SendTokenAmount extends React.Component<Props, State> {
...
+      gasLimit,
+      value: { amount },
     });
   };
 
-  getTxFeeInWei = (txSpeed?: string): BigNumber => {
+  getGasLimit = (amount?: number) => {
+    // calculate either with amount in form or provided as param
+    if (!amount) {
+      ({ amount = 0 } = this._form.getValue() || {});
```
jsidorenko: with lodash.get it could look more readable: `amount = get(this._form.getValue(), 'amount', 0);`

jsidorenko: P. S. keep in mind that the form's value will be a string and the line below won't work: `if (amount === 0) return 0;`

src/screens/SendToken/SendTokenAmount.js
poocart: If user is offline or any problems occur then it it will return our default 500k gas limit.

poocart: services/assets.js:

```
return provider.estimateGas({
    from,
    to,
    data,
    value,
  })
    ...
    .catch(() => DEFAULT_GAS_LIMIT);
```

src/screens/SendToken/SendTokenAmount.js
poocart: Fixed in newer commits

src/screens/SendToken/SendTokenAmount.js
poocart: Fixed in newer commits 👍

src/screens/SendToken/SendTokenAmount.js
poocart: Fixed in newer commits 👍

PR #1275, opened 2019-07-24
Feature/exchange fiat selector horizontal layout

poocart:
Includes improvements to make Exchange fiat select horizontal as it is in People search.

### DIFF
--- a/src/components/SelectorInput/SelectorInput.js
+++ b/src/components/SelectorInput/SelectorInput.js
@@ -382,16 +382,23 @@ export default class SelectorInput extends React.Component<Props, State> {
         || (name && name.toUpperCase().includes(query.toUpperCase())))
       : options;
 
+    const filteredHorizontalListData = (query && query.length >= MIN_QUERY_LENGTH && horizontalOptions.length)
+      ? horizontalOptions.filter(({ value: val, name }) => (val && val.toUpperCase().includes(query.toUpperCase()))
+        || (name && name.toUpperCase().includes(query.toUpperCase())))
+      : horizontalOptions;
+
+    const selectorOptionsCount = options.length + horizontalOptions.length;
+
     return (
       <React.Fragment>
         <Wrapper style={wrapperStyle}>
           {!!label && <Label>{label}</Label>}
           <ItemHolder error={!!errorMessage}>
-            {!!options.length &&
+            {!!selectorOptionsCount &&
             <Selector
               fullWidth={!hasInput}
-              onPress={options.length > 1 ? this.openSelector : noop}
-              disabled={options.length < 1}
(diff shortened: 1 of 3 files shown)


src/components/SelectorInput/SelectorInput.js
```diff
@@ -295,16 +389,23 @@ export default class SelectorInput extends React.Component<Props, State> {
         || (name && name.toUpperCase().includes(query.toUpperCase())))
       : options;
 
+    const filteredHorizontalListData = (query && query.length >= MIN_QUERY_LENGTH && horizontalOptions.length)
```
jsidorenko: You're becoming master of complex and unreadable code :D

src/screens/Exchange/Exchange.js
```diff
@@ -811,6 +817,21 @@ class ExchangeScreen extends React.Component<Props, State> {
     });
   };
 
+  generateFiatOptions = () => {
+    return fiatCurrencies.map(({ symbol, iconUrl, ...rest }) => {
+      return ({
```
jsidorenko: No need to wrap with `()` here

src/components/SelectorInput/SelectorInput.js
jsidorenko:
```
let filteredHorizontalListData = horizontalOptions;
if (query && query.length >= MIN_QUERY_LENGTH && horizontalOptions.length) {
  filteredHorizontalListData = horizontalOptions.filter(({ value: val, name }) => {
    val = val ? val.toUpperCase() : '';
    name = name ? name.toUpperCase() : '';
    const searchQuery = query.toUpperCase();
    return val.includes(searchQuery) || name.includes(searchQuery);
  });
}
```

poocart: Thanks, improved in new commit 👍

src/screens/Exchange/Exchange.js
poocart: Thanks, improved in new commit 👍

src/components/SelectorInput/SelectorInput.js
```diff
@@ -386,17 +386,19 @@ export default class SelectorInput extends React.Component<Props, State> {
...
-        || !isSearchQuery,
-    );
+    const isSearchQuery = query && query.length >= MIN_QUERY_LENGTH;
+
+    let filteredListData = options;
+    let filteredHorizontalListData = horizontalOptions;
+
+    if (isSearchQuery) {
+      filteredListData = filteredListData.filter(
+        ({ value: val, name }) => isMatchingSearch(query, val) || isMatchingSearch(query, name),
```
jsidorenko: Perfecto!

poocart: ¡muchas gracias Señor!

PR #1276, opened 2019-07-24
feature/Limit Offers By IP

alazarevski:
ip geolocation limit to services and offers

### DIFF
--- a/src/actions/exchangeActions.js
+++ b/src/actions/exchangeActions.js
@@ -153,17 +153,23 @@ export const searchOffersAction = (fromAssetCode: string, toAssetCode: string, f
     // we're requesting although it will start delivering when connection is established
     const { error } = await exchangeService.requestOffers(fromAssetCode, toAssetCode);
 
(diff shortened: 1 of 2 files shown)


poocart:
How about getting IP on Exchange screen componentDidMount and checking for network change to get IP again rather than checking IP on every form change?

src/actions/exchangeActions.js
```diff
@@ -153,17 +153,23 @@ export const searchOffersAction = (fromAssetCode: string, toAssetCode: string, f
...
-      }
-    }).catch(() => null);
+    const ipInfo = await api.getLocationByIP();
 
-    api.fetchSendWyreOffers(fromAssetCode, toAssetCode, fromAmount).then((offer) => {
-      if (!offer.error) {
-        dispatch({ type: ADD_OFFER, payload: offer });
-      }
-    }).catch(() => null);
+    if (ipInfo.isAllowed) {
```
poocart: How about wrapping is allowed for both and then checking US in another level?

alazarevski:
Closing because different approach is needed.

PR #1279, opened 2019-07-24
bugFix/improve fiat cards

alazarevski:
Add label for country support warning when IP location is determined

### DIFF
--- a/src/actions/exchangeActions.js
+++ b/src/actions/exchangeActions.js
@@ -157,21 +157,25 @@ export const searchOffersAction = (fromAssetCode: string, toAssetCode: string, f
 
     api.fetchMoonPayOffers(fromAssetCode, toAssetCode, fromAmount).then((offer) => {
       if (!offer.error) {
-        const modoffer = {
-          ...offer,
-          offerRestricted: !isAllowed ? `Unavailable in ${alpha2}` : null,
-        };
-        dispatch({ type: ADD_OFFER, payload: modoffer });
+        dispatch({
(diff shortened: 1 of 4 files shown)


src/actions/exchangeActions.js
```diff
@@ -153,15 +153,23 @@ export const searchOffersAction = (fromAssetCode: string, toAssetCode: string, f
     // we're requesting although it will start delivering when connection is established
     const { error } = await exchangeService.requestOffers(fromAssetCode, toAssetCode);
 
+    const { isAllowed = false, alpha2 = '' } = await exchangeService.getIPInformation();
+
     api.fetchMoonPayOffers(fromAssetCode, toAssetCode, fromAmount).then((offer) => {
       if (!offer.error) {
-        dispatch({ type: ADD_OFFER, payload: offer });
+        const modoffer = { ...offer, isAllowed, alpha2 };
```
poocart: How about removing `alpha2` and changing `isAllowed` to `offerRestricted: string` and if it is present the we display whole message without need to pass.

Also add it to Offer model.

src/screens/Exchange/Exchange.js
```diff
@@ -730,6 +732,11 @@ class ExchangeScreen extends React.Component<Props, State> {
                 </ButtonLabel>
               </CardButton>
               }
+              {!!isFiat && !isAllowed &&
```
poocart: I think this `CardButton` should be disabled, as well other button with amount should be also applied with logic to make it hidden.

PR #1295, opened 2019-07-30
Pillar Payment Network improvements

jsidorenko:
This PR includes fixes for those flows/parts:
- the key based wallet migration to the smart wallet flow
- Pillar Tank top up flow
- Send tokens within the Pillar Payment Network
- improves notifications
- adds method to import PPN transactions history
- displays 'pending' balances
- displays transactions available to settle

If it will be impossible to review, we can split this PR into parts


poocart:
1. Home screen warning after creating new wallet.
2. Funding tank multiple times in a row while first fund is not yet finished gives result of first fund in a row successful and others failed (probably not front-ent related).
3. "Upgrade" button is not shown after wallet imported from Private Key, reloading app solves the issue.

PR #1307, opened 2019-08-12
Feature/smart wallet account recovery

poocart:
This covers one piece of Smart Wallet Account Recovery: the setup (both during migration with button that allows to skip or manual setup through Smart Wallet options).

Additional notes:
- Required agents number is yet to be added to UI so now it uses number of agents that are added.
- Enable and (or) setup is blocked if Smart Wallet is not deployed.
- Gas fees for enable and (or) setup transactions are not shown as it is yet to be added to UI.
- This PR may contain other cleanup/fixes.


src/reducers/accountRecoveryReducer.js
```diff
@@ -0,0 +1,83 @@
...
+  enabled: false,
+  agents: [],
+  requiredAgentsCount: 0,
+  transactions: [],
+};
+
+export default function accountRecoveryReducer(
+  state: AccountRecoveryReducerState = initialState,
+  action: AccountRecoveryReducerAction,
+) {
```
grilix: We could add the returned type as well.

poocart: Thanks, missed it 👍

poocart:
WIP: needs UI.

PR #1312, opened 2019-08-13
Refactor Login request flow and QR Scanner

grilix:
Make the login request view a new screen instead of a modal, which
simplyfies the code and avoids weird border scenarios where a modal
could prevent another modal from showing up.

Login requests are passed directly to the Home/Login screen, instead of
being passed to Home/Home.

The rework on the QRScanner makes its less complex and the interactions
easier to follow. It now exposes just `onRead` and `onCancel`.

onRead: The scanner will close itself and trigger onRead when it's fully
dismissed. This way the caller doesn't have to worry about waiting for
the modal to be closed.

onCancel: When the user wants to cancel. This event is triggered when
the camera is still open. The caller needs to close the scanner.


src/navigation/appNavigation.js
```diff
@@ -285,12 +287,12 @@ const exchangeFlow = createStackNavigator({
...
 
 // ME FLOW
-const meFlow = createStackNavigator({
-  [ME]: MeScreen,
-  [MANAGE_DETAILS_SESSIONS]: ManageDetailsSessionsScreen,
-}, StackNavigatorConfig);
-
-meFlow.navigationOptions = hideTabNavigatorOnChildView;
+// const meFlow = createStackNavigator({
+//   [ME]: MeScreen,
```
ricardoul: commentend lines?

src/navigation/appNavigation.js
```diff
@@ -122,7 +123,7 @@ import { removePrivateKeyFromMemoryAction } from 'actions/walletActions';
 // constants
 import {
   ADD_TOKEN,
-  ME,
+  // ME,
```
ricardoul: also commented?

src/navigation/appNavigation.js
```diff
@@ -53,7 +53,8 @@ import SendTokenTransactionScreen from 'screens/SendToken/SendTokenTransaction';
 import SendCollectibleConfirmScreen from 'screens/SendCollectible/SendCollectibleConfirm';
 import PPNSendTokenAmountScreen from 'screens/Tank/SendToken/PPNSendTokenAmount';
 import HomeScreen from 'screens/Home';
-import MeScreen from 'screens/Me';
+import LoginScreen from 'screens/Home/Login';
+// import MeScreen from 'screens/Me';
```
ricardoul: also commented?

grilix: Yes, we need to decide if we're gonna drop that view or if we plan to link it from somewhere. Currently we are not using it, so I commented it out just to make it explicit. I could remove it but not sure :thinking:

src/actions/deepLinkActions.js
```diff
@@ -43,45 +78,12 @@ export const executeDeepLinkAction = (deepLink: string) => {
...
-                params: {
-                  showLoginApproveModal: true,
-                },
-              }),
-            });
-            navigate(navigateToAppAction);
-          }
-        }
+        const {
+          query = { loginToken: '' },
```
poocart: I'm sorry, I think default here is useless, you can just check in `beginApproveLogin` if loginToken is present or not.

grilix: Yes, the thing is that the `loginToken` field is mandatory on the `ApproveLoginQuery` type, since I thought it wouldn't make sense to call `beginApproveLogin` without a token.

I think you're right, tho, we could make that field optional and let `beginApproveLogin` handle the rest.

PR #1315, opened 2019-08-13
Feature/let users apply for beta testing

poocart:
PR includes:
1. Firebase Analytics added back again with disabled data collection by default (upon login or during onboarding). It is used to pass username to Firebase if user wants to join beta program.
2. App settings option for users to let them join beta or leave it.

Notes:
1. Both joining and leaving beta actions needs full app reload to take effect if feature flag is enabled for user in Firebase Remote Config (this is not automated).


src/screens/Profile/Profile.js
```diff
@@ -779,11 +793,21 @@ class Profile extends React.Component<Props, State> {
...
             </ListSeparator>
 
+            <ProfileSettingsItem
+              key="joinBeta"
+              label={userJoinedBeta ? 'Leave Beta Testing' : 'Join Beta Testing'}
+              onPress={() => userJoinedBeta
+                ? setUserJoinedBeta(false)
+                : this.setState({ showJoinBetaModal: true })
+              }
+            />
```
grilix: What do you think about having two different branches instead of checking the same condition over and over? i.e., something like:

```jsx
{userJoinedBeta ? (
  <ProfileSettingsItem
    key="joinBeta"
    label="Leave Beta Testing"
    onPress={() => setUserJoinedBeta(false)}
   />
) : (
  <ProfileSettingsItem
    key="joinBeta"
    label="Join Beta Testing"
    onPress={() => this.setState({ showJoinBetaModal: true })}
   />
)}
```

Even tho it results on more lines, I think it looks cleaner and more straightforward.

jsidorenko: I would go with this approach:
```
{userJoinedBeta ? (
  <ProfileSettingsItem
    key="joinBeta"
    label="Leave Beta Testing"
    onPress={() => setUserJoinedBeta(false)}
   />
)}
{!userJoinedBeta ? (
  <ProfileSettingsItem
    key="joinBeta"
    label="Join Beta Testing"
    onPress={() => this.setState({ showJoinBetaModal: true })}
   />
)}
```
we should try to avoid `else` part in render methods as it creates complexity

src/screens/Profile/Profile.js
```diff
@@ -805,6 +829,38 @@ class Profile extends React.Component<Props, State> {
...
+                    'your wallet experience for any bugs and/or crashes while testing the new functionality. ' +
+                    'You can opt out of the beta program and Firebase Analytics collection at any time ' +
+                    'via the "System" under Settings.'
+                  }
+                </Description>
+                <Button
+                  title="Join Beta Testing"
+                  onPress={() => {
+                    setUserJoinedBeta(true);
+                    this.setState({ showJoinBetaModal: false });
```
grilix: I'd move this to a single function call.

poocart: Had to refactor this a bit as success toast was appearing in modal which closed right after. Please take a look.

src/screens/Profile/Profile.js
poocart: Thanks 👍

PR #1319, opened 2019-08-15
Feature/ppn settlement

jsidorenko:


src/actions/accountsActions.js
```diff
@@ -256,5 +257,12 @@ export const initOnLoginSmartWalletAccountAction = (privateKey: string) => {
     await dispatch(connectSmartWalletAccountAction(activeAccountId));
     dispatch(fetchVirtualAccountBalanceAction());
     dispatch(syncVirtualAccountTransactionsAction());
+
+    if (blockchainNetwork === BLOCKCHAIN_NETWORK_TYPES.PILLAR_NETWORK) {
```
grilix: Why not do this always?

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -360,6 +390,18 @@ class ActivityFeed extends React.Component<Props, State> {
     } = this.state;
 
     const feedList = getSortedFeedData(tabs, activeTab, feedData);
+    const feedSections = [];
+
+    feedList.forEach((listItem) => {
+      const formattedDate = formatDate(new Date(listItem.createdAt * 1000), 'MMM D');
+      const existingSection = feedSections.find(({ title }) => title === formattedDate) || {};
```
grilix: Do we really need the default object here? I think it would be cleaner to just use the undefined return from `find`:

```js
const existingSection = feedSections.find(({ title }) => title === formattedDate);
if (existingSection) {
  existingSection.data.push(listItem);
} else {
  feedSections.push({ title: formattedDate, data: [{ ...listItem }] });
}
```

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -360,6 +390,18 @@ class ActivityFeed extends React.Component<Props, State> {
...
+      if (!Object.keys(existingSection).length) {
+        feedSections.push({ title: formattedDate, data: [{ ...listItem }] });
+      } else {
+        existingSection.data.push(listItem);
```
grilix: We are making a copy on one of the branches and using the value on the other one. I'd say we stay consistent with either option:
```js
feedSections.push({ title: formattedDate, data: [{ ...listItem }] });
existingSection.data.push({ ...listItem });
// or
feedSections.push({ title: formattedDate, data: [ listItem ] });
existingSection.data.push(listItem);
```

PR #1330, opened 2019-08-16
Add initial tests for Asset screen

grilix:
This sets up a basic test for the Asset screen

### DIFF
--- a/src/models/Transaction.js
+++ b/src/models/Transaction.js
@@ -104,4 +104,5 @@ export type SmartWalletTransferTransaction = {
   hash: string,
   value: Object,
   asset: string,
+  status: string,
 };
--- a/src/utils/smartWallet.js
+++ b/src/utils/smartWallet.js
@@ -23,10 +23,15 @@ import { TX_CONFIRMED_STATUS } from 'constants/historyConstants';
 
 import type { Accounts } from 'models/Account';
 import type { SmartWalletStatus } from 'models/SmartWalletStatus';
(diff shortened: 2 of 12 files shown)


src/models/Transaction.js
```diff
@@ -104,4 +104,5 @@ export type SmartWalletTransferTransaction = {
   hash: string,
   value: Object,
   asset: string,
+  status: string,
```
jsidorenko: Why do you need this? For PPN transactions we don't have status, as they're getting confirmed immediately

src/utils/smartWallet.js
```diff
@@ -23,10 +23,15 @@ import { TX_CONFIRMED_STATUS } from 'constants/historyConstants';
 
 import type { Accounts } from 'models/Account';
 import type { SmartWalletStatus } from 'models/SmartWalletStatus';
+import type { WalletReducerState } from 'reducers/smartWalletReducer';
```
jsidorenko: I would rename the type to `SmartWalletReducerState`

src/models/Transaction.js
grilix: After typing, flow complains on this line: https://github.com/pillarwallet/pillarwallet/blob/42e49d3fcd3800e67dcea137125633fe784d2773/src/utils/smartWallet.js#L53

image

Any suggestion?

jsidorenko: Cool, then leave the `status` field in typing

src/utils/smartWallet.js
grilix: Updated

PR #1332, opened 2019-08-19
Fix/167971554

ricardoul:


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -228,7 +228,13 @@ class ActivityFeed extends React.Component<Props, State> {
         decimals = 18,
         iconUrl,
       } = assets.find(({ symbol }) => symbol === notification.asset) || {};
-      const value = utils.formatUnits(new BigNumber(notification.value.toString()).toFixed(), decimals);
+      let bigNumber;
```
jsidorenko: that variable name means nothing

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -228,7 +228,13 @@ class ActivityFeed extends React.Component<Props, State> {
...
+      try {
+        bigNumber = new BigNumber(notification.value.toString());
```
jsidorenko: `bigNumber = new BigNumber(get(notification, 'value', 0).toString());`
and no need for try..catch

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -441,6 +447,7 @@ class ActivityFeed extends React.Component<Props, State> {
           stickySectionHeadersEnabled={false}
         />
 
+        {selectedEventData &&
```
jsidorenko: should be `{!!selectedEventData &&`

src/components/ActivityFeed/ActivityFeed.js
grilix: That would handle the `undefined` scenario, right? What do you think about the decimals? Is that something we need to take care of?

jsidorenko: I wouldn't care about decimals here

grilix:
This was fixed on another PR. Extracted the tests to #1341

PR #1334, opened 2019-08-20
bugFix/BigNumber formatting

alazarevski:
BigNumber parsing error catch

### DIFF
--- a/src/utils/common.js
+++ b/src/utils/common.js
@@ -304,6 +304,7 @@ export function formatUnits(val: string = '0', decimals: number) {
     Sentry.captureMessage(e.message, {
       level: 'info',
       extra: {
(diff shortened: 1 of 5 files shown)


src/utils/common.js
```diff
@@ -294,3 +295,19 @@ export function getGasPriceWei(gasInfo: GasInfo): BigNumber {
...
+
+export function formatUnits(val: string = '0', decimals: number) {
+  let formattedUnits = '0.0';
+  try {
+    formattedUnits = utils.formatUnits(new BigNumber(val.toString()).toFixed(), decimals);
+  } catch (e) {
+    Sentry.captureMessage(e.message, {
+      level: 'info',
+      extra: {
+        inputValue: val,
```
jsidorenko: add some note into extra, like note: 'formatUnits error'

PR #1337, opened 2019-08-20
wait for sw upgrade asset transaction to be mined

poocart:
This adds listener for Smart Wallet upgrade asset transfer transaction hash status.


jsidorenko:
Cherry picked into `feature/ppn-demo`

PR #1344, opened 2019-08-21
Feature/move wallet list to accounts list

vyvil:
PR to move wallets list to Accounts screen


src/screens/Accounts/Accounts.js
```diff
@@ -69,11 +76,12 @@ type Props = {
   accounts: Accounts,
   resetIncorrectPassword: Function,
   switchAccount: Function,
+  smartWalletState: Object,
```
grilix: Would this be `reducers/smartWalletReducer#WalletReducerState`? If so, we could explicitly type it accordingly.

PR #1358, opened 2019-08-26
Improve support for multiple networks on Accounts

grilix:
This is an effort to simplify the Accounts screen in order to have an
easier way of adding netorks.

Now the Accounts screen will have explicit types for accounts, networks
and account creation items, which will help adding different networks
like BTC.

### DIFF
--- a/src/actions/blockchainNetworkActions.js
+++ b/src/actions/blockchainNetworkActions.js
@@ -17,21 +17,22 @@
     with this program; if not, write to the Free Software Foundation, Inc.,
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
-
 import { SET_ACTIVE_NETWORK } from 'constants/blockchainNetworkConstants';
-import { UPDATE_APP_SETTINGS } from 'constants/appSettingsConstants';
-import { saveDbAction } from './dbActions';
+
+import type { Dispatch, GetState } from 'reducers/rootReducer';
+import { updateAppSettingsAction } from 'actions/appSettingsActions';
 
 export const setActiveBlockchainNetworkAction = (id: string) => {
-  return async (dispatch: Function) => {
-    dispatch({
-      type: SET_ACTIVE_NETWORK,
-      payload: id,
-    });
-    dispatch(saveDbAction('app_settings', { appSettings: { blockchainNetwork: id } }));
-    dispatch({
-      type: UPDATE_APP_SETTINGS,
-      payload: { blockchainNetwork: id },
-    });
+  return async (dispatch: Dispatch, getState: GetState) => {
+    const { blockchainNetwork: { data: networks } } = getState();
+    const network = networks.find(({ id: networkId }) => networkId === id);
+
+    if (!(network && network.isAvailable)) {
+      console.error('Trying to activate an unavailable network'); // eslint-disable-line no-console
+      return;
+    }
+
+    dispatch({ type: SET_ACTIVE_NETWORK, payload: id });
+    dispatch(updateAppSettingsAction('blockchainNetwork', id));
   };
 };
--- a/src/actions/dbActions.js
+++ b/src/actions/dbActions.js
@@ -20,16 +20,20 @@
 import Storage from 'services/storage';
 import { UPDATE_DB } from 'constants/dbConstants';
 
+import type { DbAction } from 'models/DbAction';
+
 const storage = Storage.getInstance('db');
 
-export const saveDbAction = (key: string, data: any, forceRewrite: boolean = false) => {
-  return {
-    queue: 'db',
-    type: UPDATE_DB,
-    callback: (next: Function) => {
-      storage.save(key, data, forceRewrite).then(() => {
-        next(); // eslint-disable-line
-      }).catch(() => {});
-    },
-  };
-};
+export const saveDbAction = (
+  key: string,
+  data: any,
+  forceRewrite: boolean = false,
+): DbAction => ({
+  type: UPDATE_DB,
+  queue: 'db',
+  callback: (next: () => void) => {
+    storage.save(key, data, forceRewrite)
+      .then(() => next()) // eslint-disable-line
+      .catch(() => {});
+  },
+});
--- a/src/models/Account.js
+++ b/src/models/Account.js
@@ -11,7 +11,6 @@ export type Account = {
   extra?: AccountExtra,
   isActive: boolean,
   walletId: string,
-  privateKey?: any,
 };
 
 export type Accounts = Account[];
--- a/src/models/Account.js
+++ b/src/models/Account.js
@@ -11,6 +11,7 @@ export type Account = {
   extra?: AccountExtra,
   isActive: boolean,
   walletId: string,
+  privateKey?: any,
 };
 
 export type Accounts = Account[];
--- a/src/reducers/appSettingsReducer.js
+++ b/src/reducers/appSettingsReducer.js
@@ -24,12 +24,12 @@ import merge from 'lodash.merge';
 export type AppSettingsReducerState = {
   data: Object,
   isFetched: boolean,
-}
+};
 
 export type AppSettingsReducerAction = {
   type: string,
-  payload: any
-}
+  payload: Object,
+};
 
 export const initialState = {
   data: {
@@ -42,10 +42,10 @@ export const initialState = {
   isFetched: false,
 };
 
-export default function appSettingsReducer(
+const appSettingsReducer = (
   state: AppSettingsReducerState = initialState,
   action: AppSettingsReducerAction,
-) {
+): AppSettingsReducerState => {
   switch (action.type) {
     case UPDATE_APP_SETTINGS:
       const updatedState = { data: action.payload, isFetched: true };
@@ -57,4 +57,6 @@ export default function appSettingsReducer(
     default:
       return state;
   }
-}
+};
+
+export default appSettingsReducer;
--- a/src/reducers/blockchainNetworkReducer.js
+++ b/src/reducers/blockchainNetworkReducer.js
@@ -1,44 +1,71 @@
 // @flow
+/*
+    Pillar Wallet: the personal data locker
+    Copyright (C) 2019 Stiftung Pillar Project
+
+    This program is free software; you can redistribute it and/or modify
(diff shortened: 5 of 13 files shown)


src/actions/blockchainNetworkActions.js
```diff
@@ -18,20 +18,25 @@
...
+      return;
+    }
+
+    dispatch(activateNetwork(id));
     dispatch(saveDbAction('app_settings', { appSettings: { blockchainNetwork: id } }));
-    dispatch({
-      type: UPDATE_APP_SETTINGS,
-      payload: { blockchainNetwork: id },
-    });
+    dispatch(updateAppSettings({ blockchainNetwork: id }));
```
jsidorenko: How about adding `dispatch(saveDbAction('app_settings', { appSettings: .....}));` into `updateAppSettings` action?

src/reducers/blockchainNetworkReducer.js
```diff
@@ -19,26 +26,40 @@ const initialState = {
...
       title: 'Pillar network',
       isActive: false,
+      isAvailable: true,
     },
   ],
 };
 
-export default function blockchainNetworkReducer(
+
+export const activateNetwork = (id: string): BlockchainNetworkAction => ({
```
jsidorenko: I think this should be put into actions file

src/reducers/blockchainNetworkReducer.js
```diff
@@ -4,13 +4,20 @@ import {
...
+export type BlockchainNetwork = {
+  id: string,
+  title: string,
+  isActive: boolean,
+  isAvailable: boolean,
+};
+
 export type BlockchainNetworkReducerState = {
-  data: Object[],
+  data: BlockchainNetwork[],
```
jsidorenko: 👍

src/reducers/appSettingsReducer.js
```diff
@@ -42,6 +42,11 @@ export const initialState = {
   isFetched: false,
 };
 
+export const updateAppSettings = (settings: Object): AppSettingsReducerAction => ({
```
jsidorenko: Shouldn't this be stored in the actions file?

src/actions/dbActions.js
```diff
@@ -20,16 +20,20 @@
 import Storage from 'services/storage';
 import { UPDATE_DB } from 'constants/dbConstants';
 
+import type { DbAction } from 'reducers/rootReducer';
```
jsidorenko: I would put this either near the implementation or in models

src/screens/Accounts/Accounts.js
```diff
@@ -214,7 +220,7 @@ class AccountsScreen extends React.Component<Props, State> {
     }
   };
 
-  switchToSmartWalletAccount = async (_: string, wallet: Object) => {
+  switchToSmartWalletAccount = async (_: string, wallet: Account) => {
```
jsidorenko: 👍

src/screens/Accounts/Accounts.js
```diff
@@ -224,69 +230,80 @@ class AccountsScreen extends React.Component<Props, State> {
...
-            borderColor: activeBNetworkID === BLOCKCHAIN_NETWORK_TYPES.PILLAR_NETWORK
-              ? baseColors.electricBlue
-              : baseColors.white,
-            borderRadius: 6,
-          }}
-        >
-          {activeBNetworkID === BLOCKCHAIN_NETWORK_TYPES.PILLAR_NETWORK && <CheckIcon name="check" />}
-        </ListCard>
-      );
+      return this.renderPPN(network, isActive);
```
vyvil: Considering that cards for Networks will have the same layout, we could simply reuse `renderPPN` method (renamed, of course) to render all other networks. Checking network type to call render method can result in duplicate code for card rendering :)

src/screens/Accounts/Accounts.js
```diff
@@ -295,40 +312,114 @@ class AccountsScreen extends React.Component<Props, State> {
         )}
       />
     );
+  }
+
+  renderInitializeAccount(item: NewItem) {
```
vyvil: This seems like almost identical to renderAccount.
We could save some lines by checking one prop to alter the look and action for init card in renderAccount method

jsidorenko:
image
Got this after I switched to your branch

src/screens/Accounts/Accounts.js
```diff
@@ -295,40 +312,114 @@ class AccountsScreen extends React.Component<Props, State> {
...
+        { type: 'NEW_SMART_WALLET', action: () => navigation.navigate(SMART_WALLET_INTRO) },
       ]
       : visibleAccountsWithBalance;
 
+    const ppnNetwork = blockchainNetworks.find((network) => network.id === BLOCKCHAIN_NETWORK_TYPES.PILLAR_NETWORK);
+
+    if (smartWalletFeatureEnabled && ppnNetwork) {
+      return [
+        ...wallets,
+        { type: 'NETWORK', network: ppnNetwork },
```
vyvil: Hm, why we need ppnNetwork listed in this.wallets? 
And then included once again in line 443?
If we are keeping ppnNetwork in this.wallets, maybe lets rename the prop to accounts?

src/utils/smartWallet.js
```diff
@@ -54,25 +54,34 @@ const getMessage = (
...
       };
     default:
       return {};
   }
 };
 
-export function getSmartWalletStatus(accounts: Accounts, smartWalletState: Object): SmartWalletStatus {
-  const account = accounts.find(acc => acc.type === ACCOUNT_TYPES.SMART_WALLET);
-  const activeAccount = getActiveAccount(accounts) || {};
+export const hasSmartWalletAccount = (accounts: Accounts): boolean => {
```
vyvil: Duplicate (and not working) method. Please check for `userHasSmartWallet` method in this file :)

src/screens/Accounts/Accounts.js
```diff
@@ -295,40 +312,114 @@ class AccountsScreen extends React.Component<Props, State> {
...
-      ? accounts
-      : accounts.filter(({ type }) => type !== ACCOUNT_TYPES.SMART_WALLET);
-
-    const visibleAccountsWithBalance = visibleAccounts.map((acc) => {
-      const accountBalances = balances[acc.id] || {};
-      const balance = calculatePortfolioBalance(assets, rates, accountBalances);
-      return { ...acc, balance };
+
+    const visibleAccounts = this.visibleAccounts(accounts, smartWalletFeatureEnabled);
+    const hasAccount = hasSmartWalletAccount(accounts);
```
vyvil: Please use `userHasSmartWallet` method instead. This one seems not to work. And this results in constant smart wallet init button. We only need to show it for those users who does not have smart wallet

src/utils/smartWallet.js
```diff
@@ -54,25 +54,34 @@ const getMessage = (
...
+  return accounts.indexOf(acc => acc.type === ACCOUNT_TYPES.SMART_WALLET) !== -1;
+};
+
+export const getSmartWalletStatus = (
+  accounts: Accounts,
+  smartWalletState: SmartWalletReducerState,
+): SmartWalletStatus => {
+  const hasAccount = hasSmartWalletAccount(accounts);
```
vyvil: This should also be changed into `userHasSmartWallet` in order to work

src/screens/Accounts/Accounts.js
```diff
@@ -295,39 +315,132 @@ class AccountsScreen extends React.Component<Props, State> {
...
+    const ppnNetwork = blockchainNetworks.find((network) => network.id === BLOCKCHAIN_NETWORK_TYPES.PILLAR_NETWORK);
+    const availableStakeFormattedAmount = formatMoney(availableStake, 4);
+
+    if (smartWalletFeatureEnabled && ppnNetwork) {
+      networks.push({
+        id: `NETWORK_${ppnNetwork.id}`,
+        type: 'NETWORK',
+        isInitialised: isTankInitialised,
+        network: ppnNetwork,
+        formattedStakeAmount: `${availableStakeFormattedAmount} ${PPN_TOKEN}`,
```
vyvil: Just a note: this could be renamed into balance, so this property could be reusable for other networks :)

src/utils/smartWallet.js
```diff
@@ -54,35 +54,39 @@ const getMessage = (
...
     default:
       return {};
   }
 };
 
-export function getSmartWalletStatus(accounts: Accounts, smartWalletState: Object): SmartWalletStatus {
-  const account = accounts.find(acc => acc.type === ACCOUNT_TYPES.SMART_WALLET);
-  const activeAccount = getActiveAccount(accounts) || {};
+export const userHasSmartWallet = (accounts: Accounts = []): boolean => {
+  return accounts.some(acc => acc.type === ACCOUNT_TYPES.SMART_WALLET);
```
vyvil: 👍

src/screens/Accounts/Accounts.js
```diff
@@ -295,39 +315,132 @@ class AccountsScreen extends React.Component<Props, State> {
...
-        ...visibleAccountsWithBalance,
-        { type: 'SMART_WALLET_INIT', action: () => navigation.navigate(SMART_WALLET_INTRO) },
-      ]
-      : visibleAccountsWithBalance;
+    if (showSmartWalletInitButton) {
+      wallets.push({
+        type: 'NEW_SMART_WALLET',
+        interaction: {
+          isSmartWallet: true,
+          walletBalance: 0,
```
jsidorenko: @vyvil should we show the staked balance here?

src/models/Account.js
```diff
@@ -11,6 +11,7 @@ export type Account = {
   extra?: AccountExtra,
   isActive: boolean,
   walletId: string,
+  privateKey?: any,
```
jsidorenko: I think this shouldn't be here

src/screens/Accounts/Accounts.js
vyvil: This kind of button wasn't designed.
As long as we are showing 0 here, it not a problem. I suppose.
(We are doing the same with PPN)
We could double check with Drew. I'll ping him :)

src/screens/Accounts/Accounts.js
jsidorenko: This should be not the `Account` type, but `Wallet`

PR #1359, opened 2019-08-27
Optimise activity feed rendering

vyvil:
PR to move formatting of activity feed section lists from render, applying formatting only when feed data is changed


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -138,38 +138,64 @@ type State = {
...
     };
   }
 
+  componentDidMount() {
+    this.generateFeedSections();
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    const { tabs = [], feedData = [] } = this.props;
+    if ((tabs.length && !isEqual(tabs, prevProps.tabs))
```
poocart: What if comparing without length check? Should be fine and would look nicer. Or are there any other concerns?

src/components/Tabs/Tabs.js
```diff
@@ -298,10 +300,16 @@ export default class Tabs extends React.Component<Props, State> {
...
             <TabItem
+              activeOpacity={1}
               active={isActive}
-              onPress={() => this.setState({
-                activeTab: id,
-              }, onPress)}
+              onPress={() => {
+                if (onTabChange) onTabChange(true);
+                this.setState({ activeTab: id });
+                requestAnimationFrame(() => {
```
poocart: Maybe this needs to be called after state (active tab ID) has changed or it's ok?

src/components/ActivityFeed/ActivityFeed.js
vyvil: Property `tabs` is optional. Hence, if it's not provided, "if" check will return false immediately, without a need to call `isEqual`

src/components/Tabs/Tabs.js
vyvil: You mean, as a callback of setState?
All of those actions should happen at the same time - we set an active tab, and we render view, that is related to the active tab

src/components/Tabs/Tabs.js
```diff
@@ -45,6 +45,7 @@ type Props = {
   wrapperStyle?: Object,
   isFloating?: boolean,
   coverColor?: string,
+  onTabChange?: Function,
```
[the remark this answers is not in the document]
vyvil: Okay, I'll update typing in upcoming commit :)

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -138,38 +138,64 @@ type State = {
...
-function getSortedFeedData(tabs, activeTab, feedData) {
-  if (tabs.length) {
-    const aTab = tabs.find(({ id, data = [] }) => id === activeTab && data.length) || { data: [] };
-    return aTab.data.sort((a, b) => b.createdAt - a.createdAt);
-  } else if (feedData.length) {
-    return feedData.sort((a, b) => b.createdAt - a.createdAt);
-  }
-  return [];
+  tabIsChanging: boolean,
+  formattedFeedData: Object[],
```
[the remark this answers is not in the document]
vyvil: Will update it as well

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -138,38 +138,64 @@ type State = {
...
+    }
+  }
+
+  generateFeedSections = () => {
+    const { tabs = [], activeTab, feedData = [] } = this.props;
+    const dataSections = [];
+    let feedList = feedData;
+
+    if (tabs.length) {
+      const activeTabInfo = tabs.find(({ id }) => id === activeTab) || { data: [] };
```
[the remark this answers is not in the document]
vyvil: Sure :) Will update

vyvil:
@grilix, applied changes according to your comments. Let me know what you think :)

PR #1365, opened 2019-08-28
Fix/formatting fiat values

vyvil:
This PR adds a method to formate fiat values, that also attaches fiat currency symbol to the string


src/utils/common.js
```diff
@@ -131,6 +132,15 @@ export function getCurrencySymbol(currency: string): string {
   return currencies[currency] || '';
 }
 
+export function formatFiat(src: number | string, baseFiatCurrency?: ?string): string {
```
jsidorenko: It would be good to create some simple tests for this function

vyvil: okay, will add some :)

PR #1366, opened 2019-08-28
Feature/biometrics with device keystore

poocart:
Includes:
1. Removal of "react-native-touch-id" module.
2. Added React Native module for storing data in device encrypted keychain which can be unlocked using device Biometrics.
3. Added feature to take Private Key from keychain using biometrics when confirming any transactions (regular, collectibles, exchange, etc.) in the wallet.

Notes:
1. This PR uses forked version of library origin which has Android Fingerprint flow working and changes in this forked repo are under ready-to-merge PR to origin repo.
2. iOS biometrics keychain unlock flow includes native "Passcode" modal which is a fallback if biometrics are failed to enter multiple times. "Passcode" cannot be taken out as if it's removed then simply nothing appears that would unlock keychain.
3. iOS simulator doesn't bring any biometric or passcode authentication and if biometrics are enabled in wallet then in PIN screen it just goes through as successful without any input alert.
4. `Profile.js` screen is deprecated so changes in this PR only include basic update.


poocart:
P. S. Working on migration for older users.

src/screens/Profile/Profile.js
```diff
@@ -257,9 +257,9 @@ class Profile extends React.Component<Props, State> {
     this.setState((prev: State) => ({ showTrackingModal: !prev.showTrackingModal }));
   }
 
-  handleChangeUseBiometrics = (value) => {
+  handleChangeUseBiometrics = (value, privateKey) => {
```
jsidorenko: Could we rename `value` to something else? As it means nothing

poocart: Sure 👍

poocart: Updated in `Settings.js` also per PR note don't pay too much attention on `Profile.js`

poocart:
> P. S. Working on migration for older users.

Done.

src/utils/keychain.js
```diff
@@ -0,0 +1,56 @@
...
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+import { Platform } from 'react-native';
+import * as Keychain from 'react-native-keychain';
+
+const KEYCHAIN_SERVICE = 'com.pillarproject.wallet';
+const KEYCHAIN_DATA_KEY = 'data';
+const BIOMETRICS_PROMPT_MESSAGE = 'Continue';
+
+export function setKeychainDataObject(data: Object) {
```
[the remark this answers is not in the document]
poocart: Changed to `export const ...` although for data it can be anything inside, but as for now we're using only `privateKey` in it so changed type to match it.

PR #1367, opened 2019-08-28
enhancement/remove key generation signup

alazarevski:
Do not delay first login on signup for conn key generation
Keys will be generated 10 seconds after signup and will go in background, without blocking the UI main thread of React


jsidorenko:
Can't find in your PR any mention of mentioned 10 sec :)

jsidorenko:
Git diff doesn't make any sense :(
Correct me if I'm wrong, you just wrapped the generation logic with `if (oldConnectionsCount > 0 || currentConnectionsCount > connectionIdentityKeys.length) {` right?

PR #1371, opened 2019-08-29
Fix/smart wallet address mapping fixes

poocart:
Includes:
1. All mappings in all screens/components for Smart Contract wallet address to matching contact in transaction history entry.
2. From now on send flow if receiver has Smart Wallet address then it will be used instead of key based.
3. Additional improvement for checking Pillar Payment Network in asset send flow.
4. Additional check in Pillar Payment Network send flow if receiver address is not in network then throw alert.
5. Removed access keys in the request payload when requesting Smart Wallet addresses for existing contacts as it is not needed and caused issues when returning right addresses.
6. Other minor cleanups or minor fixes.


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -261,6 +268,22 @@ class ActivityFeed extends React.Component<Props, State> {
     }
   };
 
+  findMatchingContact = (address) => {
```
jsidorenko: How about extracting this into a separate function?

src/components/EventDetails/EventDetails.js
```diff
@@ -240,10 +256,35 @@ class EventDetails extends React.Component<Props, {}> {
     navigation.navigate(CHAT, { username: contact.username });
   };
 
+  findMatchingContact = (address) => {
+    const {
+      contacts,
+      contactsSmartAddresses = [],
+    } = this.props;
+    const contact = contacts.find(({ id: contactId, ethAddress }) => {
```
jsidorenko: DRY

src/utils/accounts.js
```diff
@@ -57,3 +57,22 @@ export function getActiveAccountWalletId(accounts: Accounts = []): string {
...
+      return 'Smart Wallet';
+    case ACCOUNT_TYPES.KEY_BASED:
+      return 'Key Based account';
+    default:
+      return '';
+  }
+}
+
+export function getUserAccounts(accounts: Accounts) {
+  const accountsWithoutActive = accounts.filter(({ isActive }) => !isActive);
```
jsidorenko:
```suggestion
  const inactiveAccounts = accounts.filter(({ isActive }) => !isActive);
```

src/utils/blockchainNetworks.js
```diff
@@ -0,0 +1,27 @@
...
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+import { BLOCKCHAIN_NETWORK_TYPES } from 'constants/blockchainNetworkConstants';
+import type { BlockchainNetwork } from 'models/BlockchainNetwork';
+
+export const isPillarPaymentNetworkActive = (blockchainNetworks: BlockchainNetwork[]) => {
+  const activeBNetwork = blockchainNetworks.find((network) => network.isActive) || { id: '', title: '' };
```
jsidorenko: could be:
```
  const activeBNetwork = blockchainNetworks.find(({ isActive }) => isActive);
  return get(activeBNetworkId, 'id') === BLOCKCHAIN_NETWORK_TYPES.PILLAR_NETWORK;
```

src/utils/feedData.js
```diff
@@ -18,20 +18,43 @@
...
+  accounts: Accounts,
+  eventType: string,
+) {
+  const userAccounts = getUserAccounts(accounts);
   const concatedHistory = history
     .map(({ ...rest }) => ({ ...rest, type: eventType }))
     .map(({ to, from, ...rest }) => {
-      const contact = contacts.find(({ ethAddress }) => {
-        return addressesEqual(from, ethAddress) || addressesEqual(to, ethAddress);
+      const contact = contacts.find(({ id: contactId, ethAddress }) => {
```
jsidorenko: looks like DRY

src/screens/Asset/Asset.js
```diff
@@ -104,6 +107,7 @@ type Props = {
   history: Array<*>,
   logScreenView: (contentName: string, contentType: string, contentId: string) => void,
   availableStake: number,
+  contactsSmartAddresses: ContactSmartAddresses[],
```
jsidorenko: Type should be `ContactSmartAddresses` or `ContactSmartAddress`

src/utils/accounts.js
```diff
@@ -57,3 +58,31 @@ export function getActiveAccountWalletId(accounts: Accounts = []): string {
...
+    case ACCOUNT_TYPES.SMART_WALLET:
+      return 'Smart Wallet';
+    case ACCOUNT_TYPES.KEY_BASED:
+      return 'Key Based account';
+    default:
+      return '';
+  }
+}
+
+export function getInactiveUserAccounts(accounts: Accounts): UserAccount[] {
```
jsidorenko: This method is so confusing. What is the difference between Account and UserAccount? We need to name the second better

src/utils/accounts.js
```diff
@@ -57,3 +58,31 @@ export function getActiveAccountWalletId(accounts: Accounts = []): string {
...
+  return accounts
+    .filter(({ isActive }) => !isActive)
+    .map(account => ({
+      isUserAccount: true,
+      ethAddress: getAccountAddress(account),
+      username: getAccountName(account.type),
```
jsidorenko: Username that could be either `Smart Wallet` or `Key Based account`? Should be renamed to `label` or whatever, but not `username`

src/utils/blockchainNetworks.js
```diff
@@ -0,0 +1,25 @@
...
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+import { BLOCKCHAIN_NETWORK_TYPES } from 'constants/blockchainNetworkConstants';
+import type { BlockchainNetwork } from 'models/BlockchainNetwork';
+
+export const isPillarPaymentNetworkActive = (blockchainNetworks: BlockchainNetwork[]): boolean => {
+  return !!blockchainNetworks.find(({ isActive, id }) => !!isActive && id === BLOCKCHAIN_NETWORK_TYPES.PILLAR_NETWORK);
```
jsidorenko:
```suggestion
  return blockchainNetworks.some(({ isActive, id }) => !!isActive && id === BLOCKCHAIN_NETWORK_TYPES.PILLAR_NETWORK);
```

src/screens/Asset/Asset.js
poocart: After discussions renamed to `ContactSmartAddressData`.

PR #1376, opened 2019-09-02
chore/initial connection pre-key generation in sync

alazarevski:
Generate 2 pre connection keys in sync on new wallet, then continue in background


src/actions/connectionKeyPairActions.js
```diff
@@ -257,6 +257,18 @@ export const updateConnectionKeyPairs = (
     const { currentConnectionsCount, oldConnectionsCount, newReceivedConnectonsCount } = numberOfConnections;
     const totalConnections = currentConnectionsCount + oldConnectionsCount + newReceivedConnectonsCount;
 
+    if (generateKeys && totalConnections === 0 && connectionKeyPairs.length === 0 && lastConnectionKeyIndex === -1) {
+      const promiseJobs = await generateKeyPairPool(mnemonic, privateKey, -1, 0, 2);
+      const resultPairs = await Promise.all(promiseJobs.map(task => task()));
+      const allPairsResults = [].concat(...resultPairs);
+      const initialConnKeyPairs = allPairsResults.sort((a, b) => { return a.connIndex < b.connIndex ? -1 : 1; });
```
jsidorenko:
```suggestion
      const initialConnKeyPairs = allPairsResults.sort((a, b) => a.connIndex < b.connIndex ? -1 : 1);
```

src/actions/connectionKeyPairActions.js
```diff
@@ -257,6 +257,18 @@ export const updateConnectionKeyPairs = (
     const { currentConnectionsCount, oldConnectionsCount, newReceivedConnectonsCount } = numberOfConnections;
     const totalConnections = currentConnectionsCount + oldConnectionsCount + newReceivedConnectonsCount;
 
+    if (generateKeys && totalConnections === 0 && connectionKeyPairs.length === 0 && lastConnectionKeyIndex === -1) {
+      const promiseJobs = await generateKeyPairPool(mnemonic, privateKey, -1, 0, 2);
```
poocart: `-1` for no index?
`0` connections count?
`2` only 2 keys to generate?

poocart: seems like an edge case scenario, maybe you could add a comment why these values are chosen?

alazarevski: -1 would be no keys generated yet at all, 0 for connections number in order not to overgenerate those 2 ( countToGenerate + connectionsCount in the function algorithm otherwise)

src/actions/connectionKeyPairActions.js
```diff
@@ -257,6 +257,18 @@ export const updateConnectionKeyPairs = (
     const { currentConnectionsCount, oldConnectionsCount, newReceivedConnectonsCount } = numberOfConnections;
     const totalConnections = currentConnectionsCount + oldConnectionsCount + newReceivedConnectonsCount;
 
+    if (generateKeys && totalConnections === 0 && connectionKeyPairs.length === 0 && lastConnectionKeyIndex === -1) {
```
jsidorenko: Pls put a description under the `// NOTE:` here why do we need this

src/actions/connectionKeyPairActions.js
poocart: thanks! 👍  I'm sorry, can you add this comment to JS? 🙏

src/actions/connectionKeyPairActions.js
alazarevski: 92969f7

src/actions/connectionKeyPairActions.js
alazarevski: 92969f7

PR #1383, opened 2019-09-04
Feature/pending transaction speed up

poocart:
Includes speed up options for regular transactions that basically puts transaction with same nonce over existing transaction which should reduce confirm wait time. This also includes "View transaction" button in Smart Wallet asset transfer preview.

Note: this does not include transaction speed up for collectibles (missing piece is to build payload) and for Smart Wallet SDK transactions (payments) as main focus for this feature is to get more control over assets transfer during migration from key based wallet to Smart wallet.


poocart:
Note: I noticed that sometimes even Etherscan provider (neither it's API!) does not return pending transaction by hash (eth_getTransactionByHash) even though it's visible in Etherscan explorer.

src/actions/assetsActions.js
```diff
@@ -709,3 +734,62 @@ export const resetLocalNonceToTransactionCountAction = (wallet: Object) => {
...
+    const gasPrice = txFeeInWei.div(gasLimit).toNumber();
+    const {
+      to,
+      from,
+      asset: assetSymbol,
+      note,
+      value,
+    } = transactionInfo;
+    // $FlowFixMe
+    const asset: Asset = Object.values(assets).find(({ symbol }) => symbol === assetSymbol) || {};
```
grilix: Can't we use `const asset = assets[assetSymbol];`?

In other hand, maybe we could handle the scenario where the asset does not exist? On that case, sending the user to the next screen sounds weird.

```js
const asset = assets[assetSymbol]; // (or whatever we use here)
if (!asset) {
  // ...
  return;
}
// ...

Comment (jsidorenko) [reply src/actions/assetsActions.js]: Right, we should use `const asset = assets[assetSymbol];`

File: src/components/DeploymentView/DeploymentView.js [hunk 2]
```diff
@@ -80,6 +120,41 @@ class DeploymentView extends React.PureComponent<Props> {
         SMART_WALLET_UPGRADE_STATUSES.TRANSFERRING_ASSETS,
       ].includes(smartWalletStatus.status);
 
+    let detailedTransaction;
+    const matchingTransaction = history.find(({ hash }) => hash === 'x');
```
jsidorenko: How this could be possible? I mean to get a tx with the hash equal 'x'?

poocart: Committed by mistake, fixed.

poocart:
Putting as WIP as to properly test it we need to get pending transactions along with history API call during wallet import and currently there's an issue in BCX that doesn't return pending transactions, Issue has been reported.

PR #1385, opened 2019-09-04
Fix/wallet obj storage backup

poocart:
This includes a fix of issue when `wallet` object or `wallet` timestamp from `appSettings` are gone in storage. As we cannot find the main issue why these values become missing now we will make backup of `wallet` object and restore it if it's missing while also doing the same with `wallet` timestamp from `appSettings`. Also added Sentry log report for this case.

Additionally small cleanups on few of methods in `utils/wallet.js`.


src/utils/wallet.js
```diff
@@ -70,26 +74,68 @@ export function catchTransactionError(e: Object, type: string, tx: Object) {
...
-export async function signPersonalMessage(message: string, wallet: Object): Promise<string> {
+export function signPersonalMessage(message: string, wallet: Object): Promise<string> {
   wallet.provider = providers.getDefaultProvider(NETWORK_PROVIDER);
-  const result = await wallet.signMessage(isHexString(message) ? ethers.utils.arrayify(message) : message);
-  return result;
+  return wallet.signMessage(isHexString(message) ? ethers.utils.arrayify(message) : message);
+}
+
+// we use basic AsyncStorage implementation just to prevent backup being stored in same manner
+export async function getWalletFromStorage(dispatch: Dispatch, appSettings: Object) {
```
grilix: Small suggestions: I'd avoid the `get` on function names and use something like `walletFromStorage` or just `storedWallet`, for example. Also, what do you think about adding a type for what this function returns? I'm not sure if it would be worth it or not :thinking:

src/utils/wallet.js
```diff
@@ -70,26 +74,68 @@ export function catchTransactionError(e: Object, type: string, tx: Object) {
...
+    // only wallet timestamp was missing, let's update it to storage
+    dispatch(saveDbAction('app_settings', { appSettings: { wallet: walletTimestamp } }));
+  }
+  // we check for previous value of `appSettings.wallet` as by this point `walletTimestamp` can be already set
+  if (isWalletEmpty || !appSettings.wallet) {
+    Sentry.captureMessage('Wallet login issue spotted', {
+      extra: {
+        missing: {
+          wallet: isWalletEmpty,
+          walletTimestamp: !appSettings.wallet,
```
grilix: Maybe we can call these the same? That way it would make more sense when comparing the entry with the code. Like `missing: { isWalletEmpty, walletTimestamp, ... }`.

src/utils/wallet.js
jsidorenko: Why would you avoid the `get` on function names?

src/utils/wallet.js
jsidorenko: I would call it `walletCreationTimestamp`

src/utils/wallet.js
grilix: I think `get` adds no information to the name whatsoever. All of the functions that return information will get it from somewhere, we don't need to add that `get` into the name to make it explicit. Unless there's a constraint, I'd avoid things like `get`, `calculate`, `check`, `do`, etc.

jsidorenko: As for me, those addons make code more readable :)

grilix: Alright, that's fine :+1:

poocart: 👍

src/utils/wallet.js
poocart: Updated.

src/utils/wallet.js
```diff
@@ -70,26 +74,66 @@ export function catchTransactionError(e: Object, type: string, tx: Object) {
...
-  const result = await wallet.signMessage(isHexString(message) ? ethers.utils.arrayify(message) : message);
-  return result;
+  return wallet.signMessage(isHexString(message) ? ethers.utils.arrayify(message) : message);
+}
+
+// we use basic AsyncStorage implementation just to prevent backup being stored in same manner
+export async function getWalletFromStorage(dispatch: Dispatch, appSettings: Object) {
+  let { wallet = {} } = await storage.get('wallet');
+  const walletBackup = await AsyncStorage.getItem(WALLET_STORAGE_BACKUP_KEY);
+  const isWalletEmpty = !wallet || !Object.keys(wallet).length;
```
jsidorenko:
```suggestion
  const isWalletEmpty = isEmpty(wallet);
```

jsidorenko: from lodash

src/utils/wallet.js
```diff
@@ -70,26 +74,66 @@ export function catchTransactionError(e: Object, type: string, tx: Object) {
...
+  // wallet timestamp missing causes welcome screen
+  let walletTimestamp = appSettings.wallet;
+  if (isWalletEmpty && walletBackup) {
+    console.log('RESTORING WALLET FROM BACKUP');
+    // restore wallet to storage
+    wallet = JSON.parse(walletBackup);
```
jsidorenko: try..catch ?

src/utils/wallet.js
```diff
@@ -70,26 +74,66 @@ export function catchTransactionError(e: Object, type: string, tx: Object) {
...
+    dispatch(saveDbAction('wallet', { wallet }));
+  }
+  // we can only set new timestamp if any wallet is present (existing or backup)
+  if (!walletTimestamp && (!isWalletEmpty || walletBackup)) {
+    walletTimestamp = +(new Date());
```
jsidorenko:
```suggestion
    walletTimestamp = +new Date();
```

src/utils/wallet.js
```diff
@@ -70,26 +74,66 @@ export function catchTransactionError(e: Object, type: string, tx: Object) {
...
+  // wallet timestamp missing causes welcome screen
+  let walletTimestamp = appSettings.wallet;
+  if (isWalletEmpty && walletBackup) {
+    console.log('RESTORING WALLET FROM BACKUP');
+    // restore wallet to storage
+    wallet = JSON.parse(walletBackup);
+    dispatch(saveDbAction('wallet', { wallet }));
+  }
+  // we can only set new timestamp if any wallet is present (existing or backup)
+  if (!walletTimestamp && (!isWalletEmpty || walletBackup)) {
```
jsidorenko:
```suggestion
  if (!walletTimestamp && !isEmpty(wallet)) {
```
check once again with isEmpty as you change the wallet object above. (and because the backup parse method could fail)

src/utils/wallet.js
```diff
@@ -70,26 +74,66 @@ export function catchTransactionError(e: Object, type: string, tx: Object) {
...
+    walletTimestamp = +(new Date());
+    console.log('SETTING NEW WALLET TIMESTAMP');
+    // only wallet timestamp was missing, let's update it to storage
+    dispatch(saveDbAction('app_settings', { appSettings: { wallet: walletTimestamp } }));
+  }
+  // we check for previous value of `appSettings.wallet` as by this point `walletTimestamp` can be already set
+  if (isWalletEmpty || !appSettings.wallet) {
```
jsidorenko: Why no to join this method with the block above where you set the new timestamp?

src/utils/wallet.js
poocart: lodash to the moon

src/utils/wallet.js
poocart: We must check either if wallet was empty or walletBackup exists because wallet might be empty, but was restored from backup.

src/utils/wallet.js
poocart: Because it's the same, but under one more level of brackets.

src/utils/wallet.js
poocart: Why? Nothing to put in catch case as we just check if there was no wallet and backup exists and then restore it. I. e. if we put `wallet = JSON.parse(walletBackup)` and save in try then catch will still remain empty, or you're thinking of something else?

src/utils/wallet.js
jsidorenko: if it was restored, then it would be put into the same `wallet` variable

PR #1387, opened 2019-09-04
Auto-sync tx fee data on tx details open

jsidorenko:


src/components/EventDetails/EventDetails.js
```diff
@@ -170,11 +171,15 @@ class EventDetails extends React.Component<Props, {}> {
       getTxNoteByContact(eventData.contact.username);
     }
 
-    const txInfo = this.props.history.find(tx => tx.hash === eventData.hash) || {};
-    if (txInfo.status !== TX_PENDING_STATUS) return;
+    const txInfo = history.find(tx => tx.hash === eventData.hash) || {};
+    if (txInfo.status === TX_PENDING_STATUS) {
+      this.timeout = setTimeout(() => updateTransactionStatus(eventData.hash), 500);
+      this.timer = setInterval(() => updateTransactionStatus(eventData.hash), 10000);
```
poocart: I think we can still add return here to not iterate through following code also in speed up feature PR there's more code for pending transaction.

PR #1388, opened 2019-09-04
Sort activity feed by date

jsidorenko:


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -242,9 +243,13 @@ class ActivityFeed extends React.Component<Props, State> {
...
 
-    feedList.forEach(listItem => {
+    orderBy(feedList, ['createdAt'], ['desc']).forEach(listItem => {
       const formattedDate = formatDate(new Date(listItem.createdAt * 1000), 'MMM D YYYY');
-      const sectionTitle = formatDate(new Date(listItem.createdAt * 1000), 'MMM D');
+      // don't show the year if the event happened this year
+      const titleDateFormat = new Date(listItem.createdAt * 1000).getFullYear() === new Date().getFullYear()
+        ? 'MMM D'
+        : 'MMM D YYYY';
+      const sectionTitle = formatDate(new Date(listItem.createdAt * 1000), titleDateFormat);
```
grilix: It looks like we're repeating the `new Date(listItem.createdAt * 1000)` a couple of times, maybe we can store it in a const?

PR #1389, opened 2019-09-04
Refactor balance calculation, improve consistency

grilix:
Portfolio balance calculation was being calculated for all currencies,
it's now gonna be calculated only for the selected currency, which
should reduce the overall complexity of the involved code.

Also included on this commit, further updates improving code syntax
consitency and typing.


src/components/PortfolioBalance/PortfolioBalance.js
```diff
@@ -21,37 +21,38 @@ import * as React from 'react';
...
       rates,
       balances,
       baseFiatCurrency,
       label,
     } = this.props;
 
-    const portfolioBalances = calculatePortfolioBalance(assets, rates, balances);
     const fiatCurrency = baseFiatCurrency || defaultFiatCurrency;
-    const portfolioBalance = formatFiat(portfolioBalances[fiatCurrency], baseFiatCurrency);
+    const balance = balanceInCurrency(rates, balances, fiatCurrency);
```
jsidorenko:
```suggestion
    const balance = balanceInFiat(rates, balances, fiatCurrency);
```

jsidorenko: When you read the code without knowing the details it could be confusing as we have another method with similar naming `formatFiat()`, that's why there was a prefix `calculate` as it shows the intention and you can assume what it does without going into the details

src/utils/accounts.js
```diff
@@ -17,67 +17,82 @@
...
 
-export function checkIfSmartWalletAccount(account: Account): boolean {
+export const checkIfSmartWalletAccount = (account: Account): boolean => {
   return account.type === ACCOUNT_TYPES.SMART_WALLET;
-}
+};
 
-export function getAccountName(accountType: AccountTypes): ?string {
-  if (!accountType) return null;
+export const getAccountName = (accountType: AccountTypes): ?string => {
```
jsidorenko:
```suggestion
export const getAccountName = (accountType: AccountTypes): string => {
```

src/utils/assets.js
```diff
@@ -19,24 +19,37 @@
...
+
+  const number = new BigNumber(assetBalance.balance);
 
-export function getBalance(balances: Balances = {}, asset: string = ''): number {
-  const number = balances[asset] ? new BigNumber(balances[asset].balance) : 0;
   return +formatAmount(number.toString());
-}
+};
 
-export function getRate(rates: Rates = {}, token: string, fiatCurrency: string): number {
```
jsidorenko: Now it became less safe :)
How about using getters for `tokenRates` and `ethRate` ?

src/utils/common.js
```diff
@@ -34,36 +34,43 @@ import {
...
-  return Math.floor(Math.random() * (max - min + 1)) + min; // eslint-disable-line
-}
+export const getRandomInt = (min: number, max: number): number => {
+  return Math.floor(Math.random() * ((max - min) + 1)) + min;
+};
 
-export function decodeETHAddress(encodedAddress: string) {
-  if (!encodedAddress || encodedAddress.substr(0, 9) !== 'ethereum:') {
+export const decodeETHAddress = (encodedAddress: string): string => {
+  if (!encodedAddress || !encodedAddress.startsWith('ethereum:')) {
```
jsidorenko: 👍

src/utils/common.js
```diff
@@ -91,70 +98,70 @@ export function formatMoney(
   }
 
   return (c ? num.replace('.', c) : num).replace(new RegExp(re, 'g'), `$&${s || ','}`);
-}
+};
 
-export function parseNumber(amount: string = '0') {
-  let strg = amount.toString() || '';
+export const parseNumber = (amount: string) => {
```
jsidorenko: Are you sure we're passing string here as a param? I think in some places BN instance could occure

src/utils/common.js
```diff
@@ -91,70 +98,70 @@ export function formatMoney(
...
   strg = strg.replace(new RegExp(`[^0-9$${decimal}]`, 'g'), '');
   strg = strg.replace(',', '.');
+
   return parseFloat(strg);
-}
+};
 
-export function isValidNumber(amount: string = '0') {
-  const strg = amount.toString() || '';
+export const isValidNumber = (amount: string) => {
```
jsidorenko: Same as above, I would ensure first that we're passing the right param here

src/services/exchange.js
```diff
@@ -149,7 +149,7 @@ export default class ExchangeService {
   }
 
   getShapeshiftAuthUrl() {
-    const { sub: regId } = extractJwtPayload(this.tokens.accessToken) || {};
+    const { sub: regId } = extractJwtPayload(this.tokens.accessToken);
```
jsidorenko: What if we fail to parse the JSON object inside `extractJwtPayload()` return?

src/utils/common.js
```diff
@@ -262,51 +268,57 @@ export function getEthereumProvider(network: string) {
...
-}
+};
 
-export function extractJwtPayload(jwtToken: string): Object {
+export const extractJwtPayload = (jwtToken: string): Object => {
   // extract: header (not needed), payload, signature (not needed)
   const [, encodedPayload] = jwtToken.split('.');
+
   // do not use Buffer.toJSON
   return JSON.parse(Buffer.from(encodedPayload, 'base64').toString());
```
jsidorenko: I think it's better to wrap this with try..catch and return `{}` if we fail to parse

src/components/PortfolioBalance/PortfolioBalance.js
grilix: What do you mean? What kind of confusion?

It's perfectly safe to assume that this function is not gonna calculate anything, just give you the balance in fiat; it won't make any difference.

src/services/exchange.js
grilix: It will throw an exception? The signature of `extractJwtPayload` states that it will always return an `Object`.

src/utils/assets.js
```diff
@@ -63,59 +76,69 @@ export function getRate(rates: Rates = {}, token: string, fiatCurrency: string):
...
+
+  return balanceValues.reduce((total, item) => {
+    const balance = +item.balance;
+    const assetRates = rates[item.symbol];
+
+    if (!assetRates || balance === 0) {
+      return total;
+    }
+
+    return total + (assetRates[ETH] * balance);
```
jsidorenko: What would happen if there won't be `assetRates[ETH]` set?

src/services/exchange.js
jsidorenko: Yes, as the signature states we always get an object back, then we need to sure that we return that object in all the cases.
TLDR: just wrap the return with try .. catch

grilix: Oh, you mean updating `extractJwtPayload`? I can do that :+1:

src/utils/assets.js
grilix: Do you think we're ever passing `rates = undefined` here? Flow should prevent that, since the argument is not optional :thinking:

src/utils/common.js
grilix: The `= '0'`, should not have any effect on a BN being passed, or does it? :thinking:

src/utils/assets.js
jsidorenko: We're fetching the rates from external api and never know what it returns back :)
Probably the ` = {}` was put here because we already had this sort of issues

src/utils/common.js
jsidorenko: `= '0'` saves from undefined values, the next line `let strg = amount.toString() || '';` was saving from BigNumbers passed in

src/utils/assets.js
grilix: Pain would happen, good catch!

src/utils/assets.js
grilix: Okay, I understand it's probably not the best time to change this. Let's see if we can improve it later.

src/components/PortfolioBalance/PortfolioBalance.js
grilix: Updated.

src/utils/accounts.js
grilix: Updated.

src/utils/assets.js
grilix: Updated.

src/utils/common.js
grilix: Updated and added a couple of tests covering those scenarios to make sure we have the expected behaviour.

src/utils/common.js
grilix: Updated and added a couple of tests just in case.

src/utils/common.js
grilix: Updated.

src/utils/assets.js
grilix: Updated.

src/utils/accounts.js
```diff
@@ -17,67 +17,82 @@
     with this program; if not, write to the Free Software Foundation, Inc.,
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
-import get from 'lodash.get';
 import type { Account, Accounts, AccountTypes } from 'models/Account';
 import { ACCOUNT_TYPES } from 'constants/accountsConstants';
 import { addressesEqual } from './assets';
 
-export function getActiveAccount(accounts: Accounts = []): ?Account {
+export const getActiveAccount = (accounts: Accounts): ?Account => {
```
poocart: What if just `...  => accounts.find(({ isActive }) => isActive);`?

src/utils/accounts.js
```diff
@@ -17,67 +17,82 @@
...
   return accounts.find(({ isActive }) => isActive);
-}
+};
 
-export function getActiveAccountId(accounts: Accounts = []): string {
-  const activeAccount = getActiveAccount(accounts) || {};
-  return activeAccount.id || '';
-}
+export const getInactiveUserAccounts = (accounts: Accounts): Accounts => {
```
poocart: What if just `... => accounts.filter(({ isActive }) => !isActive);`?

src/utils/accounts.js
```diff
@@ -17,67 +17,82 @@
...
+  return accounts.filter(({ isActive }) => !isActive);
+};
 
-export function getActiveAccountType(accounts: Accounts = []): string {
-  const activeAccount = getActiveAccount(accounts) || {};
-  return activeAccount.type || '';
-}
+export const getActiveAccountId = (accounts: Accounts): string => {
+  const activeAccount = getActiveAccount(accounts);
+  if (!activeAccount) {
```
poocart: What if simple no brackets return?

src/utils/accounts.js
grilix: I don't think that's gonna improve readability :thinking:

src/utils/accounts.js
grilix: Hmm.. Well this seems a bit subjective, and I understand there could be some personal differences.

That being said, I think one liners move the code towards more dense and less legible syntax. The more things we have on each line, the more difficult it becomes to read a given block of code.

Lines are not free, but they are cheaper than complexity. Plus, if a file gets too big, we can move parts of it into other files. While if the code gets too complex, we have to refactor it, which is harder to do and harder to validate everything's still working as expected. This is not gonna be a huge difference on one way or another, but I think it's gonna be a bit more consistent with more complex functions, which should be a good thing.

Personally, I feel like this:

```js
const func = (a?: number): number => {
  if (!a) {
    return 0;
  }

  return a / 2;
}
```

Is cleaner than:

```js
const func = (a?: number): number => {
  if (!a) return 0;
  return a / 2;
}
```

Of course there are some exceptions, but normally I wouldn't try to squeeze things on less lines.

What do you think?

jsidorenko: as an old-school dev I agree with @grilix, although in modern books you'll find that code readability is not important our days, and it's normal to write the code this way: `func = () => () => () => {}`. But good luck to construct this in your head when you do a quick code reading

src/utils/assets.js
```diff
@@ -63,59 +76,71 @@ export function getRate(rates: Rates = {}, token: string, fiatCurrency: string):
...
+  currency: string,
+) => {
+  const ethRates = rates[ETH];
+  if (!ethRates) {
+    return 0;
+  }
+
+  const totalEth = balanceInEth(balances, rates);
+
+  return ethRates[currency] * totalEth;
```
jsidorenko:
```suggestion
  return get(ethRates, currency, 0) * totalEth;
```

src/utils/assets.js
```diff
@@ -19,24 +19,37 @@
...
+export const transformAssetsToObject = (assetsArray: Asset[] = []): Assets => {
   return assetsArray.reduce((memo, asset) => {
     memo[asset.symbol] = asset;
     return memo;
   }, {});
-}
+};
+
+export const getBalance = (balances: Balances, asset: string): number => {
+  const assetBalance = balances[asset];
```
jsidorenko:
```suggestion
  const assetBalance = get(balances, asset);
```

PR #1390, opened 2019-09-04
Fix assets list not showing low balance

grilix:
When the asset balance is below two digits, the assets list shows 0,
passing the `decimals` to `formatMoney` should fix this problem.

# Example

On this case, the balance in ETH is `0.001`:

## Before

assets-prev

## After

assets-after

### DIFF
--- a/src/screens/Assets/AssetsList.js
+++ b/src/screens/Assets/AssetsList.js
@@ -175,7 +175,7 @@ class AssetsList extends React.Component<Props> {
         label={name}
         avatarUrl={fullIconUrl}
         balance={{
-          balance: formatMoney(balance),
+          balance: formatMoney(balance, decimals),
           value: formattedBalanceInFiat,
           token: symbol,
         }}


grilix:
@jsidorenko damn, you're right. I think the default decimals for `formatMoney` is 2, tho, which's why `0.001` results in`"0"`.

Some alternatives I can think of:
- To be able to use a dynamic amount of decimals depending on the integer part. For example: `1.234567` => 6, `12.34567` => 5, `123.4567` => 4, `1234567.89` => 2.
- Add some indication that the value is being rounded, like a character or a symbol (ex, `~0 ETH`).
- Have a new rounding function with a more sensible default for tokens (it doesn't really make sense to have a default of more than 2 decimals on fiat).
- Use a predefined amount of decimal places on this screen, or something like `min(decimals, 6)` (for a scenario where token decimals is less than 6).

What do you think? Any other option you'd suggest?

vyvil:
@grilix, 
just a note regarding this part:

> Have a new rounding function with a more sensible default for tokens (it doesn't really make sense to have a default of more than 2 decimals on fiat).

We already have a separate method to format fiat and are no longer using formatMoney to format fiat values

grilix:
@vyvil @jsidorenko, thanks for the comments. I thought this was going to be an minimal change and it's not. We need to consider different scenarios and come up with a real solution if we want to fix this issue.

PR #1400, opened 2019-09-09
Add initial tests for WalletConnect reducer

grilix:
This adds some simple tests for WalletConnect. These are intended to be
the initial step towards reworking WalletConnect actions and the reducer
itself.

Since this includes no functional changes, it's a low risk
changeset, which should speed up review discussions and leave the
discussion for the refactor isolated from this.

### DIFF
--- a/src/reducers/walletConnectReducer.js
+++ b/src/reducers/walletConnectReducer.js
@@ -31,8 +31,8 @@ import {
 } from 'constants/walletConnectConstants';
 
 export type WalletConnectReducerState = {
-  connectors: Array<WalletConnect>,
-  pending: Array<WalletConnect>,
+  connectors: WalletConnect[],
+  pending: WalletConnect[],
   requests: Array<{
     peerId: string,
     peerMeta: Object,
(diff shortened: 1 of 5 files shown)


src/reducers/walletConnectReducer.js
```diff
@@ -58,34 +119,39 @@ const initialState = {
   waitingRequest: null,
 };
 
-export default function walletConnectReducer(
+const walletConnectReducer = (
   state: WalletConnectReducerState = initialState,
   action: WalletConnectReducerAction,
-) {
-  const { payload } = action;
```
grilix: This is just because of flow getting confused, I'll look into it on the next update.

src/reducers/walletConnectReducer.js
```diff
@@ -45,11 +45,72 @@ export type WalletConnectReducerState = {
   waitingRequest: ?string,
 };
 
-export type WalletConnectReducerAction = {
-  type: string,
-  payload: any,
+type SessionApproved = {
+  type: 'WALLETCONNECT_SESSION_APPROVED',
+  payload: {
+    pending: Array<WalletConnect>,
```
jsidorenko:
```suggestion
    pending: WalletConnect[],
```
to keep the consistency with other project types

src/testUtils/walletConnectMock.js
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+let peerId = 0;
+
+export default class WalletConnect {
+  peerId: string;
+
+  constructor() {
+    this.peerId = `peer-${peerId}`;
+    peerId++;
```
jsidorenko: Just curious if we could utilize static vars instead of global?

src/reducers/walletConnectReducer.js
grilix: Oops, you're right. Updated.

src/testUtils/walletConnectMock.js
grilix: Yeah, that sounds like a better idea, updated.

src/reducers/walletConnectReducer.js
```diff
@@ -45,11 +45,72 @@ export type WalletConnectReducerState = {
   waitingRequest: ?string,
 };
 
-export type WalletConnectReducerAction = {
-  type: string,
-  payload: any,
+type SessionApproved = {
+  type: 'WALLETCONNECT_SESSION_APPROVED',
```
jsidorenko: I think we should use constants here in order to avoid mistyping

src/testUtils/walletConnectMock.js
```diff
@@ -0,0 +1,34 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+export default class WalletConnect {
+  static _nextId: number;
```
jsidorenko:
```suggestion
  static _nextId: number = 0;
```

src/testUtils/walletConnectMock.js
```diff
@@ -0,0 +1,34 @@
...
+  peerId: string;
+
+  static nextId() {
+    this._nextId = this._nextId || 0;
```
jsidorenko: could get rid of this line by adding a default value to the static function

src/reducers/walletConnectReducer.js
grilix: Unfortunately, flow does not support using constants here, and they encourage this form: https://flow.org/en/docs/react/redux/#toc-typing-redux-actions. The only option is to use `typeof`, which results in `string`.

Using `string` here results in super weird error messages and flow forces you to even type empty arrays so it can choose the correct type.

src/testUtils/walletConnectMock.js
grilix: Perfect!

PR #1402, opened 2019-09-10
Fix/add assets to assetslist based on balance

vyvil:
This PR solves a missing assets problem - some of users owned assets tended not to appear on assets lists, hence:

1. New assets are added to assets list based on balance. It is done by updating previously used `checkForMissedAssetsAction` to take balances (not transaction history) into consideration.
2. Assets are now divided into separate list per account (just like history, rates and balances).
3. HiddenAssets list is introduced to hold all user hidden assets. As well as related actions:

- `hideAssetAction`
adds hidden asset ticker to array (each wallet has a separate array of hidden assets);
DOES NOT remove asset from the list in assetsReducer;

- `showAssetAction`
removes asset’s ticker from related array in userSettingsReducer;
DOES NOT add asset to assets list in assetsReducer

HiddenAssets list is used to filter out hidden assets (in assets selector).
This allows:
- to show all user owned assets on asset send flow (if ever needed).
- to exclude hidden assets when calculating wallet balance.

HiddenAssets list is stored in a new UserSettings reducer, that could later on be used to store other user related settings.

4. In order to use latest assets to fetch balances (and remove newly added asset balances to be overwritten on import), assets are now being taken straight from store in `fetchAssetsBalancesAction` (rather than being passed in as a prop).

5. Swipe to remove is now added to updated assets list items.


src/actions/assetsActions.js
```diff
@@ -498,32 +499,35 @@ function notifyAboutIncreasedBalance(newBalances: Balance[], oldBalances: Balanc
...
     } = getState();
 
     const activeAccount = getActiveAccount(accounts);
     if (!activeAccount) return;
+    const walletAddress = getAccountAddress(activeAccount);
+    if (!walletAddress) return;
 
-    const walletAddress = getActiveAccountAddress(accounts);
-    const accountId = getActiveAccountId(accounts);
+    const accountAssets = allAssets[walletAddress] || {};
```
jsidorenko: We could create a selector for this

src/actions/assetsActions.js
```diff
@@ -551,61 +556,68 @@ export const fetchAssetsBalancesAction = (assets: Assets, showToastIfIncreased?:
...
-
     if (!Object.keys(initialAssets).length) {
       dispatch({
         type: UPDATE_ASSETS_STATE,
         payload: FETCH_INITIAL_FAILED,
       });
       return;
     }
-
+    const activeAccountId = getActiveAccountAddress(accounts);
```
jsidorenko:
```suggestion
    const activeAccountId = getActiveAccountId(accounts);
```

src/actions/assetsActions.js
```diff
@@ -551,61 +556,68 @@ export const fetchAssetsBalancesAction = (assets: Assets, showToastIfIncreased?:
...
 export const addAssetAction = (asset: Asset) => {
   return async (dispatch: Dispatch, getState: () => Object) => {
     const {
       assets: { data: assets },
+      accounts: { data: accounts },
     } = getState();
-    const updatedAssets = { ...assets, [asset.symbol]: { ...asset } };
 
+    const activeAccount = getActiveAccount(accounts);
+    const accountId = get(activeAccount, 'id', null);
```
jsidorenko:
```suggestion
    const accountId = getActiveAccountId(accounts);
```

src/actions/assetsActions.js
```diff
@@ -630,14 +642,44 @@ export const resetSearchAssetsResultAction = () => ({
   type: RESET_ASSETS_SEARCH_RESULT,
 });
 
-export const checkForMissedAssetsAction = (transactionNotifications: Object[]) => {
+const getSupportedTokensAction = (supportedAssets: Asset[], currentAssets: AssetsByAccount, account: Account) => {
+  const accountId = getAccountAddress(account);
```
jsidorenko:
```suggestion
  const accountId = getActiveAccountId(account);
```

jsidorenko: or
```suggestion
  const accountAddress = getAccountAddress(account);
```

src/actions/assetsActions.js
```diff
@@ -647,43 +689,41 @@ export const checkForMissedAssetsAction = (transactionNotifications: Object[]) =
...
-          .reduce((memo, asset) => ({ ...memo, [asset.symbol]: asset }), {});
-        dispatch({
-          type: UPDATE_ASSETS,
-          payload: updatedAssets,
-        });
-        dispatch(saveDbAction('assets', { assets: updatedAssets }, true));
-      }
     }
 
+    const allSupportedAddedAssetsByAccount = await accounts.map((acc) => {
```
jsidorenko: No need for await here

src/actions/onboardingActions.js
```diff
@@ -172,9 +175,15 @@ const finishRegistration = async ({
     // create smart wallet account only for new wallets
     const createNewAccount = !isImported;
     await dispatch(initSmartWalletSdkAction(privateKey));
-    await dispatch(importSmartWalletAccountsAction(privateKey, createNewAccount));
+    await dispatch(importSmartWalletAccountsAction(privateKey, createNewAccount, initialAssets));
   }
 
+  const { accounts: { data: accounts } } = getState();
+
+  accounts.forEach(async (acc) => {
```
jsidorenko: This won't work :( you need to use 
`await Promise.all(accounts.map(acc => dispatch(restoreTransactionHistoryAction(acc.id, userInfo.walletId))))`

src/actions/smartWalletActions.js
```diff
@@ -529,7 +528,8 @@ export const fetchVirtualAccountBalanceAction = () => {
     if (!isConnectedToSmartAccount(connectedAccount) || !isOnline) return;
 
     const accountId = getActiveAccountId(accounts);
-    const ppnTokenAddress = getPPNTokenAddress(PPN_TOKEN, assets);
+    const accountAssets = assets[accountId];
```
jsidorenko: Selector?

src/actions/userSettingsActions.js
```diff
@@ -0,0 +1,78 @@
...
+import { saveDbAction } from './dbActions';
+
+export const hideAssetAction = (asset: Asset) => {
+  return async (dispatch: Dispatch, getState: GetState) => {
+    const {
+      accounts: { data: accounts },
+      userSettings: { data: { hiddenAssets } },
+    } = getState();
+
+    const accountId = getActiveAccountAddress(accounts);
```
jsidorenko: Same as above

src/models/Asset.js
```diff
@@ -30,10 +31,19 @@ export type Asset = {
   amount?: number,
 };
 
+
 export type Assets = {
   [symbol: string]: Asset,
 };
 
+export type AssetsByAccount = {
+  [string]: Assets,
```
jsidorenko:
```suggestion
  [accountId: string]: Assets,
```

src/reducers/userSettingsReducer.js
```diff
@@ -0,0 +1,57 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+import { SET_USER_SETTINGS, SET_HIDDEN_ASSETS } from 'constants/userSettingsConstants';
+
+type HiddenAssetsByAccount = {
+  [string]: String[],
```
jsidorenko:
```suggestion
  [accountId: string]: string[],
```

src/actions/assetsActions.js
vyvil: Hm, both of those methods return the same thing - id prop of account object:
getActiveAccountId
`const getActiveAccountId = (accounts: Accounts): string => {
  const activeAccount = getActiveAccount(accounts);
  if (!activeAccount) {
    return '';
  }
  return activeAccount.id;
};`

getActiveAccountAddress:
`export const getActiveAccountAddress = (accounts: Accounts): string => {
  const activeAccount = getActiveAccount(accounts);
  if (!activeAccount) {
    return '';
  }
  return getAccountAddress(activeAccount);
};`

where getAccountAddress is:
`const getAccountAddress = (account: Account): string => {
  return account.id;
};`

Do we need them both? It raises some confusion

src/actions/onboardingActions.js
vyvil: Ahh, true, thanks!

src/actions/assetsActions.js
vyvil: Cant use `getActiveAccountId` as it is used in the loop for all accounts

src/actions/assetsActions.js
jsidorenko: The primary idea was to separate id from address as they supposed to be different. But now I see that the address as an id worked out pretty well and there is no need to 2 same things under one name. We should unify it

src/actions/accountsActions.js
```diff
@@ -54,12 +55,7 @@ import { setActiveBlockchainNetworkAction } from './blockchainNetworkActions';
 const storage = Storage.getInstance('db');
 
 export const initDefaultAccountAction = (walletAddress: string, walletId: string, migrateData: boolean = true) => {
-  return async (dispatch: Dispatch) => {
-    const { balances = {} } = await storage.get('balances');
-    const { history = {} } = await storage.get('history');
-    const { collectibles = {} } = await storage.get('collectibles');
-    const { collectiblesHistory = {} } = await storage.get('collectiblesHistory');
-
+  return async (dispatch: Function) => {
```
grilix: Why are we removing the type here?

src/actions/assetsActions.js
```diff
@@ -630,14 +642,43 @@ export const resetSearchAssetsResultAction = () => ({
...
+
+  // HACK: Dirty fix for users who removed somehow ETH and PLR from their assets list
+  if (!currentAccountAssetsTickers.includes(ETH)) currentAccountAssetsTickers.push(ETH);
+  if (!currentAccountAssetsTickers.includes(PLR)) currentAccountAssetsTickers.push(PLR);
+
+  const updatedAccountAssets = supportedAssets
+    .filter(asset => currentAccountAssetsTickers.includes(asset.symbol))
+    .reduce((memo, asset) => ({ ...memo, [asset.symbol]: asset }), {});
+  return { id: accountId, ...updatedAccountAssets };
+};
```
grilix: This function is a bit complex, can we have a test to make sure it's doing what we expect it to do? Also, it doesn't really look like an action, I'd just drop the `Action` suffix.

src/actions/assetsActions.js
```diff
@@ -630,14 +642,43 @@ export const resetSearchAssetsResultAction = () => ({
...
+
+const getAllOwnedAssetsAction = (accountId: string, supportedAssets) => {
```
grilix: Can we type `supportedAssets`?

src/actions/assetsActions.js
```diff
@@ -630,14 +642,43 @@ export const resetSearchAssetsResultAction = () => ({
...
+    const addressErc20Tokens = await api.getAddressErc20TokensInfo(accountId); // all address' assets except ETH;
+    const accOwnedErc20Assets = {};
+    if (addressErc20Tokens.length) {
+      addressErc20Tokens.forEach((token) => {
+        const tokenTicker = get(token, 'tokenInfo.symbol', '');
+        const supportedAsset = supportedAssets.find(asset => asset.symbol === tokenTicker);
+        if (supportedAsset && !accOwnedErc20Assets[tokenTicker]) accOwnedErc20Assets[tokenTicker] = supportedAsset;
+      });
+    }
+    return accOwnedErc20Assets;
```
grilix: Just a question; Isn't it con to use the return of a `dispatch()` call? I feel like `dispatch()` calls should be "fire and forget". In this particular case, it doesn't seem like we need either `dispatch` or `getState`, so we don't even need to wrap this on a `dispatch()` call just to get the `api` object . What if instead of having

```js
const getAllOwnedAssetsAction = (...) => {
  return async (..., api: Object) {
    // ...
  };
};
```

We have just:

```js
const getAllOwnedAssetsAction = async (..., api: Object) => {
  // ...
};
```
..? The caller should pass the `api` object, but this function is gonna be a bit less obscure, and the caller doesn't have to wait for a `dispatch()`. Thoughts?

jsidorenko: In some exceptional situations, we could return something. But normally we should update the state and the `getAllOwnedAssetsAction()` caller should await for the method to be completed and get the updated state

jsidorenko:
```await dispatch(getAllOwnedAssetsAction());
const smth  = get(getState(), 'whatever');
```

grilix: Hmmm.. Yeah, I see your point, but I feel like adding an `await` to the `dispatch()` call goes against the redux design to be asynchronous. I understand exceptions can exist, but then exceptions should be there only when there's no other feasible way around it.

Not sure of other cases, but on this one it looks like this function doesn't even need to go through the dispatch queue.

jsidorenko: redux design to be asynchronous slightly conflicts with the real-life use-cases :D that's why someone created a redux-saga project ;)

regarding your suggestion - yes, we can remove the `return async (..., api: Object) {` and in the caller instead of writing `dispatch(getAllOwnedAssetsAction())` we would call as `dispatch(getAllOwnedAssetsAction)`. As for me, it looks nearly the same as it was before, although we'll have 99% of code written in one style plus that method written in slightly another style. I think that moment would create even more confusion for the future devs

grilix: > instead of writing dispatch(getAllOwnedAssetsAction()) we would call as dispatch(getAllOwnedAssetsAction)

My point is that we don't need the `dispatch` there. And it doesn't look like an action either, so we could remove the `Action` part from the name, even move it to another file if there's confusion. It would be just `const assets = await getAllOwnedAssets(accountId, supportedAssets, api)`.

jsidorenko: Just curious, how would you extract the api variable out of the redux middleware?

jsidorenko: sorry, just found that we're dispatching it not from the component, but from another action

jsidorenko: yes, then I agree, we can move it out from actions into services, remove the dispatch and pass the `api` as a param. Not a critical thing, mostly nice-to-have

jsidorenko: and then we could safely use the return statement, without breaking the redux `dispatch only` pattern

grilix: Indeed, not a critical thing :+1:

src/actions/accountsActions.js
vyvil: Sorry, that was made unintentionally. I've put it back :)

src/actions/assetsActions.js
vyvil: Done :)

src/actions/assetsActions.js
vyvil: Sure - done :)

src/actions/assetsActions.js
vyvil: Removed dispatch, but I've left this helper method in the same file - we are not reusing anywhere else

PR #1403, opened 2019-09-10
BugFix/sync contact identity keys

alazarevski:
Use updated identity mapping for contact info sync

### DIFF
--- a/src/actions/__tests__/contactsActions.test.js
+++ b/src/actions/__tests__/contactsActions.test.js
@@ -37,12 +37,14 @@ describe('Contacts Actions', () => {
       username: 'foobar',
       profileImage: 'foobar-image',
       ethAddress: 'eth-address-foobar',
+      status: undefined,
     },
     {
       id: 'user-lorem-ipsum',
       username: 'loremipsum',
       profileImage: 'loremipsum-image',
       ethAddress: 'eth-address-loremipsum',
+      status: undefined,
     },
   ];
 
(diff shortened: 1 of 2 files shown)


src/actions/__tests__/contactsActions.test.js
```diff
@@ -186,7 +194,7 @@ describe('Contacts Actions', () => {
       });
     });
 
-    it('should return and do nothing if accessToken does not exist', async () => {
+    xit('should return and do nothing if accessToken does not exist', async () => {
```
jsidorenko:
```suggestion
    it('should return and do nothing if accessToken does not exist', async () => {
```

src/actions/contactsActions.js
```diff
@@ -77,16 +77,25 @@ export const syncContactAction = (userId: string) => {
...
-    const accessToken = accessTokens.find(token => token.userId === userId);
-    if (!accessToken) return;
-
-    const userInfo = await api.userInfoById(userId, {
-      walletId,
-      userAccessKey: accessToken.myAccessToken,
-      targetUserAccessKey: accessToken.userAccessToken,
-    });
+    let userInfo;
+    const connection = connections.find((conn: ConnectionIdentityKey) => conn.targetUserId === userId);
```
jsidorenko: This action suppose to call the api in order to sync the userinfo, we shouldn't just grab the cached data from the connections

alazarevski: the data of connectionIdentityKeys is updated on every login or people screen refresh, the backend sends the map with the latest info of the contacts.

src/actions/__tests__/contactsActions.test.js
alazarevski: This test needs updating since accesskeys will be removed (or invalid in new_connections)

src/actions/contactsActions.js
jsidorenko: 👍

PR #1404, opened 2019-09-10
BugFix/disable biometrics on keygen

alazarevski:
Do not show biometrics when keys need to be generated

### DIFF
--- a/src/actions/authActions.js
+++ b/src/actions/authActions.js
@@ -45,6 +45,7 @@ import { LOG_OUT } from 'constants/authConstants';
 import { UPDATE_APP_SETTINGS } from 'constants/appSettingsConstants';
 import { UPDATE_SESSION } from 'constants/sessionConstants';
 import { BLOCKCHAIN_NETWORK_TYPES } from 'constants/blockchainNetworkConstants';
+import { PRE_KEY_THRESHOLD } from 'configs/connectionKeysConfig';
 import { delay } from 'utils/common';
 import Storage from 'services/storage';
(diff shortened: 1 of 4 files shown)


src/actions/authActions.js
```diff
@@ -99,7 +99,7 @@ export const loginAction = (
     const { wallet: encryptedWallet } = await storage.get('wallet');
     const { oAuthTokens } = await storage.get('oAuthTokens');
 
-    const generateNewConnKeys = !(connectionKeyPairs.length > 20 && lastConnectionKeyIndex > -1);
+    const generateNewConnKeys = connectionKeyPairs.length <= 20 || lastConnectionKeyIndex === -1;
```
jsidorenko: 👍 not it looks more human-readable

src/screens/PinCodeUnlock/PinCodeUnlock.js
```diff
@@ -108,9 +108,9 @@ class PinCodeUnlock extends React.Component<Props, State> {
   };
 
   showBiometricLogin() {
-    const { loginWithPrivateKey } = this.props;
+    const { loginWithPrivateKey, connectionKeyPairs: { data: connKeys, lastConnectionKeyIndex } } = this.props;
     const { biometricsShown } = this.state;
-    if (biometricsShown) return;
+    if (biometricsShown || connKeys.length <= 20 || lastConnectionKeyIndex === -1) return;
```
jsidorenko: It's better to extract this magic number 20 from here and the file above into some constant

PR #1419, opened 2019-09-23
Switch to new smart accounts sdk

jsidorenko:


src/services/smartWallet.js
```diff
@@ -293,13 +294,15 @@ class SmartWallet {
...
-  getDeployEstimate(gasPrice: BigNumber) {
-    /**
-     * can also call `this.sdk.estimateAccountDeployment(REGULAR);`,
-     * but it needs sdk init and when migrating we don't have SDK initiated yet
-     * so we're using calculation method below that is provided by SDK creators
-     */
-    return utils.bigNumberify(790000).mul(gasPrice);
+  async estimateAccountDeployment(gasInfo: GasInfo) {
+    const deployEstimate = await this.sdk.estimateAccountDeployment().catch(() => {});
+    const gasAmount = get(deployEstimate, 'gasFee', 790000);
```
jsidorenko: `const gasAmount = get(deployEstimate, 'gasFee', new BigNumber(790000));`
to keep consistency

src/services/smartWallet.js
```diff
@@ -171,7 +140,15 @@ class SmartWallet {
     const deployEstimate = await this.sdk.estimateAccountDeployment().catch(this.handleError);
 
     const accountBalance = this.getAccountRealBalance();
-    if (get(deployEstimate, 'totalCost') && accountBalance.gte(deployEstimate.totalCost)) {
+    if (get(deployEstimate, 'gasFee')
+      && get(deployEstimate, 'signedGasPrice.gasPrice')
+      && accountBalance.gte(deployEstimate.gasFee.mul(deployEstimate.signedGasPrice.gasPrice))
```
grilix: Would it make sense to extract this into a function?

grilix: Maybe it would just unnecessarily add extra complexity

poocart: Separated, please take a look and resolve if OK.

src/services/smartWallet.js
poocart: Updated, please resolve if OK.

src/services/smartWallet.js
```diff
@@ -56,6 +56,11 @@ type AccountTransaction = {
 
 let subscribedToEvents = false;
 
+const parseEstimatePayload = (estimatePayload) => ({
+  gasFee: get(estimatePayload, 'gasFee'),
+  gasPrice: get(estimatePayload, 'signedGasPrice.gasPrice'),
+});
```
jsidorenko: let's add total calculation here as well

src/services/smartWallet.js
```diff
@@ -56,6 +56,11 @@ type AccountTransaction = {
 
 let subscribedToEvents = false;
 
+const parseEstimatePayload = (estimatePayload) => ({
+  gasFee: get(estimatePayload, 'gasFee'),
```
jsidorenko: maybe it's better to rename gasFee to gasAmount?

PR #1420, opened 2019-09-23
Allow callbacks to be passed to Toast messages

grilix:
Toast messages will receive an `onPress` callback which, if passed, will
be called when the user presses on the Toast message text. This allows
making the Toast more useful.

The wallet backup message will now use the Toast callback to redirect
the user to the page where the backup functionality lives. The message
text itself was also updated to reflect the new location of the backup
feature.


src/utils/toasts.js
```diff
@@ -17,15 +17,32 @@
...
+
+  Toast.show({
+    message: BACKUP_MESSAGE,
+    type: 'warning',
+    title: 'Please ensure you backup your wallet now',
+    autoClose: false,
+    onPress: () => {
+      const action = NavigationActions.navigate({
+        routeName: WALLET_SETTINGS,
+        params: { wallet },
```
poocart: As we're trying to avoid wallet `privateKey` or `mnemonic` to not use where it's not needed I would highly suggest to remove those from `wallet` object or only pass here what's really needed in WalletSettings screen and it's only one param – `wallet.type`.

poocart: Or even better to pass already truncated wallet object (with type only) to this toast constructor.

jsidorenko: Pls update your code according to changes in this PR: https://github.com/pillarwallet/pillarwallet/pull/1436

grilix: Updated

src/actions/onboardingActions.js
```diff
@@ -203,15 +214,18 @@ const finishRegistration = async ({
   });
 };
 
-const navigateToAppFlow = (isWalletBackedUp: boolean, accountId: string) => {
+const navigateToAppFlow = (isWalletBackedUp: boolean, getState: GetState) => {
+  const accounts = getState().accounts.data;
+  const keyBasedAccount = findKeyBasedAccount(accounts);
+  const accountId = keyBasedAccount ? getAccountId(keyBasedAccount) : '';
```
poocart: Maybe `keyBasedAccount && getAccountId(keyBasedAccount)`? Because `getAccountId` will already return empty string if not found.

jsidorenko: getAccountId() requires an account to be passed, while findKeyBasedAccount() could return null

PR #1421, opened 2019-09-26
Feature/ppn withdrawal

jsidorenko:
This PR adds ability for PPN users to withdraw their balance


src/actions/smartWalletActions.js
```diff
@@ -961,6 +974,119 @@ export const topUpVirtualAccountAction = (amount: string) => {
...
+      .catch((e) => {
+        Toast.show({
+          message: e.toString(),
+          type: 'warning',
+          autoClose: false,
+        });
+        return {};
+      });
+
+    if (!estimated || !Object.keys(estimated).length) return;
```
poocart: if `isEmpty(estimated)` ?

PR #1431, opened 2019-09-27
Add api related typing

grilix:
This adds typing around the API SDKWrapper, which should help with the ongoing work for the API parametrization.

The only small functional change is to move the BCX instance inside the SDKWrapper to allow customization of the url used for creating said instance in the future.


src/actions/authActions.js
```diff
@@ -177,7 +179,7 @@ export const loginAction = (
           dispatch(setActiveBlockchainNetworkAction(BLOCKCHAIN_NETWORK_TYPES.ETHEREUM));
         }
       } else {
-        api.init();
+        api.init(null, undefined, null);
```
jsidorenko: looks very scary

grilix: Yeah, this is a bit shady. I've added a new parameter (coming on a following PR) which made flow go nuts. I'll try removing this from here and see how it goes. Will report back :+1:

grilix: Alright, it's not required. Removed.

PR #1432, opened 2019-09-27
[WIP] Allow user side network switching

grilix:
This is a currently non-working proposition for allowing users to switch
networks using two predefined networks: Main and Testing.

Network details can be configured using environment variables for each
one, then the user can switch from one network to another from the
settings screen. SystemInfoModal will show the current settings for the
selected network.

After the user selects a new network, the application must be restarted
to use the new settings.

Known issues with this implementation: Registration related information
is missing from the underlying backend, since the app will be using a
different instance. So user should complete the registration steps after
switching the network for the first time, but for that, we would need to
scope that information to each network and use the applicable depending
on the selected network.


grilix:
Due to the complexity and risk involved on this one, we're gonna put it on hold for now.

PR #1437, opened 2019-09-30
Fix/unsettled item to settled pending status

poocart:
PR includes:
- Fixes UI issues where status is not shown transactions that does not contain `itemValue` in `ListItemWithImage` component. 
- UI fix that was noticed before – if status is present for transaction then its icon was not vertically aligned inside list item.
- History sorting fix when PPN payment was being sorted ahead compared to its settled entry in history right after settlement was processed and this was because we were updating PPN payment create date every time when PPN state updates. Note: we map transaction history `createdAt` date top PPN payment `updatedAt` date because SDK doesn't provide `createdAt` value.

### DIFF
--- a/src/components/ActivityFeed/ActivityFeed.js
+++ b/src/components/ActivityFeed/ActivityFeed.js
@@ -63,7 +63,11 @@ import {
   TYPE_SENT,
 } from 'constants/invitationsConstants';
 import { COLLECTIBLE_TRANSACTION } from 'constants/collectiblesConstants';
-import { TRANSACTION_EVENT, CONNECTION_EVENT } from 'constants/historyConstants';
+import {
+  TRANSACTION_EVENT,
+  CONNECTION_EVENT,
+  TX_PENDING_STATUS,
+} from 'constants/historyConstants';
 import { CONTACT } from 'constants/navigationConstants';
 import { CHAT } from 'constants/chatConstants';
 import {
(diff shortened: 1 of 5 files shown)


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -300,6 +301,7 @@ class ActivityFeed extends React.Component<Props, State> {
     } = this.props;
 
     const navigateToContact = partial(navigation.navigate, CONTACT, { contact: notification });
+    const itemStatusIcon = notificationStatus === 'pending' ? 'pending' : '';
```
jsidorenko: we have Pending constant for tx status, let's use it here

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -326,15 +328,18 @@ class ActivityFeed extends React.Component<Props, State> {
...
       const tag = get(notification, 'tag', '');
       if (tag === PAYMENT_NETWORK_TX_SETTLEMENT) {
         return (
           <SettlementItem
             settleData={notification.extra}
-            onPress={() => this.selectEvent({ ...notification, value, contact }, type, notification.status)}
+            onPress={() => this.selectEvent({ ...notification, value, contact }, type, notificationStatus)}
             type={feedType}
             asset={asset}
+            isPending={notificationStatus === 'pending'}
```
jsidorenko: same as above

src/components/ActivityFeed/ActivityFeed.js
poocart: Didn't notice – was reusing the code, will surely update, thanks 👍

PR #1440, opened 2019-10-01
Feature/hide key wallet under collapse

vyvil:
This PR includes changes on Accounts screen:
1. Hides Key Wallet (that is now renamed to Legacy Wallet) under collapsible panel for non legacy users - those who started off with both: Key based and Smart wallet.
2. Unifies rendering of Accounts cards, as network cards are now the same as wallets cards.


src/services/api.js
```diff
@@ -840,3 +840,19 @@ SDKWrapper.prototype.fetchSendWyreOffers = function (fromAsset: string, toAsset:
...
+  const config = {
+    headers: {
+      Authorization: `Bearer ${accessToken}`,
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+    },
+  };
+  return Promise.resolve()
+    .then(() => fetch(`${SDK_PROVIDER}/user`, config))
+    .then(response => response.text())
```
grilix: Can we use `response.json()` here?

src/services/api.js
```diff
@@ -840,3 +840,19 @@ SDKWrapper.prototype.fetchSendWyreOffers = function (fromAsset: string, toAsset:
     })
     .catch(() => ({ error: true }));
 };
+
+SDKWrapper.prototype.getUserWallets = function (accessToken) {
```
grilix:
```suggestion
SDKWrapper.prototype.getUserWallets = function (accessToken: string) {
```


Not sure if we could also add the return type, like `Promise<Account[]>`

src/screens/Assets/Assets.js
```diff
@@ -103,6 +104,7 @@ type Props = {
   backupStatus: Object,
   availableStake: number,
   checkForMissedAssets: Function,
+  labelUserAsLegacy: Function,
```
grilix:
```suggestion
  labelUserAsLegacy: () => void,
```

src/actions/userActions.js
```diff
@@ -135,3 +136,37 @@ export const updateUserAvatarAction = (walletId: string, formData: any) => {
   };
 };
 
+export const labelUserAsLegacyAction = () => {
+  return async (dispatch: Dispatch, getState: GetState, api: Object) => {
```
grilix: If possible, add the type to the `api` object (we'd need to import the wrapper: `import SDKWrapper from 'services/api';`)

```suggestion
  return async (dispatch: Dispatch, getState: GetState, api: SDKWrapper) => {
```

src/actions/userActions.js
```diff
@@ -135,3 +136,37 @@ export const updateUserAvatarAction = (walletId: string, formData: any) => {
...
+    const {
+      oAuthTokens: { data: oAuthTokens },
+      user: { data: user },
+    } = getState();
+
+    const userWallets = await api.getUserWallets(oAuthTokens.accessToken);
+    if (!userWallets.length) return;
+
+    const keyWallet = userWallets.find(({ type }) => type === ACCOUNT_TYPES.KEY_BASED) || {};
+    const smartWallet = userWallets.find(({ type }) => type === ACCOUNT_TYPES.SMART_WALLET) || {};
```
grilix: I feel like we could avoid a couple of conditions and make this easier to understand by removing the `|| {}` and just checking if the wallets were found.

For example, it's not very intuitive what would happen if a smart wallet is found but not a key wallet. It looks like the only way to enter on `if (smartWalletCreationTime && diff && diff <= 5) {` is to have both a key wallet and a smart wallet, since the `diff` would be `null` if either is missing, right?

Assuming I understood this correctly, I'd suggest something like this:

```js
    const keyWallet = userWallets.find(({ type }) => type === ACCOUNT_TYPES.KEY_BASED);
    const smartWallet = userWallets.find(({ type }) => type === ACCOUNT_TYPES.SMART_WALLET);
    let isLegacyUser = true;

    if (keyWallet && smartWallet) {
      const { createdAt: keyWalletCreationTime } = keyWallet;
      const { createdAt: smartWalletCreationTime } = smartWallet;

      const diff = Math.floor((new Date(smartWalletCreationTime) - new Date(keyWalletCreationTime)) / 1000 / 60);

      // to those users who gets smart wallet created for them, key based and smart wallets are created one by one
      // in couple of minutes difference
      if (diff <= 5) {
        isLegacyUser = false;
      }
    }

    dispatch({
      type: UPDATE_USER,
      payload: { user: { ...user, isLegacyUser }, state: REGISTERED },
    });
```

src/screens/Accounts/Accounts.js
```diff
@@ -150,27 +153,41 @@ const IconWrapper = styled.View`
...
 
-  state = {
-    showPinModal: false,
-    changingAccount: false,
-  };
+  constructor(props) {
+    const isActiveKeyWallet = getActiveAccountType(props.accounts) === ACCOUNT_TYPES.KEY_BASED;
+    const { user } = props;
+    const forceShowLegacyWallet = !user.isLegacyUser && isActiveKeyWallet;
+    super(props);
```
jsidorenko: This should always go first

src/services/api.js
vyvil: Sure :)

src/services/api.js
vyvil: I'll check and update typing

src/screens/Assets/Assets.js
vyvil: Will update, thanks :)

src/actions/userActions.js
vyvil: Well, you can't have a smart wallet if you do not have a key based one :)

src/screens/Accounts/Accounts.js
vyvil: Shoot, yeah. I'll fix that. Thanks :)

src/screens/Accounts/Accounts.js
```diff
@@ -416,31 +393,87 @@ class AccountsScreen extends React.Component<Props, State> {
...
+    const { showPinModal, changingAccount, isLegacyWalletVisible } = this.state;
+    const { blockchainNetworks, user } = this.props;
+    const { isLegacyUser } = user;
 
     const activeNetwork = blockchainNetworks.find((net) => net.isActive);
     const walletsToShow = this.wallets(activeNetwork);
     const networksToShow = this.networks();
 
+    const smartAccountCard = walletsToShow
+      .find(({ type, isSmartWallet }) => !!isSmartWallet || type === 'NEW_SMART_WALLET');
```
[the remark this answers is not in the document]
vyvil: Will do

src/actions/userActions.js
vyvil: But yeah, I've revisited this part, and I agree that it could be clearer. I'll push an update in a minute

src/services/api.js
```diff
@@ -840,3 +841,19 @@ SDKWrapper.prototype.fetchSendWyreOffers = function (fromAsset: string, toAsset:
     })
     .catch(() => ({ error: true }));
 };
+
+SDKWrapper.prototype.getUserWallets = function (accessToken: string): Promise<Accounts> {
```
jsidorenko: we already have api method for this: `listAccounts()`

vyvil: Thanks, nor Meggie, nor I didn't know that ;/
Didn't see this on swagger either.
I've reuse it here and will push a change shortly

PR #1458, opened 2019-10-08
smart wallet transaction estimate improvements

poocart:
Includes:
- We were calculating gas estimate with key based methods even when assets are sent from Smart Wallet SDK and this caused failed transactions once the transaction reaches SDK method to deliver as gas estimates were calculated differently (100% failure when full balance is sent).
- Added loading spinners to asset send amount screen for "Send All" (Smart Wallet SDK estimate takes a bit longer to provide result) and "Submit" (once submit is pressed) buttons.
- Fixed blocking send button if there are errors on amount input.


src/screens/SendToken/SendTokenAmount.js
```diff
@@ -187,57 +204,84 @@ class SendTokenAmount extends React.Component<Props, State> {
...
   };
 
-  handleFormSubmit = () => {
+  handleFormSubmit = async () => {
+    const { submitPressed } = this.state;
+    if (submitPressed) return;
     this.formSubmitted = true;
-    const txFeeInWei = this.getTxFeeInWei();
+    this.setState({ submitPressed: true });
+    const txFeeInWei = await this.getTxFeeInWei();
```
grilix: Don't we have this on the state already?

src/services/smartWallet.js
```diff
@@ -79,6 +81,24 @@ export const parseEstimatePayload = (estimatePayload: EstimatePayload): ParsedEs
...
+
+  if (!gasAmount) {
+    gasAmount = new BigNumber(defaultGasAmount);
+  }
+  if (!gasPrice) {
+    const defaultGasPrice = get(gasInfo, `gasPrice.${speed}`, 0);
+    gasPrice = utils.parseUnits(defaultGasPrice.toString(), 'gwei');
+  }
+  return gasPrice.mul(gasAmount);
+};
```
grilix: Can we add some tests for this function considering the scenarios it tries to cover?

src/services/smartWallet.js
```diff
@@ -313,16 +332,26 @@ class SmartWallet {
...
+      value,
+      data,
+      transactionSpeed = TransactionSpeeds[AVG],
+    } = transaction;
+    const estimatedTransaction = await this.sdk.estimateAccountTransaction(
+      recipient,
+      value,
+      data,
+      transactionSpeed,
+    ).catch(() => {});
```
grilix: It looks like `calculateEstimate` should receive an `EstimatePayload`, maybe we can return if we know the sdk call failed instead of passing an empty object.

src/screens/SendToken/SendTokenAmount.js
poocart: We want to get it according latest price this is why we were doing the same with previous calculation. It can may change every second and holding user for 1s on this purpose is fine.

src/services/smartWallet.js
```diff
@@ -313,16 +332,26 @@ class SmartWallet {
 
   async estimateAccountDeployment(gasInfo: GasInfo) {
     const deployEstimate = await this.sdk.estimateAccountDeployment().catch(() => {});
-    let { gasAmount, gasPrice } = parseEstimatePayload(deployEstimate);
+    return calculateEstimate(deployEstimate, gasInfo, SPEED_TYPES.FAST, 790000);
```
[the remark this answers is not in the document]
poocart: It was there before the changes, I'll put as as constant, thanks 👍

src/services/smartWallet.js
poocart: OK 👍

src/services/smartWallet.js
poocart: Well if if fails then it will return default values as defined on `calculateEstimate`.

src/services/smartWallet.js
poocart: Will add tests later once back next week, feel free to merge if else looks fine.

src/services/smartWallet.js
grilix: Yeah, but this way we can't add a type to the `estimate` parameter on `calculateEstimate`, right?

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -187,57 +204,84 @@ class SendTokenAmount extends React.Component<Props, State> {
...
+      const gasPrice = txFeeInWei.div(gasLimit).toNumber();
+      transactionPayload = {
+        ...transactionPayload,
+        gasPrice,
+        gasLimit,
+        txSpeed: transactionSpeed,
+      };
+    }
     Keyboard.dismiss();
+    this.setState({ submitPressed: false }); // reset in case of user navigates back
```
jsidorenko: it's possible that setState would start after we navigate the user to another screen thus causing an error. I would suggest to use `setState` callback:
```suggestion
this.setState({ submitPressed: false }, () => {
  navigation.navigate(SEND_TOKEN_CONFIRM, {
    transactionPayload,
    source: this.source,
  });
});
```

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -187,57 +204,84 @@ class SendTokenAmount extends React.Component<Props, State> {
...
 
   handleChange = (value: Object) => {
-    this.setState({ value });
-    this.updateGasLimit();
+    // first update the amount, then after state is updated check for errors
+    this.setState({ value, gettingFee: true });
+    this.getTxFeeInWei()
+      .then(txFeeInWei => this.setState({ txFeeInWei, gettingFee: false }))
+      .catch(() => null);
+    this.checkFormInputErrors();
```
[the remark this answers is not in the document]
poocart: Does not matter and adding input validation to happen after fee is fetched would even take more for validation message to appear on input.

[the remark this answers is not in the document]
poocart: Although will add small improvement here.

src/screens/SendToken/SendTokenAmount.js
poocart: 👍

src/screens/SendToken/SendTokenAmount.js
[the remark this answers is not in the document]
poocart: Added improvements, please review and resolve if all good

PR #1459, opened 2019-10-09
Feature/modals

vyvil:
This PR includes changes on legal info and event modals that
provides a way to add scrollable content to them and still allow to close those modals by swiping down when top of the scrollable content is reached.

### DIFF
--- a/src/components/EventDetails/EventDetails.js
+++ b/src/components/EventDetails/EventDetails.js
@@ -21,7 +21,7 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import isEqual from 'lodash.isequal';
 import type { NavigationScreenProp } from 'react-navigation';
-import { Linking, Dimensions, ScrollView } from 'react-native';
+import { Linking, Dimensions, ScrollView, Clipboard } from 'react-native';
 import styled from 'styled-components/native';
 import { utils } from 'ethers';
 import { TX_DETAILS_URL } from 'react-native-dotenv';
@@ -83,6 +83,7 @@ import { accountAssetsSelector } from 'selectors/assets';
 
 // local components
 import EventHeader from './EventHeader';
+import Toast from '../Toast';
 
 type Props = {
   transaction: Transaction,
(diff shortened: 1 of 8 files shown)


src/components/EventDetails/EventDetails.js
```diff
@@ -364,80 +367,57 @@ class EventDetails extends React.Component<Props, {}> {
...
+            onPress={Object.keys(relatedUser).length ? null : () => this.writeToClipboard(relatedUserTitle)}
+          />
+          }
+          {listSettledAssets &&
+          <ListItemUnderlined
+            label="ASSETS"
+            value={extra.map(item => <BoldText key={item.hash}> {item.value} {item.symbol}</BoldText>)}
+          />
+          }
+          {(toMyself || !isReceived) && !isPending && (freeTx || !!fee) &&
```
jsidorenko: Let's extract this to some variable like `showFeeBlock`

src/components/EventDetails/EventDetails.js
```diff
@@ -103,22 +105,36 @@ type Props = {
   contactsSmartAddresses: ContactSmartAddressData[],
   inactiveAccounts: Accounts,
   supportedAssets: Asset[],
+  getRef?: () => Object,
+  getScrollOffset?: (number) => ScrollToProps,
+  getMaxScrollOffset?: (number) => number,
 }
 
+type State = {
+  containerHeight?: number,
```
jsidorenko: I don't think it's a good idea (react wise) to have optional state params as this could drive crazy react optimiser. Could we change this to `containerHeight: ?number,`?

src/components/Modals/HTMLContentModal/HTMLContentModal.js
```diff
@@ -36,6 +37,9 @@ type Props = {
 type State = {
   isHtmlFetched: boolean,
   htmlData: string,
+  scrollOffset?: ?number,
```
jsidorenko: Same a above

src/components/EventDetails/EventDetails.js
vyvil: Updated :)

PR #1462, opened 2019-10-10
Allow viewing collectible images on full screen

grilix:
Allows collectible images to be shown on full screen, with the ability to zoom it in.

image

image

image


src/screens/Collectible/Collectible.js
```diff
@@ -111,6 +115,12 @@ class CollectibleScreen extends React.Component<Props> {
     this.props.navigation.navigate(SEND_COLLECTIBLE_FROM_ASSET_FLOW, { assetData });
   };
 
+  onTouchImage = () => {
+    const { navigation: { state: { params: { assetData } } } } = this.props;
```
jsidorenko:
```suggestion
    const { navigation: { navigate, state: { params: { assetData } } } } = this.props;
```

src/screens/Collectible/Collectible.js
```diff
@@ -111,6 +115,12 @@ class CollectibleScreen extends React.Component<Props> {
...
+
+    this.props.navigation.navigate(FULL_SCREEN_COLLECTIBLE, { collectible: assetData });
```
jsidorenko:
```suggestion
    navigate(FULL_SCREEN_COLLECTIBLE, { collectible: assetData });
```

src/screens/Collectible/FullScreenCollectible.js
```diff
@@ -0,0 +1,79 @@
...
+
+const genericCollectible = require('assets/images/no_logo.png');
+
+class FullScreenCollectible extends React.PureComponent<Props> {
+  collectible: Collectible;
+
+  constructor(props: Props) {
+    super(props);
+
+    const { navigation: { state: { params: { collectible } } } } = props;
```
jsidorenko:
```suggestion
    const collectible = props.navigation.getParam('collectible')
```

src/screens/Collectible/FullScreenCollectible.js
```diff
@@ -0,0 +1,79 @@
...
+  justify-content: center;
+  min-height: 300px;
+  min-width: 300px;
+  align-self: center;
+`;
+
+const genericCollectible = require('assets/images/no_logo.png');
+
+class FullScreenCollectible extends React.PureComponent<Props> {
+  collectible: Collectible;
```
jsidorenko: Just wondering is it allowed to use internal variables for PureComponents?

src/screens/Collectible/Collectible.js
grilix: Damn, they need to add multi-line suggestions :sweat_smile:

src/screens/Collectible/FullScreenCollectible.js
grilix: It looks like it is :smile:! Good catch, didn't realize I did this.

grilix:
@jsidorenko thanks for the comments; it turns out we don't even need a new screen since `react-native-image-view` shows the image on a modal, which makes everything simpler and gives the user a better experience.

Please let me know your thoughts.

src/screens/Collectible/Collectible.js
```diff
@@ -97,20 +103,58 @@ const CollectibleImage = styled(CachedImage)`
...
+  };
+
+  shouldComponentUpdate(nextProps: Props, nextState: State) {
     const isFocused = this.props.navigation.isFocused();
     if (!isFocused) {
       return false;
     }
-    const isEq = isEqual(this.props, nextProps);
-    return !isEq;
+    if (!isEqual(this.state, nextState)) {
```
vyvil: Those could be combined: 
```suggestion
    if (!isEqual(this.state, nextState) || !isEqual(this.props, nextProps)) return true;
```

src/screens/Collectible/Collectible.js
```diff
@@ -97,20 +103,58 @@ const CollectibleImage = styled(CachedImage)`
...
+        title: name,
+      },
+    ];
+
+    return (
+      <ImageView
+        images={imageViewImages}
+        imageIndex={0}
+        isVisible={isImageViewVisible}
+        onClose={this.onCloseImageView}
```
vyvil: Would be great to add close icon that we use on other modals. There's a way to add custom close component using controls prop: https://www.npmjs.com/package/react-native-image-view#controls-prop

src/screens/Collectible/Collectible.js
grilix: Updated

src/screens/Collectible/Collectible.js
grilix: Updated

PR #1464, opened 2019-10-15
Add Bitcoin network to accounts list

grilix:
This adds the Bitcoin network option for accounts list as an invisible placeholder.


src/screens/Accounts/Accounts.js
```diff
@@ -389,13 +406,35 @@ class AccountsScreen extends React.Component<Props, State> {
...
+    );
     const hasAccount = userHasSmartWallet(accounts);
     const hasSmartWallet = hasAccount && smartWalletFeatureEnabled;
 
+    const bitcoinNetwork = blockchainNetworks.find(
+      (network) => network.isAvailable && network.id === BLOCKCHAIN_NETWORK_TYPES.BITCOIN,
+    );
+
+    if (bitcoinNetwork) {
+      networks.push({
```
vyvil: Bitcoin card should go after Pillar Network card :)

grilix: Updated

grilix:
@vyvil I'm not sure of where to add the `TODO` comment, really. I've added it on the reducer, since it's the next place we'll have to change and we'll see it. We can move it as we go, or if you have a better idea please let me know.

PR #1467, opened 2019-10-17
sort user accounts to top in send token contact select screen

poocart:
Fixes issue where user accounts are not sorted to top on send token contact select screen.


src/screens/SendToken/SendTokenContacts.js
```diff
@@ -302,6 +299,11 @@ class SendTokenContacts extends React.Component<Props, State> {
         });
     }
 
+    // asset transfer between user accounts only in regular, but not in PPN send flow
+    if (!this.isPPNTransaction) {
+      contactsToRender = [...userAccounts, ...contactsToRender];
```
jsidorenko: That means even if I search for the contact username, I will still see my account on top in search results. Wouldn't be that confusing?

poocart: Updated 👍

PR #1475, opened 2019-10-18
Fix collectible image closing on iPhone X

grilix:
Fixes an issue where the close icon is not clickable on iPhone X.

# iPhone X

collectible-ipX

# iPhone 7

collectible-ip7


src/screens/Collectible/Collectible.js
```diff
@@ -98,6 +99,8 @@ const CollectibleImage = styled(CachedImage)`
 `;
 
 const IconWrapper = styled.View`
+  margin-top: ${isIphoneX() ? '30px' : '0'};
```
poocart: For me would look nicer `${isIphoneX() && 'margin-top: 30px'};`

PR #1478, opened 2019-10-21
Make assets sorting more uniform

grilix:
- Store supported assets sorted, so we don't have to sort it everywhere.
- Expose asset sorting functions on utils/assets: sortAssets and sortAssetsArray.
- Remove redundant sorting on Exchange.


src/utils/assets.js
```diff
@@ -30,6 +30,20 @@ import get from 'lodash.get';
...
+const sortAssetsFn = (a: Asset, b: Asset): number => {
+  return a.symbol.localeCompare(b.symbol);
+};
+
+export const sortAssetsArray = (assets: Asset[]): Asset[] => {
+  return assets.sort(sortAssetsFn);
+};
+
+export const sortAssets = (assets: Assets): Asset[] => {
+  const assetsList = Object.keys(assets).map((key: string) => assets[key]);
```
jsidorenko: I think you can use `getAssetsAsList` here

grilix: Didn't realise that. Updated.

jsidorenko: no worries, it's a new thing ;)

PR #1481, opened 2019-10-23
Feature/wallets icons

vyvil:
This PR adds user accounts / wallets icons:
1. On Account cards (changes Smart wallet's icon into new one).
2. On user wallet item in contacts (receivers) list on asset send flow.

It also lists user's wallet as a receiver in transaction confirmation screen when transferring to one of user's wallets.


src/screens/SendCollectible/SendCollectibleConfirm.js
```diff
@@ -156,6 +165,7 @@ class SendCollectibleConfirm extends React.Component<Props, State> {
     const contact = contacts.find(({ ethAddress }) => to.toUpperCase() === ethAddress.toUpperCase());
     const recipientUsername = getUserName(contact);
     const canProceedTesting = parseFloat(rinkebyETH) > parseFloat(txFee) || NETWORK_PROVIDER !== 'ropsten';
+    const userAccount = !recipientUsername ? accounts.find(({ id }) => addressesEqual(id, to)) : null;
```
poocart: How about just:
```
const userAccount = !recipientUsername && accounts.find(({ id }) => addressesEqual(id, to));
```

jsidorenko: @poocart what do you think will be the result of your suggestion if the `recipientUsername` will be not empty?

poocart: 1. `recipientUsername` empty (1) and account exists (2): true && account = account
2. `recipientUsername` empty (1) and account does not exist (2): true && undefined = undefined
3. `recipientUsername` not  empty (1) and account exists (2): false && account = false
4. `recipientUsername` not  empty (1) and account does not exist (2): false && undefined = undefined

Answer: undefined which later becomes false with !! and prevents from rendering.

https://jsbin.com/lidafowumu/edit?js,console.

jsidorenko: Right, 10 points ;)
In Vy's approach, we're dealing with `null` and `account`, in yours with `false`, `null` and `account`. The suggested change would add an additional point of failure if we decide to extend the code in the future

PR #1482, opened 2019-10-25
Fix/sentry issues 10 24

poocart:
Fixes couple issues:
- Do not crash app if by any mean account comes as undefined object (Sentry issue).
- Reload Assets screen even if not focused, but active account prop has changed – this was an issue on faster devices that caused Assets screen to stay as Legacy wallet after swtiching from it to Smart Wallet or PPN screens.


src/screens/Assets/Assets.js
```diff
@@ -152,8 +152,9 @@ class AssetsScreen extends React.Component<Props, State> {
   }
 
   shouldComponentUpdate(nextProps: Props, nextState: State) {
+    const { activeAccount } = this.props;
     const isFocused = this.props.navigation.isFocused();
-    if (!isFocused) {
+    if (!isFocused && isEqual(nextProps.activeAccount, activeAccount)) {
```
jsidorenko: It's faster to compare account ids rather than 2 objects here

src/screens/Accounts/Accounts.js
```diff
@@ -475,6 +499,9 @@ class AccountsScreen extends React.Component<Props, State> {
 
     const walletsInList = isLegacyUser ? walletsToShow : [smartAccountCard];
 
+    // in case if undefined object comes to array (ref: Sentry issue ID 1292819194)
+    const accountsList = [...walletsInList, ...networksToShow].filter(item => !isEmpty(item));
```
jsidorenko: How could you get `undefined` here?

src/screens/Accounts/Accounts.js
```diff
@@ -39,24 +43,46 @@ import { ScrollWrapper } from 'components/Layout';
...
+  Rates,
+} from 'models/Asset';
+import type {
+  Account,
+  Accounts,
+} from 'models/Account';
+import type {
+  Dispatch,
+  RootReducerState,
+} from 'reducers/rootReducer';
```
jsidorenko: I think it became less readable now

src/screens/Accounts/Accounts.js
poocart: As per PR description - we're not sure yet. If we put undefined here manually we get exactly same crash as in Sentry, the issue itself is yet to debug.

src/screens/Accounts/Accounts.js
poocart: Well, TBH it was auto lint fixes, but will revert to one line imports.

src/screens/Accounts/Accounts.js
poocart: @vyvil finally found an issue, will make another commit re this

PR #1483, opened 2019-10-25
Update to new Wyre widget

grilix:
This uses the new SendWyre widget version to allow Apple Pay and Google Pay.

The widget is required to be show inside a browser tab instead of using the WebView, the package `@matt-block/react-native-in-app-browser` provides that.


src/services/sendwyre.js
```diff
@@ -0,0 +1,45 @@
...
+export const wyreWidgetUrl = (
+  destAddress: string,
+  destCurrency: string,
+  sourceCurrency: string,
+  sourceAmount: string,
+): string => {
+  const url =
+    `${SENDWYRE_WIDGET_URL}?destCurrency=${
+      destCurrency
+    }&dest=ethereum:${
```
jsidorenko: looks cool, but unreadable

jsidorenko:
Please try to follow the project standards:
1) every file should have a disclaimer in the beginning
2) import of external libraries goes first
3) branch name starts from `feature/` or `fix/`

src/services/sendwyre.js
grilix: Damn, yeah. This was the easiest for testing adding/removing parameters and got used to it so I left it. Will update.

PR #1486, opened 2019-10-28
Preparation for tx history refactoring part 2

jsidorenko:
The second part of the upcoming tx history refactoring PR.
Added getAssetsAsList() util


src/utils/assets.js
```diff
@@ -37,6 +37,10 @@ export const transformAssetsToObject = (assetsArray: Asset[] = []): Assets => {
   }, {});
 };
 
+export const getAssetsAsList = (assetsObject: Assets = {}): Asset[] => {
```
grilix: Why is this argument optional? It looks like every place already used directly `Object.keys(assets)` without checking anything, can't we just do the same here?

jsidorenko: Sure thing!

PR #1488, opened 2019-10-29
Feature/synthetics

poocart:
This includes full flow for sending synthetic asset.


poocart:
Two concerns for WIP:
1. Change Send Synthetic Asset select screen into two: .
2. Check liquidity pools from synthetics service and filter out unavailable assets.

src/actions/smartWalletActions.js
```diff
@@ -830,8 +830,9 @@ export const onSmartWalletSdkEventAction = (event: Object) => {
       const { decimals = 18 } = accountAssets[PPN_TOKEN] || {};
       const txAmountFormatted = formatUnits(txAmount, decimals);
 
-      if (activeAccountAddress === txReceiverAddress
-        && txReceiverAddress !== txSenderAddress
+      // check if received transaction
+      if (isCaseInsensitiveMatch(activeAccountAddress, txReceiverAddress)
```
jsidorenko: We have addressEqual function for this

src/screens/SendSynthetics/SendSyntheticsConfirm.js
```diff
@@ -0,0 +1,233 @@
...
+`;
+
+class ExchangeConfirmScreen extends React.Component<Props, State> {
+  syntheticTransaction: SyntheticTransaction;
+  assetData: Asset;
+  state: State = { note: null };
+
+  constructor(props) {
+    super(props);
+    const { navigation } = this.props;
```
jsidorenko:
```suggestion
    const { navigation } = props;
```

src/screens/SendSynthetics/SendSyntheticsSelect.js
```diff
@@ -0,0 +1,373 @@
...
+      || !Object.keys(selector).length
+      || !input
+      || !isValidNumber(input)
+      || !!intentError) return false;
+
+    const { decimals } = selector;
+
+    amount = parseFloat(input);
+
+    return !(decimals === 0 && amount.toString().indexOf('.') > -1);
```
jsidorenko:
```suggestion
    return decimals !== 0 || !amount.toString().includes('.'));
```

src/screens/SendSynthetics/SendSyntheticsSelect.js
```diff
@@ -0,0 +1,373 @@
...
+      return 'Incorrect number entered.';
+    }
+
+    if (!Object.keys(selector).length) {
+      return 'Asset should be selected.';
+    } else if (!input) {
+      return false; // should still validate (to not trigger search if empty), yet error should not be visible to user
+    } else if (parseFloat(input) < 0) {
+      return 'Amount should be bigger than 0.';
+    } else if (decimals === 0 && amount.toString().indexOf('.') > -1) {
```
jsidorenko:
```suggestion
    } else if (decimals === 0 && amount.toString().includes('.')) {
```

src/screens/SendSynthetics/SendSyntheticsSelect.js
```diff
@@ -0,0 +1,373 @@
...
+`;
+
+class SendSyntheticsSelect extends React.Component<Props, State> {
+  syntheticsForm: t.form;
+  fromInputRef: TextInput;
+  receiver: string;
+
+  constructor(props: Props) {
+    super(props);
+    this.receiver = this.props.navigation.getParam('receiver', '');
```
jsidorenko:
```suggestion
    this.receiver = props.navigation.getParam('receiver', '');
```

src/screens/SendSynthetics/SendSyntheticsSelect.js
```diff
@@ -0,0 +1,373 @@
...
+      const validation = this.syntheticsForm.validate();
+      const { errors = [] } = validation;
+      const assetCode = get(value, 'fromInput.selector.symbol');
+      if (errors.length || !assetCode) {
+        this.setState({ submitPressed: false });
+        return;
+      }
+      const amount = parseNumericAmount(value);
+      const { navigation, assets, supportedAssets } = this.props;
+      const assetsData = Object.keys(assets).map(id => assets[id]);
```
jsidorenko: Use getAssetsAsList()

src/services/synthetics.js
```diff
@@ -0,0 +1,92 @@
...
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+import { SYNTHETICS_URL } from 'react-native-dotenv';
+
+type SyntheticsConfig = {
+  accessToken: string,
+}
```
jsidorenko:
```suggestion
};
```

src/screens/SendSynthetics/SendSyntheticsSelect.js
poocart: This branch didn't have it yet, will update

src/actions/syntheticsActions.js
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+    syntheticsService.init({ accessToken });
+  };
+};
+
+export const commitSyntheticsTransaction = (transactionId: string, paymentHash: string) => {
+  return (dispatch: Dispatch) => {
+    dispatch(initSyntheticsServiceAction());
+    console.log('syntheticTransactionId: ', transactionId);
```
grilix: Do we need this?

src/screens/SendSynthetic/SendSyntheticAmount.js
```diff
@@ -0,0 +1,382 @@
...
+import { accountAssetsSelector } from 'selectors/assets';
+
+// models
+import type { Asset, Assets, Rates } from 'models/Asset';
+import type { SyntheticTransaction, TokenTransactionPayload } from 'models/Transaction';
+
+type Props = {
+  accountAssets: Assets,
+  supportedAssets: Asset[],
+  initSyntheticsService: Function,
```
grilix:
```suggestion
  initSyntheticsService: () => void,
```

src/screens/SendSynthetic/SendSyntheticAmount.js
```diff
@@ -0,0 +1,382 @@
...
+class SendSyntheticAsset extends React.Component<Props, State> {
+  syntheticsForm: t.form;
+  receiver: string;
+  source: string;
+  assetData: Asset;
+  assetBalance: number;
+
+  constructor(props: Props) {
+    super(props);
+    this.source = props.navigation.getParam('source', '');
```
grilix:
```suggestion
    const { navigation: { getParam } } = props;

    this.source = getParam('source', '');
```

src/screens/SendSynthetic/SendSyntheticAmount.js
```diff
@@ -0,0 +1,382 @@
...
+  submitPressed: boolean,
+  intentError: ?string,
+  inputHasError: boolean,
+};
+
+const { Form } = t.form;
+
+const generateFormStructure = (intentError: ?string, maxAmount: number, decimals: number) => {
+  const Amount = t.refinement(t.String, (amount): boolean => {
+    if (!isValidNumber(amount.toString()) || !!intentError) return false;
```
grilix: I don't think enforcing boolean values is necessary on `if` conditions, we can avoid that extra noise.

```suggestion
    if (!isValidNumber(amount.toString()) || intentError) return false;
```

src/screens/SendSynthetic/SendSyntheticAmount.js
```diff
@@ -0,0 +1,382 @@
...
+    );
+  }
+}
+
+const mapStateToProps = ({
+  assets: { supportedAssets },
+  rates: { data: rates },
+  appSettings: { data: { baseFiatCurrency } },
+  session: { data: { isOnline } },
+}) => ({
```
grilix: Allow linter detect some issues here:

```suggestion
}: RootReducerState): $Shape<Props> => ({
```

We're also gonna need to add

```js
import { RootReducerState } from 'reducers/rootReducer';
```

src/screens/SendSynthetic/SendSyntheticAmount.js
```diff
@@ -0,0 +1,382 @@
...
+const structuredSelector = createStructuredSelector({
+  accountAssets: accountAssetsSelector,
+});
+
+const combinedMapStateToProps = (state) => ({
+  ...structuredSelector(state),
+  ...mapStateToProps(state),
+});
+
+const mapDispatchToProps = (dispatch: Function) => ({
```
grilix:
```suggestion
const mapDispatchToProps = (dispatch: Dispatch) => ({
```

src/utils/common.js
```diff
@@ -380,3 +380,25 @@ export const groupAndSortByDate = (
...
+    const cancelledErrorPayload = { isCanceled: true };
+    promise
+      .then((res) => isCancelled
+        ? reject(cancelledErrorPayload)
+        : resolve(res),
+      )
+      .catch((err) => isCancelled
+        ? reject(cancelledErrorPayload)
+        : reject(err),
+      );
```
grilix: What about something like:

```js
if (isCancelled) {
  const cancel = () => ({ isCanceled: true });

  promise.then(cancel).catch(cancel);
} else {
  promise.then(resolve).catch(reject);
}
```

src/screens/SendSynthetic/SendSyntheticAsset.js
```diff
@@ -0,0 +1,239 @@
...
+              balance: assetBalance,
+              isSynthetic: true,
+            });
+          }
+          return availableList;
+        }, []);
+        this.setState({ availableAssets, loadingAssets: false });
+      })
+      .catch(({ isCanceled }) => {
+        if (!isCanceled) this.setState({ loadingAssets: false });
```
grilix: What's the difference here if it was cancelled or just failed? Don't we want to do this regardless of if this was cancelled or not?

poocart: The cancel part is because if it's triggered and we go back to previous screen and we don't want state to be updated in the component that is unmounted.

src/utils/common.js
poocart: Not really, thats the point that we want to reject cancel promise and in your suggestion it would wait for promise to be completed.

src/services/synthetics.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  async handleResponse(responseResult: any) {
+    const response = await responseResult.json();
+    if (response.statusCode === 200) return response;
+    return Promise.reject(response);
+  }
+}
+
+const syntheticsService = new SyntheticsService();
```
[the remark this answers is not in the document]
poocart: Yes, it's singleton, the only piece we want to update it for API calls is OAuth tokens so we don't need to pass those on each API request.

src/utils/common.js
poocart: Actually I see your point, this actually would work, updated, thanks 👍

poocart: I'm sorry, this approrach actually doesn't work so had to restore original one. The reason is that `isCanceled` is not yet updated and original callbacks are then built for promise which leads that promise cannot be "cancelled".

poocart: Please mark as resolved if everything makes sense.

src/services/synthetics.js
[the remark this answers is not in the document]
poocart: Just adding more explanation – we need to store OAuth tokens in the instance just to not get them from the state and then pass in each API call.

Please mark as resolved if everything makes sense.

src/screens/SendSynthetic/SendSyntheticAsset.js
poocart: Please mark as resolved if everything makes sense.

src/utils/common.js
grilix: Well, my point is that's really confusing to understand what is going on here. I think if we have an `if (..) { .. } else { .. }` it would be more straightforward even if it results on more lines.

src/models/Asset.js
```diff
@@ -29,9 +29,9 @@ export type Asset = {
   patternUrl?: string,
   decimals: number,
   amount?: number,
+  balance?: number,
```
jsidorenko: What's the difference between amount and balance?

src/screens/SendSynthetic/SendSyntheticAmount.js
```diff
@@ -0,0 +1,384 @@
...
+const { Form } = t.form;
+
+const generateFormStructure = (intentError: ?string, maxAmount: number, decimals: number) => {
+  const Amount = t.refinement(t.String, (amount): boolean => {
+    if (!isValidNumber(amount.toString()) || intentError) return false;
+
+    if (decimals === 0 && amount.toString().includes('.')) return false;
+    amount = parseFloat(amount.toString());
+
+    return (decimals !== 0 || !amount.toString().includes('.')) && amount <= maxAmount;
```
jsidorenko: You can extract `decimals === 0 && amount.toString().includes('.')` into variable like `const wrongDecimals = decimals === 0 && amount.toString().includes('.');` and then re-use it in ther return statement: `return !wrongDecimals && amount <= maxAmount;`

src/screens/SendSynthetic/SendSyntheticAmount.js
```diff
@@ -0,0 +1,384 @@
...
+  Amount.getValidationErrorMessage = (amount) => {
+    if (!isValidNumber(amount.toString())) {
+      return 'Incorrect number entered.';
+    }
+
+    amount = parseNumber(amount.toString());
+
+    if (amount > maxAmount) {
+      return 'Amount should not exceed the max available';
+    } else if (decimals === 0 && amount.toString().includes('.')) {
```
jsidorenko: Maybe we should extract this to separate function like `validateNumberDecimals`?

src/screens/SendSynthetic/SendSyntheticConfirm.js
```diff
@@ -0,0 +1,238 @@
...
+    const transactionPayload = {
+      amount: fromAmount,
+      to: SYNTHETICS_CONTRACT_ADDRESS,
+      symbol,
+      contractAddress,
+      decimals,
+      note,
+      usePPN: true,
+      extra: {
+        syntheticTransaction: this.syntheticTransaction,
```
jsidorenko:
```suggestion
        syntheticTransaction: { ...this.syntheticTransaction },
```

src/screens/SendSynthetic/SendSyntheticAsset.js
```diff
@@ -0,0 +1,240 @@
...
+  componentDidMount() {
+    this.props.initSyntheticsService();
+    this.getAvailableSyntheticsAssets();
+  }
+
+  componentWillUnmount() {
+    this.cancelSyntheticServiceRequest();
+  }
+
+  getAvailableSyntheticsAssets = () => {
```
jsidorenko: I think that part should be extracted into the action, as this is a business logic thing. Then you won't need a hack with cancelable promise

jsidorenko:
👍 nice&clean!

PR #1489, opened 2019-10-29
Feature/rework smart wallet transactions

jsidorenko:
In this PR:

- added new method `fetchSmartWalletTransactionsAction()` to fetch smart wallet transactions
- added method to process and store transactions' metadata for smart accounts
- created new `fetchTransactionsHistoryAction()` method to fetch transaction history depending on active account type
- improved tx history syncing algorithm for smart accounts
- reworked smart wallet notifications
- removed duplicate code from tx history fetch methods


src/actions/__tests__/historyActions.test.js
```diff
@@ -157,14 +160,15 @@ const transformedImportedPlrTransaction = buildHistoryTransaction({
 describe('History Actions', () => {
   const transactionsHistoryStep = 10;
 
-  const api = {
+  // $FlowFixMe
+  const api: SDKWrapper = {
```
grilix: Not sure, but maybe we can use `$Shape<SDKWrapper>` to avoid the `$FlowFixMe` if it's the case?

poocart: Good catch 👍

src/models/Transaction.js
```diff
@@ -24,7 +24,11 @@ export type TxSettlementItem = {
   hash: string,
 };
 
-export type TransactionExtra = TxSettlementItem[];
+export type TxWithdrawalExtra = {
```
grilix: I wouldn't export symbols by default, since linters can't help detect unused stuff for exported names. This one looks like is not being used outside this file.

src/screens/Asset/Asset.js
```diff
@@ -230,10 +231,10 @@ class AssetScreen extends React.Component<Props, State> {
   };
 
   handleScrollWrapperEndDrag = e => {
-    const { fetchAssetTransactions, history } = this.props;
-    const {
-      assetData: { token },
-    } = this.props.navigation.state.params;
+    const { fetchAssetTransactions, history, activeAccount } = this.props;
+    if (!activeAccount || checkIfSmartWalletAccount(activeAccount)) return;
```
grilix: The prop `activeAccount` is not marked as optional, I'd avoid the `!activeAccount` check (or make it optional if makes more sense).

src/services/smartWallet.js
```diff
@@ -301,17 +301,41 @@ class SmartWallet {
     return this.sdk.searchAccount({ address });
   }
 
-  async getAccountPayments(lastSyncedHash: ?string, page?: number = 0) {
+  /**
+   * SDK API call results are sorted descending by creation date
+   * lastSyncedId is used to determine whether this page was already fetched
+   */
+  async getAccountPayments(lastSyncedId: ?number, page?: number = 0) {
```
grilix: What do we expect to happen if `lastSyncedId` is null? Couldn't we just return right away?

src/utils/accounts.js
```diff
@@ -96,16 +100,14 @@ export const getAccountName = (accountType: AccountTypes, accounts: Accounts): s
...
-export const findAccountById = (
-  accountId: string,
-  accounts: Accounts,
-): ?Account => {
+export const findAccountById = (accountId: string, accounts: Accounts): ?Account => {
   return accounts.find(({ id }) => id === accountId);
 };
+
+export const getAccountWalletId = (account: Account): string => {
+  return get(account, 'walletId', '');
```
grilix: Wouldn't it be the same as `account.walletId || ''`? This function requires an `account` argument, and the only caller already does a null check.

src/utils/assets.js
```diff
@@ -151,11 +152,34 @@ export const addressesEqual = (address1: ?string, address2: ?string): boolean =>
...
   supportedAssetsData: Asset[],
   assetSymbol: string,
 ): Asset | Object => {
-  return assetsData.find(({ symbol }: Asset) => symbol === assetSymbol)
+  return userAssets.find(({ symbol }: Asset) => symbol === assetSymbol)
   || supportedAssetsData.find(({ symbol }: Asset) => symbol === assetSymbol)
   || {};
 };
+
+export const getAssetDataByAddress = (
```
grilix: I'd avoid exporting symbols that are only used on the same file.

src/utils/assets.js
```diff
@@ -151,11 +152,34 @@ export const addressesEqual = (address1: ?string, address2: ?string): boolean =>
...
+  userAssets: Asset[],
+  supportedAssetsData: Asset[],
+  assetAddress: string,
+): Asset | Object => {
+  return userAssets.find(({ address }: Asset) => isCaseInsensitiveMatch(address, assetAddress))
+  || supportedAssetsData.find(({ address }: Asset) => isCaseInsensitiveMatch(address, assetAddress))
+  || {};
```
grilix: As a suggestion, I'd make the return value of this function optional and then make the caller check the result, since making this return `Asset | Object` is almost the same as making it to be just `Object`.

We could write this one as:

```js
const getAssetDataByAddress = (
  userAssets: Asset[],
  supportedAssetsData: Asset[],
  assetAddress: string,
): ?Asset => {
  return userAssets.find(({ address }: Asset) => isCaseInsensitiveMatch(address, assetAddress))
  || supportedAssetsData.find(({ address }: Asset) => isCaseInsensitiveMatch(address, assetAddress));
};
```

And then the caller could be something like:

```js
export const getAssetSymbolByAddress = (assets: Asset[], supportedAssets: Asset[], address: ?string): ?string => {
  if (!address) return;

  // NOTE: ZERO_ADDRESS usually means it's ETH transaction
  if (address === ZERO_ADDRESS) return ETH;

  const assetData = getAssetDataByAddress(assets, supportedAssets, address);
  if (assetData) return assetData.symbol;
};
```

src/services/smartWallet.js
jsidorenko: if `lastSyncedId` is null, then we will return all the elements we would be able to fetch

src/screens/Asset/Asset.js
poocart: Added change to checkIfSmartWalletAccount just to check if passed account object is valid.

src/utils/assets.js
poocart: What do you mean? It's for checking against method argument.

src/utils/assets.js
poocart: Wanted this method of `getAssetDataByAddress` to have the same approach as `getAssetData` as it has same purpose while for `getAssetDataByAddress` we are destructuring right away from method in many places of the code, if you OK changing that then we might need to revisit all places where it's used and apply the removal of destructuring from method.

src/utils/assets.js
jsidorenko: He doesn't like the `export` part

src/utils/assets.js
jsidorenko: Looks like it's better to do the suggested refactoring in a new PR

src/utils/assets.js
grilix: Sorry, I didn't explain this because I mentioned it on another comment.

When linters search for unused symbols, they don't go through all of the codebase, so they don't check if exported symbols are being used somewhere else.

This means that if in the future, we remove the code that uses this function, we'd have to manually search for it on the rest for the project, or just leave it there unused. Unused code will have to be maintained just because it's already there, or since it's not being used we won't know if it's broken.

I think we should help linters catch issues like this and others, before they get into the upstream.

PR #1493, opened 2019-10-31
deeplink parser fixes, wallet connect auth flow check, other chore

poocart:
This PR includes:
1. Small deep link parser improvements.
2. Prevention of showing incoming Wallet Connect request screen if user is locked our from wallet, now it will appear after PIN is entered.
3. Improvement to Wallet Connect requests to show that method is not supported if it's from 3rd RPC method that is not within our Wallet Connect implementation.

### DIFF
--- a/src/actions/deepLinkActions.js
+++ b/src/actions/deepLinkActions.js
@@ -22,6 +22,7 @@ import { Alert } from 'react-native';
 import url from 'url';
 import Toast from 'components/Toast';
 import get from 'lodash.get';
+import isEmpty from 'lodash.isempty';
 import { updateNavigationLastScreenState, navigate } from 'services/navigation';
 import { requestShapeshiftAccessTokenAction } from 'actions/exchangeActions';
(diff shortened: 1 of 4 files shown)


src/actions/deepLinkActions.js
```diff
@@ -76,29 +59,26 @@ const beginApproveLogin = (query: ApproveLoginQuery) => {
 
 export const executeDeepLinkAction = (deepLink: string) => {
   return async (dispatch: Dispatch) => {
-    const params: Object = url.parse(deepLink, true);
-    if (params.protocol !== 'pillarwallet:') return;
-    switch (params.host) {
+    const { host, protocol, query = {} } = url.parse(deepLink, true) || {};
```
grilix: I think it would make sense to explicitly handle the case where `url.parse(deepLink, true)` fails.

Maybe something like:

```js
const params = url.parse(deepLink, true);
if (!params) {
  return; // or whatever, but don't continue going
}

const { host, protocol, query = {} } = params;
```

poocart: Not necessary, but makes sense so added, thanks 👍

PR #1499, opened 2019-11-01
check allAssets array type and report to sentry if wrong received

poocart:
Added a try to identify failure reported to Sentry.

https://sentry.io/organizations/pillar-project-worldwide-ltd/issues/1308336621/?environment=production&project=1294444&referrer=alert_email

### DIFF
--- a/src/actions/historyActions.js
+++ b/src/actions/historyActions.js
@@ -19,7 +19,6 @@
 */
 import get from 'lodash.get';
 import orderBy from 'lodash.orderby';
-import isEmpty from 'lodash.isempty';
 import { NETWORK_PROVIDER } from 'react-native-dotenv';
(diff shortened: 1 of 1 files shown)


src/actions/historyActions.js
```diff
@@ -265,7 +273,13 @@ export const restoreTransactionHistoryAction = (walletAddress: string, walletId:
       api.importedErc20TransactionHistory(walletAddress),
       api.importedEthTransactionHistory(walletAddress),
     ]);
-    if (!allAssets || !allAssets.length) return;
+
+    if (isEmpty(allAssets)) return;
```
jsidorenko:
```suggestion
    if (!allAssets || !allAssets.length) return;
```

PR #1505, opened 2019-11-05
Add Bitcoin support services

grilix:
Adds the base services and the reducer that we're gonna use for BTC support, the UI will
come on a separate PR.


.env
```diff
@@ -25,3 +25,6 @@ SENDWYRE_ACCOUNT_ID=AC_ERP9DMNTAMB
 SENDWYRE_ENVIRONMENT=test
 MOONPAY_KEY=pk_test_6IqwtCJDawLWNcmzaTeBfE4mBF0GwE4
 WIDGET_SIGNATURE=gFg8YE4qZpvya75B
+BITCOIN_NETWORK=testnet
+# BITCOIN_INSIGHT_URL=http://3.8.237.170:3001/insight-api
+BITCOIN_INSIGHT_URL=https://test-insight.bitpay.com/api
```
poocart: How does are different? Why other .envs contains only the one with IP address?

src/actions/bitcoinActions.js
```diff
@@ -0,0 +1,150 @@
...
+
+export const loadBitcoinAddresses = () => {
+  return async (dispatch: Dispatch) => {
+    const { keys = {} } = await loadDb();
+
+    const loaded: string[] = Object.keys(keys);
+
+    dispatch({ type: SET_BITCOIN_ADDRESSES, payload: { addresses: loaded } });
+
+    return loaded;
```
poocart: Why we are returning loaded if we're also setting those in redux state?

src/actions/bitcoinActions.js
```diff
@@ -0,0 +1,150 @@
...
+
+export const sendTransactionAction = (plan: BitcoinTransactionPlan) => {
+  return async () => {
+    const { keys = {} } = await loadDb();
+
+    const rawTransaction = transactionFromPlan(plan, (address: string) => {
+      const wif = keys[address];
+
+      return importKeyPair(wif);
+    });
```
poocart: Could be jus

```
const rawTransaction = transactionFromPlan(plan, (address: string) => importKeyPair(keys[address]);
```

src/actions/bitcoinActions.js
```diff
@@ -0,0 +1,150 @@
...
+        Toast.show({
+          message: 'The transaction was sent to the Bitcoin network',
+          type: 'success',
+          title: 'Transaction sent',
+          autoClose: true,
+        });
+      })
+      .catch(() => {
+        transactionSendingFailed();
+      });
```
poocart: Could be just 
```
.catch(() => transactionSendingFailed());
```
or even
```
.catch(transactionSendingFailed);
```

src/models/Bitcoin.js
```diff
@@ -0,0 +1,48 @@
...
+  isChange?: boolean,
+};
+
+export type BitcoinUtxo = {
+  address: string,
+  txid: string,
+  vout: number,
+  scriptPubKey: string,
+  amount: number,
+  satoshis: number,
```
poocart: Just wondering, in Ether transaction model we also have `amount` with decimal places and even though Ether transactions are actually in wei we keep only `amount` key in the mode. So just thinking, Bitcoin has even less decimals than Ether, so how about keeping `amount` key and parsing as `satoshis` only when needed?

src/reducers/bitcoinReducer.js
```diff
@@ -0,0 +1,144 @@
...
+//     ...state,
+//     data: {
+//       addresses: [
+//         ...addresses,
+//         { address, updatedAt: 0 },
+//       ],
+//       unspentTransactions,
+//     },
+//   };
+// };
```
poocart: This will be needed later? If so maybe adding TODO would explain when and why

src/services/bitcoin.js
```diff
@@ -0,0 +1,161 @@
...
+  }) : [];
+  const planInputs = isValid ? inputs : [];
+
+  return {
+    fee,
+    isValid,
+    inputs: planInputs,
+    outputs: planOutputs,
+  };
+};
```
poocart: How about this

```
if (!isValid) return {
fee,
    isValid,
    inputs: [],
    outputs: [],
};
```

poocart: I wonder if `isValid` needed at all?

src/services/bitcoin.js
```diff
@@ -0,0 +1,161 @@
...
+    .then(({ txid }) => txid)
+    .catch(() => null);
+};
+
+export const transactionFromPlan = (
+  plan: BitcoinTransactionPlan,
+  inputSigner: (address: string) => ECPair,
+  networkName?: string,
+): string => {
+  const network = networkName ? networks[networkName] : NETWORK;
```
poocart:
```suggestion
  const network = networks[networkName] || NETWORK;
```

src/services/bitcoin.js
```diff
@@ -0,0 +1,161 @@
...
+      prevOutScriptType: 'p2pkh',
+      vin: utxoIndex++,
+    });
+  });
+
+  return txb.build().toHex();
+};
+
+export const rootFromMnemonic = async (mnemonic: string, networkName?: string): ECPair => {
+  const network = networkName ? networks[networkName] : NETWORK;
```
poocart: How about exporting this as util as it's being repeated multiple times?

src/services/bitcoin.js
```diff
@@ -0,0 +1,161 @@
...
+  const seed = await bip39.mnemonicToSeed(mnemonic);
+
+  return bip32.fromSeed(seed, network);
+};
+
+export const keyPairAddress = (keyPair: ECPair): string => {
+  const { address } = payments.p2pkh({
+    pubkey: keyPair.publicKey,
+    network: keyPair.network,
+  });
```
poocart: You sure this will always return object that can be deconstructed?

poocart: Just checked and I think not – https://github.com/bitcoinjs/bitcoinjs-lib/blob/19a126c205dee75abbbd9ab185b11eca3996d587/src/payments/p2pkh.js.

I think adding either `lodash.get` or `|| {}` would  make sense

src/services/bitcoin.js
```diff
@@ -0,0 +1,161 @@
...
+
+  return address;
+};
+
+export const importKeyPair = (s: string, networkName?: string): ECPair => {
+  const network = networkName ? networks[networkName] : NETWORK;
```
poocart: DRY

src/services/bitcoin.js
```diff
@@ -0,0 +1,161 @@
...
+
+  return ECPair.fromWIF(s, network);
+};
+
+export const exportKeyPair = (keyPair: ECPair): string => {
+  return keyPair.toWIF();
+};
```
poocart: Could be only
```
export const exportKeyPair = (keyPair: ECPair): string => keyPair.toWIF();
```

src/services/insight.js
```diff
@@ -0,0 +1,50 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+import { BITCOIN_INSIGHT_URL } from 'constants/bitcoinConstants';
+
+const successfulResponse = (response): boolean => {
+  return response.ok && response.status >= 200 && response.status < 300;
```
poocart: Any explanations under `response.status >= 200 && response.status < 300`? Because 204 can be and error while 301 can be content result that has been redirected.

poocart: As well `response.ok` is the same as the `response.status >= 200 && response.status < 300`, I think it's enough to have `response.ok && response.status === 200` here.

Ref - https://developer.mozilla.org/en-US/docs/Web/API/Response/ok.

poocart:
```suggestion
  return response.ok && response.status === 200;
```

src/utils/common.js
```diff
@@ -56,18 +56,38 @@ export const getRandomInt = (min: number, max: number): number => {
...
+ *   decodeAddress('ethereum', 'ethereum:0xaddress') -> 0xaddress
+ *   decodeAddress( 'bitcoin',  'bitcoin:1address' ) -> 1address
+ *
+ * @param prefix         String the prefix part
+ * @param encodedAddress String the '[prefx]:[address]' string
+ *
+ * @return String the address part
+ */
+const decodeAddress = (prefix: string, encodedAddress?: string): string => {
+  if (!encodedAddress || encodedAddress === '') return '';
```
poocart: `lodash.isEmpty` would make this cleaner

```
if (isEmpty(encodedAddress)) return '';
```

src/utils/common.js
```diff
@@ -56,18 +56,38 @@ export const getRandomInt = (min: number, max: number): number => {
...
 
-  if (encodedAddress.length >= 51) {
-    return encodedAddress.substr(9, 42);
+  const len = prefix.length + 1;
+
+  if (encodedAddress.substr(0, len) === `${prefix}:`) {
```
poocart: Maybe it can also be
```
if (encodedAddress.startsWith(`${prefix}:`)) {
```

src/utils/common.js
```diff
@@ -56,18 +56,38 @@ export const getRandomInt = (min: number, max: number): number => {
...
 };
 
+export const decodeBTCAddress = (encodedAddress: string): string => {
+  return decodeAddress('bitcoin', encodedAddress);
+};
+
+export const decodeETHAddress = (encodedAddress: string): string => {
+  return decodeAddress('ethereum', encodedAddress);
+};
+
```
poocart: Maybe adding 'ethereum' and 'bitcoin' prefixes as constants would make sense?

.env
grilix: The IP address is our own node, but it doesn't always work, so we use bitpay's node for now.

src/models/Bitcoin.js
grilix: You are correct, we don't need both, but this is used to represent what insight returns, we don't build these. Since they return both, I thought it would make sense to declare both here.

In other hand, one of the key things we do with these values is select which ones will be used for a transaction. That job is done by `coinselect`, which expect satoshis but in a `value` key, so if we remove the satoshis we'd have to calculate it again on all the cases.

What do you think?

src/reducers/bitcoinReducer.js
grilix: Yeah, well.. This is gonna be needed before we release this on production, this was how I imagined it to be on this approach. The thing is that we wont have multiple addresses on the MVP, so this code wasn't going to actually be used.

We should probably _uncomment_ it so it's not ignored, I don't think it's gonna be a problem.

src/services/insight.js
grilix: You're probably right, will check that. 204 should not represent an error, BTW.

src/services/bitcoin.js
grilix: This is always gonna return an object or throw an error. Updated accordingly.

.env
poocart: Can we add a comment here?

src/models/Bitcoin.js
poocart: If it's return type then I think it's fine 👍

.env
grilix: Okay, added a comment.

src/actions/bitcoinActions.js
```diff
@@ -0,0 +1,169 @@
...
+  UpdateBitcoinBalanceAction,
+} from 'reducers/bitcoinReducer';
+import type { Wallet } from 'models/Wallet';
+import type { BitcoinTransactionPlan, BitcoinUtxo } from 'models/Bitcoin';
+
+import { saveDbAction } from 'actions/dbActions';
+
+const storage = Storage.getInstance('db');
+
+type BitcoinStore = {
```
jsidorenko: I would extract this to models, as we could re-use it in other parts

src/actions/bitcoinActions.js
```diff
@@ -0,0 +1,169 @@
...
+const updateBitcoinBalance = (
+  address: string,
+  unspentTransactions: BitcoinUtxo[],
+): UpdateBitcoinBalanceAction => ({
+  type: UPDATE_BITCOIN_BALANCE,
+  address,
+  unspentTransactions,
+});
+
+export const initializeBitcoinWalletAction = (wallet: Wallet) => {
```
jsidorenko: what type of wallet is it? Generic, ethereum, bitcoin?

jsidorenko: I mean in params

grilix: Good catch, this is actually used from the views so there's not much context.

I might be confusing things here, let's see if this makes sense. We need to make sure the BTC addresses are generated in an idempotent way. My understanding is that we can use the same mnemonic we use for generating the Ethereum address, since it's the one that's never gonna change.

The [checkAuthAction](https://github.com/pillarwallet/pillarwallet/blob/feature/btc/src/screens/Bitcoin/Bitcoin.js#L96) passes this param when validating the pin, which I'm not sure what it is but the `Wallet` type seems to work fine.

We'd use this action in a way similar to what we had on the MVP: https://github.com/pillarwallet/pillarwallet/blob/feature/btc/src/screens/Bitcoin/Bitcoin.js#L96 (You can find some extra context there, althought that's not *exactly* how's gonna be the new implementation)

jsidorenko: Btw, if the wallet was imported from the private key - it won't have a mnemonic

grilix: Damn, you're right 🤔. So we're gonna need to have independent import/export for BTC and we could still store that into `Wallet` with a `btc` prefix, for example (i.e. `btcMnemonic`, `btcPrivateKey`) or, even under a new key (i.e. `{ "btc": { "mnemonic": "...", "privateKey": ... } }`).

What do you think?

jsidorenko: You could just rename `Wallet` into `GenericWallet` or `EthereumWallet` that has a `mnemonic` property marked as optional

grilix: Alright, updated.

PR #1506, opened 2019-11-05
Feature/feed update

vyvil:
(N.B. There are **only** 29 files that needs to be checked in this PR, other changes are bundles, snapshots or binary files :) )

This PR:
1. Adds userEvents reducer and related actions to hold and manage user related events that needed to be shown on wallet's activity feed. 
2. Adds badge received event to the activity feed.
3. Updates ActivityFeed component by:
3.1. Reducing image props being passed to ListItemWithImage component;
3.2. Adding additional parameters, needed to be passed down to specific EventDetails modal;
3.3. Handling two new event types - user event and badge received event.
4. Updates  ListItemWithImage by adding prop form conditional shadow, allowing to pass in diameter and wrapper styles.


src/components/EventDetails/EventHeader.js
```diff
@@ -145,14 +155,14 @@ const EventIcon = styled(Icon)`
...
+  height: ${props => props.imageDiameter || 58}px;
 `;
 
 const ImageTouchable = styled.TouchableOpacity`
-  width: 58px;
-  height: 58px;
-  border-radius: 29px;
+  width: ${props => props.imageDiameter || 58}px;
+  height: ${props => props.imageDiameter || 58}px;
+  border-radius: ${props => props.imageDiameter ? props.imageDiameter / 2 : 29}px;
```
poocart:
```suggestion
  border-radius: ${props => props.imageDiameter / 2 || 29}px;
```

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -314,93 +319,137 @@ class ActivityFeed extends React.Component<Props, State> {
...
+          textStyle={{ color: baseColors.scarlet }}
+          monoColor
+        />);
+        trxData.txType = 'Withdrawal';
+        trxData.hideAmount = true;
+        trxData.hideSender = true;
       }
 
-      // centers line right addons side vertically if status is present
+      // centers line right addons side vertixcally if status is present
```
poocart:
```suggestion
      // centers line right addons side vertically if status is present
```

src/actions/badgesActions.js
```diff
@@ -79,3 +80,32 @@ export const selfAwardBadgeAction = (badgeType: string) => {
     dispatch(offlineApiCall('selfAwardBadge', walletId, badgeType));
   };
 };
+
+export const fetchBadgeAwardHistoryAction = () => {
+  return async (dispatch: Dispatch, getState: GetState, api: SDKWrapper) => {
+    const {
+      user: { data: { walletId } },
+    } = getState();
+    const badgeAwardEvents = await api.fetchNotifications(walletId, BADGE_REWARD_EVENT) || [];
```
jsidorenko: I would force `fetchNotifications()` method to always return an array instead of writing `|| []`

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -314,93 +319,137 @@ class ActivityFeed extends React.Component<Props, State> {
...
+
+      // transaction to / from key wallet / smart wallet
+      if (notification.accountType) {
+        imageProps.itemImageSource = notification.accountType === ACCOUNT_TYPES.KEY_BASED
+          ? keyWalletIcon
+          : smartWalletIcon;
+      }
+
+      if (!imageProps.itemImageSource) {
+        if ((!isContact || showArrowsOnly)) {
```
jsidorenko:
```suggestion
        if (!isContact || showArrowsOnly) {
```

src/components/EventDetails/EventHeader.js
vyvil: imageDiameter could not be passed in as it is not required, thus reducing this line might cause crashes (NaN px as a border-radius value)

poocart: It won't, NaN is false so it would be 29, but in this case there's another piece you want to consider changing this (it has my suggested approach):

https://github.com/pillarwallet/pillarwallet/pull/1506#pullrequestreview-311693976

to both lines match either one or another approach.

src/constants/notificationConstants.js
```diff
@@ -25,3 +25,4 @@ export const COLLECTIBLE = 'COLLECTIBLE';
 export const CONNECTION = 'CONNECTION';
 export const SIGNAL = 'SIGNAL';
 export const BCX = 'BCX';
+export const BADGE = 'badge';
```
jsidorenko: Why lowercased?

src/reducers/badgesReducer.js
```diff
@@ -25,15 +25,17 @@ import {
...
-import type { Badges } from 'models/Badge';
+import type { Badges, BadgeRewardEvent } from 'models/Badge';
 
 export type BadgesReducerState = {
   data: Badges,
   contactsBadges: {
     [contactId: string]: Badges,
   },
   isFetchingBadges: boolean,
+  badgesEvents: BadgeRewardEvent[]
```
jsidorenko:
```suggestion
  badgesEvents: BadgeRewardEvent[],
```

src/components/EventDetails/EventHeader.js
vyvil: Yeah, you are right. I take that back :) and will update this part. Thanks :)

src/constants/notificationConstants.js
vyvil: It is returned like this from the backend. Added a constant to match it

src/reducers/badgesReducer.js
vyvil: 🤦
thought I was getting better at not missing those

src/constants/notificationConstants.js
jsidorenko: 👍

PR #1507, opened 2019-11-05
asset input amount UI adjustments

poocart:
This PR goal is to ignore showing "greater than min amount" input error message if amount is 0 while keeping the input invalid.

As well some minor cleanups as usual.

### DIFF
--- a/src/utils/formHelpers.js
+++ b/src/utils/formHelpers.js
@@ -40,13 +40,16 @@ export function makeAmountForm(
     }
 
     amount = parseNumber(amount.toString());
-    const isValid = enoughForFee && amount <= maxAmount && amount >= minAmount;
+
+    const isValid = enoughForFee
+      && amount <= maxAmount
+      && amount >= minAmount;
 
     if (formSubmitted) return isValid && amount > 0;
     return isValid;
   });
 
-  Amount.getValidationErrorMessage = (amount): string => {
+  Amount.getValidationErrorMessage = (amount): ?string => {
     if (!isValidNumber(amount.toString())) {
       return 'Incorrect number entered.';
     }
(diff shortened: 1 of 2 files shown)


src/utils/formHelpers.js
```diff
@@ -40,13 +40,16 @@ export function makeAmountForm(
     }
 
     amount = parseNumber(amount.toString());
-    const isValid = enoughForFee && amount <= maxAmount && amount >= minAmount;
+
+    const isValid = enoughForFee
```
jsidorenko: 👍

src/utils/formHelpers.js
```diff
@@ -56,6 +59,14 @@ export function makeAmountForm(
...
+      /**
+       * 0 is the first number that can be typed therefore we don't want
+       * to show any error message on the input, however,
+       * the form validation would still not go through,
+       * but it's obvious that you cannot send 0 amount
+       */
+      return null;
     } else if (amount < minAmount) {
       return 'Amount should be greater than 1 Wei (0.000000000000000001 ETH)';
     } else if (decimals === 0 && amount.toString().indexOf('.') > -1) {
```
jsidorenko:
```suggestion
    } else if (decimals === 0 && amount.toString().includes('.')) {
```

src/utils/formHelpers.js
```diff
@@ -56,6 +59,14 @@ export function makeAmountForm(
...
+    } else if (amount === 0) {
+      /**
+       * 0 is the first number that can be typed therefore we don't want
+       * to show any error message on the input, however,
+       * the form validation would still not go through,
+       * but it's obvious that you cannot send 0 amount
+       */
+      return null;
     } else if (amount < minAmount) {
       return 'Amount should be greater than 1 Wei (0.000000000000000001 ETH)';
```
jsidorenko:
```suggestion
      return `Amount should be greater than 1 Wei (0.${'1'.padStart(decimals, '0')} ETH)`;
```

jsidorenko: Or maybe it's even better to put the `minAmount` variable instead

PR #1508, opened 2019-11-06
feature/Test Upgrade Ethers

alazarevski:
Includes all updates from the ethersjs merged with the latest develop changes
Conflicts resolution and updates to recent changes
Tested on iOS ropsten key-based wallet, transactions work


android/rn-libs/react-native-randombytes/android/src/main/java/com/bitgo/randombytes/RandomBytesModule.java
```diff
@@ -0,0 +1,46 @@
...
+import com.facebook.react.bridge.ReactContextBaseJavaModule;
+
+import java.security.SecureRandom;
+import java.util.Map;
+import java.util.HashMap;
+
+import android.util.Base64;
+
+public class RandomBytesModule extends ReactContextBaseJavaModule {
+  private static final String SEED_KEY = "seed";
```
grilix: Is this the key we're gonna use? Maybe we can have a stronger one?

alazarevski: The seed is default from the react-native-randombytes module

PR #1512, opened 2019-11-07
change format of smart wallet settle transaction values

poocart:
Change format of locally saved settlement transaction values to match incoming from SDK.

### DIFF
--- a/src/actions/smartWalletActions.js
+++ b/src/actions/smartWalletActions.js
@@ -126,7 +126,13 @@ import type { TransactionsStore } from 'models/Transaction';
 import { buildHistoryTransaction, updateAccountHistory, updateHistoryRecord } from 'utils/history';
 import { getActiveAccountAddress, getActiveAccountId } from 'utils/accounts';
 import { isConnectedToSmartAccount } from 'utils/smartWallet';
-import { addressesEqual, getBalance, getPPNTokenAddress } from 'utils/assets';
+import {
+  addressesEqual,
+  getAssetData,
+  getAssetsAsList,
+  getBalance,
+  getPPNTokenAddress,
+} from 'utils/assets';
 import {
   formatMoney,
   formatUnits,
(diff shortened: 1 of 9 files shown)


src/actions/smartWalletActions.js
```diff
@@ -1192,14 +1198,19 @@ export const settleTransactionsAction = (txToSettle: TxToSettle[]) => {
...
       const accountId = getActiveAccountId(accounts);
       const accountAddress = getActiveAccountAddress(accounts);
-      const settlementData = txToSettle.map(({ symbol, value, hash }) => ({
-        symbol,
-        value: value.toString(),
-        hash,
-      }));
+      const settlementData = txToSettle.map(({ symbol, value, hash }) => {
+        const { decimals = 18 } = getAssetData(accountAssetsData, supportedAssets, symbol);
+        const parsedValue = parseFloat(value) * (10 ** decimals);
```
jsidorenko: we need to be sure that we store that number as integer (or string), but not the float

src/utils/common.js
```diff
@@ -137,6 +137,7 @@ export const formatFullAmount = (amount: string | number): string => {
 };
 
 export const parseAmountToWei = (amount: string | number, decimals: number): number => {
```
jsidorenko: If the decimals is now 18, then you can't call it Wei ;)

src/utils/common.js
```diff
@@ -137,6 +137,7 @@ export const formatFullAmount = (amount: string | number): string => {
...
+  if (decimals === 0) return Math.floor(+(utils.bigNumberify(amount).toString()));
```
jsidorenko:
```
const convertedNumber = decimals === 0 
? utils.bigNumberify(amount)
: utils.parseUnits(amount.toString(), decimals);
return Math.floor(+convertedNumber.toString());
```

PR #1514, opened 2019-11-08
Feature/fiat exchange supported assets

vyvil:
This PR allows users to select only those assets that are supported by fiat / crypto exchange providers.
(It extends previous work on allowing users to exchange only those assets that are supported by crypto exchange providers)


src/services/api.js
```diff
@@ -827,6 +827,18 @@ SDKWrapper.prototype.fetchMoonPayOffers = function (fromAsset: string, toAsset:
     .catch(() => ({ error: true }));
 };
 
+SDKWrapper.prototype.fetchMoonPaySupportedAssetsTickers = function () {
+  const url = `${MOONPAY_API_URL}/v3/currencies`;
+
+  return Promise.resolve()
+    .then(() => fetch(url))
+    .then(resp => resp.json())
```
grilix: Could this be something like this.. ?

```js
return fetch(url)
  .then(resp => resp.json())
  ...
```

PR #1517, opened 2019-11-09
fixed token contract transfer amount parse to be in right format

poocart:
We recently changed contract transfer method amount parsing util and combined it with transaction estimate and send methods. This caused one issue as Ethers accept hexlified input which used to be result of parsing and after change it longer was.


poocart:
P. S. If hexlify causes confusion this is in other words BigNumer of ethers js lib. This from ethers js lib:
```
defineProperty(BigNumber.prototype, 'toHexString', function() {
    var hex = this._bn.toString(16);
    if (hex.length % 2) { hex = '0' + hex; }
    return '0x' + hex;
});

function bigNumberify(value) {
    if (isBigNumber(value)) { return value; }
    return new BigNumber(value);
}

module.exports = {
    bigNumberify: bigNumberify,
    BigNumber: BigNumber
};
```

PR #1527, opened 2019-11-12
feature/fiat_exchange_min_value

vyvil:
This PR allows showing fiat exchange offer card even if less than min required value is entered but with a note showing min amount required.

(requested by Aldin and support team (per user feedback))


src/screens/Exchange/Exchange.js
```diff
@@ -839,13 +839,15 @@ class ExchangeScreen extends React.Component<Props, State> {
...
+    const disableFiatExchange = isFiat && (minOrMaxNeeded || !!offerRestricted);
+
     return (
       <ShadowedCard
         wrapperStyle={{ marginBottom: 10 }}
         contentWrapperStyle={{ paddingHorizontal: 16, paddingVertical: 6 }}
       >
         <CardWrapper
-          disabled={isTakeButtonDisabled || disableNonFiatExchange}
+          disabled={isTakeButtonDisabled || disableNonFiatExchange || disableFiatExchange}
```
poocart: Hm, what do you think of adding single variable for disabled? This part confuses a bit `... disableNonFiatExchange || disableFiatExchange`.

src/services/api.js
```diff
@@ -794,31 +794,35 @@ SDKWrapper.prototype.getAddressErc20TokensInfo = function (walletAddress: string
 };
 
 SDKWrapper.prototype.fetchMoonPayOffers = function (fromAsset: string, toAsset: string, amount: number) {
+  const minAmount = 20;
```
grilix: Wouldn't it be better to have this number on a constant?

PR #1528, opened 2019-11-13
Feature/migrate to async storage

jsidorenko:
Database migration from pouchdb to AsyncStorage


src/services/storage.js
```diff
@@ -17,135 +17,119 @@
...
+  return this.prefix + id;
 };
 
-Storage.prototype.get = function (id: string) {
-  return this.db().get(id).catch(() => ({}));
+Storage.prototype.get = async function (id: string) {
+  const data = await AsyncStorage.getItem(this.getKey(id))
+    .then(JSON.parse)
+    .catch(() => {});
+  return data || {};
```
grilix: Just curious, wouldn't it make sense to return `null`/`undefined` when the key is not found?

src/services/storage.js
```diff
@@ -17,135 +17,119 @@
...
+  const key = this.getKey(id);
+
+  if (activeDocs[key]) {
+    Sentry.captureMessage('Race condition spotted', {
+      extra: {
+        id,
+        data,
+        forceRewrite,
+      },
+    });
```
grilix: What should happen after this? Just carry on?

In other hand, in development we won't notice anything happened, right? :thinking:

src/services/storage.js
```diff
@@ -17,135 +17,119 @@
...
+  return data || {};
 };
 
-Storage.prototype.getConflicts = function (): Promise<String[]> {
-  return this.getAllDocs()
-    .then(({ rows }) => (
-      rows.map(({ doc }) => doc._conflicts).reduce((memo, item) => memo.concat(item), [])
-    ));
-};
+const activeDocs = {};
```
grilix: I'd have this on the instance, or at the top so it's not hidden here.

src/services/storage.js
```diff
@@ -17,135 +17,119 @@
...
-            _rev: doc._rev,
-          },
-          data,
-        );
-      return this.db().put(record, options);
-    })
-    .then(doc => {
-      activeDocs[id] = false;
-      return doc;
+  const newValue = forceRewrite ? data : merge({}, currentValue, data);
```
grilix: It looks like we don't need the `currentValue` unless we have `forceRewrite`.

Maybe having it on a function could make it simpler:

```js
const mergeValue = async (key: string, data: Object): Object => {
  const currentValue = await this.get(id);
  return merge({}, currentValue, data);
};

const newValue = forceRewrite ? data : mergeValue(key, data);
```

src/services/storage.js
```diff
@@ -17,135 +17,119 @@
...
-    .then(({ rows }) => (
-      rows.map(({ doc }) => doc._conflicts).reduce((memo, item) => memo.concat(item), [])
-    ));
-};
+const activeDocs = {};
+Storage.prototype.save = async function (id: string, data: Object, forceRewrite: boolean = false) {
+  const currentValue = await this.get(id);
+  const key = this.getKey(id);
+
+  if (activeDocs[key]) {
```
grilix: Not really sure, but looking at the rest, it seems like it should include the prefix :thinking:

src/services/storage.js
jsidorenko: The main idea is to keep this storage wrapper backwards-compatible with the previous one

src/services/storage.js
jsidorenko: yeah, it's better to put it on instance. Thanks 👍

src/services/storage.js
jsidorenko: It includes the prefix ;) `const key = this.getKey(id);` returns you the full name with the prefix

jsidorenko: Btw, it actually doesn't matter if it goes with the prefix or without. It's just an old check for race conditions

src/services/storage.js
jsidorenko: It's a primitive check for a race condition. Once we get the error report in sentry - we should review the algorithm.
> In other hand, in development we won't notice anything happened, right? 🤔
As it usually happens, you don't get this kind of errors in development :D plus we have a test for that case

src/services/storage.js
jsidorenko: 👍 good suggestion. This should speed up the execution as well

android/gradle.properties
```diff
@@ -1 +1,2 @@
 org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
+AsyncStorage_db_size_in_MB=60
```
poocart: are you sure this is enough?

ios/Podfile.lock
```diff
@@ -287,6 +289,7 @@ DEPENDENCIES:
   - React/RCTWebSocket (from `../node_modules/react-native`)
   - ReactNativeFabric (from `../node_modules/react-native-fabric`)
   - ReactNativePermissions (from `../node_modules/react-native-permissions`)
+  - "RNCAsyncStorage (from `../node_modules/@react-native-community/async-storage`)"
```
poocart:
```suggestion
  - RNCAsyncStorage (from `../node_modules/@react-native-community/async-storage`)
```

android/gradle.properties
jsidorenko: I would say 6 MB is enough, but 10 x 'enough' = more than enough ;D

ios/Podfile.lock
poocart: P. S. this change might cause different checksum of file, so might need to update it

jsidorenko: I just ran `pod install` locally - lock file looks unchanged

jsidorenko: After I changed to `RNCAsyncStorage (from `../node_modules/@react-native-community/async-storage`)` and re-ran `pod install` it returned me back the quotes. Maybe it's because of `@` in the package name? Or maybe it's because of my new xcode verison..

src/services/storage.js
```diff
@@ -17,135 +17,135 @@
...
+Storage.prototype.migrateFromPouchDB = async function () {
+  const { storageSettings = {} } = await this.get(STORAGE_SETTINGS_KEY);
+  if (storageSettings.pouchDBMigrated) return Promise.resolve();
+
+  try {
+    const pouchDBStorage = PouchDBStorage.getInstance('db');
+    const pouchDocs = await pouchDBStorage.getAllDocs()
+      .then(({ rows }) => rows.map(({ doc }) => doc));
+
+    await Promise.all(pouchDocs.map((doc) => {
```
alazarevski: If this can be done in batches, even more optimized can be AsyncStorage.multiSet(keyvaluePairs,cb)

jsidorenko: It's a one-time operation, I think users could wait for an extra second once in their life ;)

jsidorenko: but the overall suggestion is good ;)

ios/Podfile.lock
poocart: could be due XC, nevermind, this change is not crucial, just look cleanier

PR #1529, opened 2019-11-13
Add redirectUrl to sendwyre widget

grilix:
The redirectUrl makes the screen close when the user finishes the
exchange transaction.


environments/.env.production
```diff
@@ -23,6 +23,7 @@ SENDWYRE_API_URL=https://api.sendwyre.com
 SENDWYRE_WIDGET_URL=https://pay.sendwyre.com/purchase
 SENDWYRE_ACCOUNT_ID=AC_ERP9DMNTAMB
 SENDWYRE_ENVIRONMENT=prod
+SENDWYRE_RETURN_URL=https://offers-webapp-qa.nonprod.pillarproject.io/sendwyre
```
jsidorenko: Do we have a prod link?

src/services/__tests__/sendwyre.test.js
```diff
@@ -0,0 +1,37 @@
...
+    it('returns the txid', async () => {
+      const url = wyreWidgetUrl('0x000', '0.1', '0x111', '1.0');
+      expect(url).toEqual(
+        'https://pay.sendwyre.com/purchase' +
+        '?destCurrency=0.1' +
+        '&dest=ethereum:0x000' +
+        '&sourceAmount=1.0' +
+        '&sourceCurrency=0x111' +
+        '&accountId=AC_ERP9DMNTAMB' +
+        '&redirectUrl=https%3A//offers-webapp-qa.nonprod.pillarproject.io/sendwyre',
```
alazarevski: You can use env here as well SENDWYRE_ACCOUNT_ID, SENDWYRE_RETURN_URL

environments/.env.production
grilix: We don't, actually. I'm asking for it and will update.

src/services/__tests__/sendwyre.test.js
grilix: This is more like a snapshot, we want that literal string to come out.

PR #1534, opened 2019-11-15
Add initial Bitcoin activation flow

grilix:
- Lets the user initialise the Bitcoin wallet
- Lets the user select the Bitcoin wallet
- Provides an address to receive funds


src/constants/featureFlagsConstants.js
```diff
@@ -23,7 +23,9 @@ export const ENABLE_FEATURE_FLAG = 'ENABLE_FEATURE_FLAG';
 export const DISABLE_FEATURE_FLAG = 'DISABLE_FEATURE_FLAG';
 export const INITIAL_FEATURE_FLAGS = {
   SMART_WALLET_ENABLED: false,
+  BITCOIN_ENABLED: false,
 };
 export const DEVELOPMENT_FEATURE_FLAGS = {
   SMART_WALLET_ENABLED: true,
+  BITCOIN_ENABLED: false,
```
poocart: This can be true if we want to check it in DEV

poocart:
I see you added feature flag for Bitcoin, but why it's not implemented in the flows?

src/actions/__tests__/bitcoinActions.test.js
```diff
@@ -51,21 +51,23 @@ describe('Bitcoin actions', () => {
     store = mockStore(initialState);
   });
 
-  describe('refreshAddressBalanceAction', () => {
+  describe('refreshBitcoinBalanceAction', () => {
     describe('for existing address', () => {
       it('updates the balance', async () => {
-        await store.dispatch(refreshAddressBalanceAction(address, false));
+        return store.dispatch(refreshBitcoinBalanceAction(false))
```
jsidorenko: Why you didn't like the linear approach with `await` as it was before?
It looks weird that you're using async/await inside `.then()`

src/actions/appActions.js
```diff
@@ -197,6 +199,8 @@ export const initAppAndRedirectAction = (appState: string, platform: string) =>
 
       await loadAndMigrate('history', dispatch, getState);
 
+      dispatch(loadBitcoinAddresses());
```
jsidorenko:
```suggestion
      dispatch(loadBitcoinAddressesAction());
```
let's keep consistency

src/actions/bitcoinActions.js
```diff
@@ -54,10 +55,10 @@ import { saveDbAction } from 'actions/dbActions';
 const storage = Storage.getInstance('db');
 
 const saveDb = (data: BitcoinStore) => {
-  return saveDbAction('bitcoin', data, true); // TODO: +true+ required?
+  return saveDbAction('bitcoin', data);
```
jsidorenko: `true` flag means you want to override the value instead of merging. Merging could do you a bad job when you merge arrays and when you want to remove some options from the objects

src/actions/bitcoinActions.js
```diff
@@ -54,10 +55,10 @@ import { saveDbAction } from 'actions/dbActions';
...
 };
 
-const loadDb = async (): Promise<BitcoinStore> => {
+const loadDb = async (): BitcoinStore => {
```
jsidorenko:
```suggestion
const loadDb = (): Promise<BitcoinStore> => {
```

src/actions/bitcoinActions.js
```diff
@@ -116,14 +117,22 @@ export const loadBitcoinAddresses = () => {
 
     const loaded: string[] = Object.keys(keys);
 
-    dispatch(setBitcoinAddresses(loaded));
+    if (loaded.length) {
+      dispatch(setBitcoinAddresses(loaded));
```
jsidorenko:
```suggestion
      dispatch(setBitcoinAddressesAction(loaded));
```

src/actions/bitcoinActions.js
```diff
@@ -116,14 +117,22 @@ export const loadBitcoinAddresses = () => {
...
+    }
   };
 };
 
-const fetchBalanceAction = async (address: string): Promise<BitcoinReducerAction> => {
-  const unspentTransactions = await getAddressUtxos(address);
+const outdatedAddresses = (addresses: BitcoinAddress[]): BitcoinAddress[] => {
+  const threshold = Date.now() - REFRESH_THRESHOLD;
```
jsidorenko: Why do you need the `REFRESH_THRESHOLD` param? Maybe it would be worth mentioning this in the comment

src/utils/bitcoin.js
```diff
@@ -0,0 +1,34 @@
...
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+import { MIN_CONFIRMATIONS } from 'constants/bitcoinConstants';
+import type { BitcoinUtxo } from 'models/Bitcoin';
+
+export const satoshisToBtc = (satoshis: number): number => satoshis * 0.00000001;
+export const btcToSatoshis = (btc: number): number => btc * 100000000;
```
jsidorenko:
```suggestion
export const btcToSatoshis = (btc: number): number => Math.floor(btc * 100000000);
```

src/actions/bitcoinActions.js
grilix: If we need a comment for such a simple line, something's not fine. Maybe those are really bad names?

Outdated addresses would be the addresses that haven't been updated for some time (The `REFRESH_THRESHOLD`). This specific constant will determine what's the oldest we still consider "updated", so we return only the ones that are older than that.

What do you think about looking for better wordings here? I'm open to suggestions.

src/actions/__tests__/bitcoinActions.test.js
grilix: I was having some issues but it turns out they weren't caused by this. Reverted.

src/constants/featureFlagsConstants.js
grilix: I wasn't sure of how to handle this feature flag so I just added it as a placeholder back then. Now it's fixed :muscle:

src/actions/bitcoinActions.js
grilix: Alright, added it back.

src/constants/featureFlagsConstants.js
```diff
@@ -23,7 +23,10 @@ export const ENABLE_FEATURE_FLAG = 'ENABLE_FEATURE_FLAG';
 export const DISABLE_FEATURE_FLAG = 'DISABLE_FEATURE_FLAG';
 export const INITIAL_FEATURE_FLAGS = {
   SMART_WALLET_ENABLED: false,
+  // TODO: review 'bitcoin coming soon' message on accounts list
```
poocart: is this sorted?

src/screens/BitcoinNetwork/BitcoinNetworkIntro.js
```diff
@@ -0,0 +1,250 @@
...
+
+const ButtonWrapper = styled(Wrapper)`
+  padding: 0 46px 20px;
+`;
+
+const features = [
+  {
+    key: 'btc',
+    label: 'Bitcoin.',
+    subtext: 'No waiting for miner or block confirmations. Seriously instant transactions.',
```
jsidorenko: This text is incorrect, we don't have ppn for bitcoin

src/services/insight.js
```diff
@@ -28,19 +40,15 @@ export const sendRawTransactionToNode = async (rawtx: string) => {
     },
     body: JSON.stringify({ rawtx }),
   })
-    .then(response => {
-      if (!response.ok) {
-        return new Error('failed');
-      }
-      return response;
-    });
+    .then(receiveResponse('sendRawTransactionToNode'));
```
jsidorenko:
```suggestion
    .then(vaidateResponse('sendRawTransactionToNode'));
```

src/screens/BitcoinNetwork/BitcoinNetworkIntro.js
grilix: You're right, I've used the PPN view as a template, but we don't know what's gonna be here for now, so it's been removed :+1:

src/constants/featureFlagsConstants.js
grilix: Nope. This is for when BTC goes live. I've updated the text so it's a bit clearer, but not sure if it makes sens now. Let me know.

src/screens/Assets/BTCView.js
```diff
@@ -0,0 +1,197 @@
...
+import type { BitcoinAddress, BitcoinUtxo } from 'models/Bitcoin';
+
+// utils
+import { formatMoney } from 'utils/common';
+import { baseColors, fontSizes, fontStyles, spacing } from 'utils/variables';
+import { unspentAmount, satoshisToBtc } from 'utils/bitcoin';
+
+type Props = {
+  baseFiatCurrency: ?string,
+  rates: Object,
```
poocart:
```suggestion
  rates: Rates,
```

src/screens/Assets/BTCView.js
```diff
@@ -0,0 +1,197 @@
...
+    },
+  },
+}: RootReducerState): $Shape<Props> => ({
+  rates,
+  baseFiatCurrency,
+  addresses,
+  unspentTransactions,
+});
+
+const mapDispatchToProps = (dispatch) => ({
```
poocart:
```suggestion
const mapDispatchToProps = (dispatch: Dispatch) => ({
```

src/utils/__tests__/bitcoin.test.js
```diff
@@ -0,0 +1,63 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+import { unspentAmount, btcToSatoshis } from 'utils/bitcoin';
+import type { BitcoinUtxo } from 'models/Bitcoin';
+
+const buildUtxo = (values: $Shape<BitcoinUtxo>): BitcoinUtxo => {
```
poocart: Can be just

```
const buildUtxo = (values: $Shape<BitcoinUtxo>): BitcoinUtxo => ({
...
});
```

PR #1535, opened 2019-11-15
Fix/format units util improvements

poocart:
This PR includes:
1. Fix for Ethplorer SDK parsed values to make them as BigNumber without decimals.
2. Fix for `formatUnits` method to not return `.0` decimal if value has 0 decimals.
3. Covered more values to be parseable using `formatUnits`, those include exponential values with 0 decimals and BigNumber that itself was passed with decimal pointer where it shouldn't but it can happen when converting numeric/string values back and forth.
4. Improved parsed value for Ethplorer service `parseEth` method.
5. Added more testable values for `formatUnits` unit test and Ethplorer service `parseEth` method of Ethplorer on history import unit test.

Related Sentry tickets can be found here – https://sentry.io/organizations/pillar-project-worldwide-ltd/issues/1217401677/events/d9d2d9862902499e820ee5a2f4ccdfa2/?project=1294444&query=is%3Aunresolved.

### DIFF
--- a/src/actions/__tests__/historyActions.test.js
+++ b/src/actions/__tests__/historyActions.test.js
@@ -354,7 +354,7 @@ describe('History Actions', () => {
       });
 
       it('should handle the 0.00104898 eth value that api could return', () => {
-        expect(parseEthValue(0.00104898)).toBe('1048979999999999');
+        expect(parseEthValue(0.00104898)).toBe('1048980000000000');
       });
 
       it('should handle the 1.0 eth value that api could return', () => {
--- a/src/actions/__tests__/historyActions.test.js
+++ b/src/actions/__tests__/historyActions.test.js
@@ -349,8 +349,24 @@ describe('History Actions', () => {
         });
       });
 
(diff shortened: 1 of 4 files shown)


src/actions/__tests__/historyActions.test.js
```diff
@@ -349,9 +349,25 @@ describe('History Actions', () => {
         });
       });
 
-      it('should handle the wrong eth value that api could return', () => {
+      it('should handle the wrong 1e-22 eth value that api could return', () => {
         expect(parseEthValue(1e-22)).toBe('100000000000000');
       });
+
+      it('should handle the 0.00104898 eth value that api could return', () => {
+        expect(parseEthValue(0.00104898)).toBe('1048979999999999');
```
jsidorenko: Shouldn't the `0.00104898` be equal to `1048980000000000`?

poocart: haha, good catch, overthought something, fixed! 👍

PR #1549, opened 2019-11-25
Feature/general settings

vyvil:
This PR moves wallet settings into general settings screen and removes Users screen (as it is not currently needed as we only have one user)

So: 
- no wallet settings on Accounts screen - just regular list items to switch between wallets / networks;
- toast and insight to backup wallet now leads to settings screen.
- no Users screen - clicking to User icon on screen header would lead straight to user settings screen. (For this a default navigation transition is needed - see changes on common.js for navigation transition changes)


src/utils/common.js
```diff
@@ -231,32 +234,59 @@ export const getiOSNavbarHeight = (): number => {
...
+
+const getTransitionDuration = (isFaster: boolean) => {
+  let duration = 400;
+  if (isFaster) {
+    duration = Platform.OS === 'ios' ? 500 : 250;
+  }
+  return duration;
+};
+
+const getTransitionSpec = (isFasterAnimation: boolean) => {
```
jsidorenko: Minor: either there is no need to wrap return with `(...)` or you can just remove `{ return`

PR #1551, opened 2019-11-25
Feature/theme init

vyvil:
This PR inits theming:
1. It adds a way to toggle between light and dark themes (for devs only). (Please note that toggle button is here only temporarily, to make testing faster)
2. It adds themed colours to some components to showcase and test usability on various cases.
3. It renames theme prop of current components to "customTheme" in order not to overwrite global theme prop.


src/actions/appSettingsActions.js
```diff
@@ -17,7 +17,7 @@
     with this program; if not, write to the Free Software Foundation, Inc.,
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
-import { UPDATE_APP_SETTINGS } from 'constants/appSettingsConstants';
+import { DARK_MODE, LIGHT_MODE, UPDATE_APP_SETTINGS } from 'constants/appSettingsConstants';
```
poocart: I just wonder if it's theme maybe it should be `_THEME` instead of `_MODE`?

src/actions/appSettingsActions.js
```diff
@@ -179,3 +180,24 @@ export const setUserJoinedBetaAction = (userJoinedBeta: boolean, ignoreSuccessTo
     });
   };
 };
+
+
+export const changeAppThemeAction = () => {
+  return (dispatch: Dispatch, getState: GetState) => {
+    const {
+      appSettings: { data: { theme: { mode } } },
```
poocart: if changing constants to `_THEME` then maybe this can be
```suggestion
      appSettings: { data: { theme: { current } } },
```

src/actions/appSettingsActions.js
```diff
@@ -179,3 +180,24 @@ export const setUserJoinedBetaAction = (userJoinedBeta: boolean, ignoreSuccessTo
...
+    const colors = mode === LIGHT_MODE ? darkThemeColors : lightThemeColors;
+
+    const theme = { mode: newMode, colors };
+    dispatch(saveDbAction('app_settings', { appSettings: { theme } }));
+    dispatch({
+      type: UPDATE_APP_SETTINGS,
+      payload: {
+        theme,
+      },
+    });
```
poocart: action dispatch in the line above takes almost the same space of code, maybe this can be done in one liner as well?

poocart:
From my personal experience when switched to dark theme on my device I really enjoyed when some of the apps manually switched to dark theme without me to enable it so I wonder maybe adding this `https://github.com/expo/react-native-appearance` would also do this trick for default theme?

vyvil:
> From my personal experience when switched to dark theme on my device I really enjoyed when some of the apps manually switched (and still continue to switch when new app starts to adds dark theme to it) to dark theme without me to enable it so I wonder maybe adding this `https://github.com/expo/react-native-appearance` would also do this trick for default theme? Though this probably needs to be discussed with Dmitry as well.

Yeah, I'm thinking about the same :)
For now I've just added manual toggle for testing purposes

src/actions/appSettingsActions.js
grilix: I don't think we need to squeeze things on one line just because. The important thing is not just the length, but the complexity, and the line above seems a bit simpler, in my opinion. Maybe we could have a function instead that we can reuse on the rest of the actions:

```js
const updateSettingsAction = (payload: Object) => ({
  type: UPDATE_APP_SETTINGS,
  payload,
});

// then...

dispatch(updateSettingsAction({ theme }));
```

Anyways, we don't need to add extra stuff on this PR, we can do that later and leave this as is. What do you think?

vyvil: Hm, maybe it's just me, but I'm personally more used to dispatches that list type and payload in separate lines. In sake of fast readability. (And we usually use the same approach / way across this app).
I've put theme in one line to make it shorter :)

```
dispatch({
   type: UPDATE_APP_SETTINGS,
   payload: { theme },
});
```

Not sure about additional function, though. Isn't basic dispatch more understandable?

grilix: Well we could discuss if we prefer one way or another. I don't think it's really a problem anyway, but IMHO, but there could be a couple of advantages on using a function for that.

The first one is removing some redundant stuff that gets repeated everywhere, like `type:` and `payload:`, this results on less noisy code, which is easier to write and read.

In other hand, it detaches the action structure from the callers. This helps avoid common errors like typos or missing/extra information. This also helps us write the code in a way that's easier for linters to detect those errors before we even have to test the code.
And of course, if we ever want to change that structure, it would be way easier, and with the help of the linter, any outdated call should be detected right away.

Another less probable scenario would be if we want to convert that action into a thunk action (or a saga, probably, when we get there 😄). In that case we would only need to change the function and that's it.

As I said before, I don't like the sole motivation of having fewer lines. I think we should focus on reducing the complexity, instead. This is not a particularly complex portion of code, so I wouldn't worry too much about it right now.

PR #1552, opened 2019-11-26
Bitcoin Transaction list expansion

alazarevski:
Add basic UI and BTC transaction breakdown in BTCView


src/components/EventDetails/EventDetails.js
```diff
@@ -335,7 +341,12 @@ class EventDetails extends React.Component<Props, State> {
     if (eventType === TRANSACTION_EVENT) {
       let txInfo = history.find(tx => tx.hash === eventData.hash);
       if (!txInfo) {
-        txInfo = this.cachedTxInfo || {};
+        if (eventData.asset === 'BTC') {
+          txInfo = eventData;
+          txInfo.value *= 100000000;
```
grilix:
```js
import { btcToSatoshis } from 'utils/bitcoin';
```

```suggestion
          txInfo.value = btcToSatoshis(txInfo.value);
```

src/components/EventDetails/EventDetails.js
```diff
@@ -174,8 +175,12 @@ const Icon = styled(CachedImage)`
   margin-bottom: ${spacing.small}px;
 `;
 
-const viewTransactionOnBlockchain = (hash: string) => {
-  Linking.openURL(TX_DETAILS_URL + hash);
+const viewTransactionOnBlockchain = (hash: string, asset?: ?string) => {
+  let url = TX_DETAILS_URL + hash;
+  if (asset && asset === 'BTC') {
```
grilix: Can we just ..?

```suggestion
  if (asset === 'BTC') {
```

src/components/EventDetails/EventDetails.js
```diff
@@ -118,6 +118,7 @@ type Props = {
   getScrollOffset?: (number) => ScrollToProps,
   getMaxScrollOffset?: (number) => number,
   accounts: Accounts,
+  bitcoinAddresses: Object[],
```
grilix:
```suggestion
  bitcoinAddresses: BitcoinAddress[],
```

src/fixtures/assets.js
```diff
@@ -54,6 +54,23 @@ export const initialAssets = [
     isDefault: true,
     balance: 0,
   },
+  {
+    address: '0x9471ea41772925e3ff477772681740c9f520d3af',
```
grilix: What would this address be? :thinking:

src/reducers/__tests__/bitcoinReducer.test.js
```diff
@@ -152,4 +153,23 @@ describe('Bitcoin reducer', () => {
...
+          type: UPDATE_BITCOIN_TRANSACTIONS,
+          transactions: [],
+          address,
+        };
+
+        const state = reducer(initialState, updateTransactions);
+
+        expect(reducer(state, updateTransactions)).toMatchObject({
+          data: { transactions: [] },
+        });
```
grilix: What are we testing here? It looks like we should add some transactions to make sure this does what's expected to do

environments/.env.production
```diff
@@ -28,3 +28,5 @@ MOONPAY_KEY=pk_live_NlrCbkw5fwVPsdyN0hGT4wsYXXDaFfjv
 WIDGET_SIGNATURE=_widget_signature_
 BITCOIN_NETWORK=mainnet
 BITCOIN_INSIGHT_URL=https://api.bitcore.io/api/BTC/mainnet
+BITCOIN_TX_DETAILS_URL=https://live.blockcypher.com/btc/tx/
+
```
grilix:
```suggestion
```

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -161,6 +161,7 @@ type Props = {
   contactsSmartAddresses: ContactSmartAddressData[],
   emptyState?: EmptyState,
   supportedAssets: Asset[],
+  bitcoinAddresses: Object[],
```
grilix:
```suggestion
  bitcoinAddresses: BitcoinAddress[],
```

src/fixtures/assets.js
```diff
@@ -54,6 +54,23 @@ export const initialAssets = [
     isDefault: true,
     balance: 0,
   },
+  {
+    address: 'mXx',
```
grilix: What would we receive here?

src/services/insight.js
```diff
@@ -62,3 +62,29 @@ export const getAddressBalanceFromNode = (address: string) => {
...
+  return fetch(`${BITCOIN_INSIGHT_URL}/address/${address}/txs?limit=0`, {
+    method: 'GET',
+    headers: {
+      Accept: 'application/json',
+    },
+  }).then(validateResponse('getBTCTransactionsFromNode'))
+    .then(response => response.json())
+    .then(txs => {
+      const fullTxs = txs.map(e => {
+        return fetch(`${BITCOIN_INSIGHT_URL}/tx/${e.mintTxid}/populated`, {
```
grilix: Are we gonna fetch all transactions and then fetch the transaction details for each one? I think this could become a problem quite quickly :thinking:

alazarevski: we will need the info from the populated tx info to form the current way of presenting transactions. The best way would be to form and host the bitcore node where we could bind the information and build our own api.

src/fixtures/assets.js
alazarevski: this is a sample fixture, the address is disregarded in the forming of the asset list.

src/components/ActivityFeed/ActivityFeed.js
alazarevski: Need to import the type

src/reducers/__tests__/bitcoinReducer.test.js
alazarevski: Just verifying that the reducer exists. Please feel free to fill it up with mock data

src/fixtures/assets.js
alazarevski: this prop doesn't get used. If used in tests it helps as mock data

src/components/EventDetails/EventDetails.js
alazarevski: just making sure if this function is used with/without the asset param

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -324,7 +327,8 @@ class ActivityFeed extends React.Component<Props, State> {
     if (type === TRANSACTION_EVENT) {
       const tag = get(notification, 'tag', '');
       const isReceived = addressesEqual(notification.to, activeAccountAddress)
-        || tag === PAYMENT_NETWORK_ACCOUNT_WITHDRAWAL;
+        || tag === PAYMENT_NETWORK_ACCOUNT_WITHDRAWAL
+        || bitcoinAddresses.some(e => e.address === notification.to);
```
jsidorenko: `|| bitcoinAddresses.some(e => addressesEqual(e.address, notification.to));`

src/components/EventDetails/EventDetails.js
```diff
@@ -335,7 +342,12 @@ class EventDetails extends React.Component<Props, State> {
     if (eventType === TRANSACTION_EVENT) {
       let txInfo = history.find(tx => tx.hash === eventData.hash);
       if (!txInfo) {
-        txInfo = this.cachedTxInfo || {};
+        if (eventData.asset === 'BTC') {
+          txInfo = eventData;
```
jsidorenko: maybe it's better to create a copy of the object? `txInfo = { ...eventData };`

src/components/EventDetails/EventDetails.js
```diff
@@ -352,7 +364,9 @@ class EventDetails extends React.Component<Props, State> {
         extra,
       } = txInfo;
 
-      const isReceived = addressesEqual(to, activeAccountAddress) || tag === PAYMENT_NETWORK_ACCOUNT_WITHDRAWAL;
+      const isReceived = addressesEqual(to, activeAccountAddress)
+        || tag === PAYMENT_NETWORK_ACCOUNT_WITHDRAWAL
+        || bitcoinAddresses.some(e => e.address === to);
```
jsidorenko: `|| bitcoinAddresses.some(e => addressesEqual(e.address, to));`

src/components/EventDetails/EventDetails.js
```diff
@@ -562,11 +576,15 @@ class EventDetails extends React.Component<Props, State> {
     if (eventType === TRANSACTION_EVENT || eventType === COLLECTIBLE_TRANSACTION) {
       let txInfo = history.find(tx => tx.hash === eventData.hash);
       if (!txInfo) {
-        txInfo = this.cachedTxInfo || {};
+        if (eventData.asset === 'BTC') {
+          txInfo = eventData;
```
jsidorenko: same as above

src/components/EventDetails/EventDetails.js
```diff
@@ -655,7 +673,11 @@ class EventDetails extends React.Component<Props, State> {
     if (eventType === TRANSACTION_EVENT) {
       let txInfo = history.find(tx => tx.hash === eventData.hash);
       if (!txInfo) {
-        txInfo = this.cachedTxInfo || {};
+        if (eventData.asset === 'BTC') {
+          txInfo = eventData;
```
jsidorenko: same here

src/reducers/bitcoinReducer.js
```diff
@@ -17,20 +17,28 @@
...
   UPDATE_UNSPENT_TRANSACTIONS,
+  UPDATE_BITCOIN_TRANSACTIONS,
 } from 'constants/bitcoinConstants';
 
 export type BitcoinReducerState = {
   data: {
     addresses: BitcoinAddress[],
     unspentTransactions: BitcoinUtxo[],
     balances: BitcoinBalance,
+    transactions: BTCTransaction[]
```
jsidorenko: missed comma in the end

src/reducers/bitcoinReducer.js
```diff
@@ -120,6 +146,43 @@ const updateUnspentTransactions = (
...
+): BitcoinReducerState => {
+  const { transactions: updateTransactions, address } = action;
+  const {
+    data: {
+      addresses, transactions, balances, unspentTransactions,
+    },
+  } = state;
+
+  const filteredTransactions: BTCTransaction[] = transactions.filter(
+    ({ address: txAddress }) => address !== txAddress,
```
jsidorenko: how about `!addressesEqual(address, txAddress)`?
same for other similar places in this file

src/services/bitcoin.js
```diff
@@ -172,3 +174,8 @@ export const getAddressBalance = (address: string): Promise<BTCBalance> => {
   return getAddressBalanceFromNode(address)
     .then(response => response.json());
 };
+
+export const getBTCTransactions = (address: string): Promise<BTCTransaction[]> => {
+  return getBTCTransactionsFromNode(address)
+    .then(response => response);
```
jsidorenko: What's the point in having that line?

src/components/ActivityFeed/ActivityFeed.js
grilix: Bitcoin addresses have upper and lower case letters, we should not use case-insensitive matching.

jsidorenko: Eth addresses have upper and lower case letters as well, sometimes different api could return you the same hash but with differently capitalized letters and that's why we're using addressesEqual() helper

grilix: Yes, on ethereum addresses are hashes, so they are the same no matter the capitalization. On bitcoin they're not hashes, and capitalization matters. No API should change the capitalization of bitcoin addresses, since they'd be different (or even invalid) addresses.

jsidorenko: got it! good to know ;)

grilix: Sorry, maybe "hashes" vs "not hashes" is misleading, I mean they're not hexadecimal numbers or anything like that.

jsidorenko: :D 👍

PR #1573, opened 2019-12-04
Separate send token amount into a component

grilix:
In preparation to be able to handle different types of amount selection, this PR separates the Ethereum specific code from SendTokenAmount into a component.

### DIFF
--- a/src/components/SendTokens/ETHTokens.js
+++ b/src/components/SendTokens/ETHTokens.js
@@ -52,8 +52,9 @@ import type { NavigationScreenProp } from 'react-navigation';
 import type { Account } from 'models/Account';
 import type { GasInfo } from 'models/GasInfo';
 import type { TokenTransactionPayload } from 'models/Transaction';
-import type { Balances, Rates } from 'models/Asset';
+import type { Balances, Rates, AssetData } from 'models/Asset';
 import type { RootReducerState } from 'reducers/rootReducer';
+import type { SessionData } from 'models/Session';
 
 // constants
 import { SEND_TOKEN_CONFIRM } from 'constants/navigationConstants';
@@ -100,24 +101,18 @@ const BackgroundWrapper = styled.View`
 `;
 
 type Props = {
-  assetData: Object,
+  assetData: AssetData,
   receiver: string,
   source: string,
-  token: string;
-  address: string,
-  totalBalance: number,
-  contractAddress: string,
   navigation: NavigationScreenProp<*>,
-  isVisible: boolean,
-  formValues?: Object,
   balances: Balances,
-  session: Object,
-  fetchGasInfo: Function,
+  session: SessionData,
+  fetchGasInfo: () => void,
   gasInfo: GasInfo,
   rates: Rates,
   baseFiatCurrency: ?string,
   transactionSpeed: ?string,
-  updateAppSettings: Function,
+  updateAppSettings: (path: string, value: any) => void,
   activeAccountAddress: string,
   activeAccount: ?Account,
 };
@@ -488,7 +483,7 @@ const mapStateToProps = ({
 
 const mapDispatchToProps = (dispatch) => ({
   fetchGasInfo: () => dispatch(fetchGasInfoAction()),
-  updateAppSettings: (path, value) => dispatch(updateAppSettingsAction(path, value)),
+  updateAppSettings: (path: string, value: any) => dispatch(updateAppSettingsAction(path, value)),
 });
 
 export default connect(mapStateToProps, mapDispatchToProps)(SendETHTokens);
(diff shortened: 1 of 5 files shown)


src/components/SendTokens/ETHTokens.js
```diff
@@ -0,0 +1,495 @@
...
+const { Form } = t.form;
+const MIN_TX_AMOUNT = 0.000000000000000001;
+
+const SPEED_TYPE_LABELS = {
+  [SPEED_TYPES.SLOW]: 'Slow',
+  [SPEED_TYPES.NORMAL]: 'Normal',
+  [SPEED_TYPES.FAST]: 'Fast',
+};
+
+class SendTokenAmount extends React.Component<Props, State> {
```
jsidorenko: I would change the class name as well. Could be `SendETHTokens` as in your reference, or `ETHTokens` as in file name, or something similar

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -139,15 +63,6 @@ type State = {
...
-const MIN_TX_AMOUNT = 0.000000000000000001;
-
-const SPEED_TYPE_LABELS = {
-  [SPEED_TYPES.SLOW]: 'Slow',
-  [SPEED_TYPES.NORMAL]: 'Normal',
-  [SPEED_TYPES.FAST]: 'Fast',
-};
-
 class SendTokenAmount extends React.Component<Props, State> {
   _form: t.form;
```
jsidorenko: You don't need this variable here

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -19,109 +19,33 @@
 */
 import * as React from 'react';
 import { connect } from 'react-redux';
-import { TouchableOpacity, Keyboard } from 'react-native';
 import t from 'tcomb-form-native';
```
jsidorenko: I think you can remove this as well

src/screens/SendToken/SendTokenAmount.js
```diff
@@ -139,15 +63,6 @@ type State = {
...
-const { Form } = t.form;
-const MIN_TX_AMOUNT = 0.000000000000000001;
-
-const SPEED_TYPE_LABELS = {
-  [SPEED_TYPES.SLOW]: 'Slow',
-  [SPEED_TYPES.NORMAL]: 'Normal',
-  [SPEED_TYPES.FAST]: 'Fast',
-};
-
 class SendTokenAmount extends React.Component<Props, State> {
```
jsidorenko: Pls clean the state object's type

grilix: Oops, forgot about the state. Done.

src/components/SendTokens/ETHTokens.js
grilix: Fixed.

src/screens/SendToken/SendTokenAmount.js
grilix: Removed.

src/components/SendTokens/ETHTokens.js
```diff
@@ -0,0 +1,495 @@
...
+  background-color: ${UIColors.defaultBackgroundColor};
+`;
+
+const BackgroundWrapper = styled.View`
+  background-color: ${UIColors.defaultBackgroundColor};
+  flexGrow: 1;
+`;
+
+type Props = {
+  assetData: Object,
```
vyvil: Typing? 😏😛

src/components/SendTokens/ETHTokens.js
```diff
@@ -0,0 +1,495 @@
...
+      value: { amount },
+      gettingFee: false,
+      calculatingMaxValue: false,
+      txFeeInWei,
+    });
+    this.checkFormInputErrors();
+  };
+
+  getGasLimit = (amount: number) => {
+    // calculate either with amount in form or provided as param
```
vyvil: I'm sensing, that this comment is obsolete, as we are not taking amount from form data, just passing it as param

src/components/SendTokens/ETHTokens.js
jsidorenko: Trolling? :D

vyvil: :)) In a friendly manner 😛

grilix: Haha, I'm glad we're all aware of typing :muscle:. I've previously copied this code and just left what was there since I wasn't sure what type that would be. Now I've added a new type (the `Asset` type doesn't have `token` or `icon`), and updated the rest. Please let me know what you think.

jsidorenko: It would be good to refactor that `assetData` at some point (in a separate PR). Before we had a one-screen send flow, so it was very easy to construct the final `assetData` object and pass it to the send action.
Now we have multi-steps and multichains form, so I think it's better to pass only really needed info between screens like the token, amount and selected speed/fee, no other fields needed.

PR #1574, opened 2019-12-05
Feature/walkthrough concept

vyvil:
This PR adds a way to create walkthroughs to onboard users.
It allows to add walkthrough scenario - list walkthrough steps and add wrap related components with a HOC that calculates item position and manages whether tooltip should be shown or not.

Based on [designs](https://app.zeplin.io/project/5db9b5d14c30ae56faaa3030/dashboard?seid=5de0ea737b323b780fa0981a), walkthroughs (at least, now) can have two ways of showing user contextual info:
shade - when info is showed on a backdrop (to provide general information about screen, for example);
tooltip - when providing info about specific item on the screen.
Both of those ways allow to show title, body text and button.
Button usually changes the step or ends walkthrough (if item is the last in line).

**LIMITATIONS (BACKGROUND):**
RN Android does not allow to have clickable overflowing components, hence tooltips on Android are now acting as dialog window / modal. Meaning, that tooltip is being drawn on top of all items based on related item's position and does not stick to item itself if user tries to interact with the screen. 
That being said, walkthrough on Android whenever step is active has a backdrop that blocks other screen components leaving just a shade's or tooltip's components clickable.

iOS allows clickable overflowing components (with some limitations listed below). Yet in most cases tooltips get covered by elements positioned lower in the code (due to zIndex). To avoid it, I've added another HOC that should be used to wrap the parent of related walkthrough item - to set it's zIndex to higher one.

iOS does not allow to have clickable overflowing components in FlatView / ScrollView.
([demo snack](https://snack.expo.io/ByTjfEI6H))

**FURTHER DEVELOPMENT**
(as this is only a concept, I've left some improvements for later. I've list them here)
1. User should be able to close tooltip by solely pressing on it, rather than the button in the tooltip (as it might not be present).
2. There might be a case when walkthrough step should be triggered once state has changed and user gets back to related screen, so pending walkthroughs might be needed.


src/components/Walkthrough/WalkthroughItem.js
```diff
@@ -0,0 +1,163 @@
...
+}
+
+const mapStateToProps = ({
+  walkthroughs: {
+    type: walkthroughType,
+    waitingForStepId,
+    activeStepId,
+    steps,
+  },
+}) => ({
```
grilix:
```suggestion
}: RootReducerState): $Shape<Props> => ({
```

src/components/Walkthrough/Walkthrough.js
```diff
@@ -0,0 +1,292 @@
...
+import Toast from 'components/Toast';
+import { spacing, fontStyles } from 'utils/variables';
+import { endWalkthroughAction, setWaitingForStepIdAction, setActiveStepIdAction } from 'actions/walkthroughsActions';
+import type { Measurements, Steps } from 'reducers/walkthroughsReducer';
+import { WALKTHROUGH_TYPES } from 'constants/walkthroughConstants';
+import { themedColors } from 'utils/themes';
+import { hexToRgba } from 'utils/ui';
+import { WalkthroughTooltip } from './WalkthroughTooltip';
+
+const { width, height: h } = Dimensions.get('window');
```
poocart:
```suggestion
const { width, height: windowHeight } = Dimensions.get('window');
```
?

src/reducers/walkthroughsReducer.js
```diff
@@ -0,0 +1,120 @@
...
+          return {
+            ...step,
+            measure: action.payload.measure,
+          };
+        }),
+      };
+    case END_WALKTHROUGH:
+      return {
+        ...state,
+        id: '',
```
poocart: Not needed as it's from Step object?

poocart: How about returning merged `initialState` param here?

src/components/Walkthrough/WalkthroughTooltip.js
```diff
@@ -0,0 +1,161 @@
...
+  const {
+    x: stepXPos,
+    y: stepYPos,
+    h: stepItemHeight,
+    posOverwrites = {},
+  } = targetMeasurements;
+
+  const { x: xOverwrite, y: yOverwrite } = posOverwrites;
+
+  const adjustedY = Platform.OS === 'ios' ? stepYPos : stepYPos + ExtraDimensions.getStatusBarHeight();
```
poocart: `Platform.select` for me would look cleaner in this case, but not necessary.

src/components/Walkthrough/WalkthroughItem.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+  };
+
+  showNext = () => {
+    const { showNextStep } = this.props;
+    this.toggleTooltip();
+    showNextStep();
+  };
+
+  setWalkthroughStepMeasures = async () => {
```
poocart:
```suggestion
  setWalkthroughStepMeasures = () => {
```

src/components/Walkthrough/WalkthroughItem.js
```diff
@@ -0,0 +1,163 @@
...
+  posOverwrites?: PosOverwrites,
+}
+
+type State = {
+  showTooltip: boolean,
+  refMeasures: ?Measurements,
+  tooltipInfo: ?TooltipInfo,
+};
+
+const measure = async (ref: View): Promise<Measurements> =>
```
poocart:
```suggestion
const measure = (ref: View): Promise<Measurements> =>
```

src/components/Walkthrough/Walkthrough.js
vyvil: Since I've only using h on the next line to set window height by platform (as I needed it adjusted for Android), I simply named it h. But yeah, I could rename it :)

src/components/Walkthrough/WalkthroughTooltip.js
vyvil: Hm, we usually use Platform.select for styled components. 
Just following the trend (not sure if I'm the one that started it, though :D)

src/actions/walkthroughsActions.js
```diff
@@ -0,0 +1,70 @@
...
+  payload: id,
+});
+
+export const setActiveStepIdAction = (id: string) => ({
+  type: SET_ACTIVE_STEP_ID,
+  payload: id,
+});
+
+export const showNextStepExternalAction = () => {
+  return async (dispatch: Dispatch, getState: GetState) => {
```
poocart:
```suggestion
  return (dispatch: Dispatch, getState: GetState) => {
```

src/components/Walkthrough/Walkthrough.js
```diff
@@ -0,0 +1,291 @@
...
+      this.endWalkthrough();
+    }
+  }
+
+  initWalkthrough = () => {
+    this.setState({ isActiveWalkthrough: true });
+    this.nextStep();
+  };
+
+  nextStep = async (forcedStepIndex?: number) => {
```
poocart:
```suggestion
  nextStep = (forcedStepIndex?: number) => {
```

src/screens/Exchange/Exchange.js
```diff
@@ -43,6 +43,8 @@ import SelectorInput from 'components/SelectorInput';
 import Button from 'components/Button';
 import Spinner from 'components/Spinner';
 import DeploymentView from 'components/DeploymentView';
+import WalkthroughItem from 'components/Walkthrough/WalkthroughItem';
+import { WalkthroughItemParent } from 'components/Walkthrough/WalkthroughItemParent';
```
poocart: How about export it and import it as single default?

vyvil: Yeah, could be done :) Will adjust

App.js
```diff
@@ -43,10 +43,12 @@ import { Container } from 'components/Layout';
 import Root from 'components/Root';
 import Toast from 'components/Toast';
 import Spinner from 'components/Spinner';
+import Walkthrough from 'components/Walkthrough';
 import type { RootReducerState } from 'reducers/rootReducer';
 import { getThemeByType, defaultTheme } from 'utils/themes';
 
 import configureStore from './src/configureStore';
+import type { Steps } from './src/reducers/walkthroughsReducer';
```
poocart:
```suggestion
import type { Steps } from 'reducers/walkthroughsReducer';
```

PR #1582, opened 2019-12-09
Feature/updated onboarding

vyvil:
This PR updates onboarding UX.
And adds theme to most of onboarding flow components and all it's screens.
Changes follows those designs (onboarding part of it):
https://app.zeplin.io/project/5db9b5d14c30ae56faaa3030?seid=5de0ea737b323b780fa0981a.

(I've also introduced new spacing variable - layoutSides. As default layout sides tend to keep on changing :) So having variable that is solely responsible for layout horizontal spacing would make it easier for us to update layouts)

Still remaining and left for later changes:
1. Ens addon for text input.
2. Referral award modal.

NOTE: we might need to remove intro text ("Pillar is next generation smart wallet, payment network and identity manager.") before merging in, as smart wallet is not yet available for all users.


src/screens/PinCodeConfirmation/PinCodeConfirmation.js
```diff
@@ -56,7 +65,8 @@ class PinCodeConfirmation extends React.Component<Props, State> {
   }
 
   handlePinSubmit = (pin: string) => {
-    const { onboarding: wallet } = this.props.wallet;
+    const { wallet: walletData, confirmPinForNewWallet, navigation } = this.props;
```
jsidorenko:
```suggestion
    const { wallet: { onboarding: wallet }, confirmPinForNewWallet, navigation } = this.props;
```

src/screens/BiometricsPrompt/BiometricsPrompt.js
```diff
@@ -0,0 +1,95 @@
...
+  proceedToRegisterWallet = (setBiometrics) => {
+    const { registerWallet } = this.props;
+    registerWallet(setBiometrics);
+  };
+
+  render() {
+    return (
+      <ContainerWithHeader headerProps={{ centerItems: [{ title: 'Make crypto easy' }] }}>
+        <ContentWrapper contentContainerStyle={{ paddingVertical: 20, paddingHorizontal: 30, flexGrow: 1 }}>
+          <HeaderText>{'Would you like to use\nTouch ID with your\nwallet?'}</HeaderText>
```
grilix: Not sure of how important this could be, but "Touch ID" is an iOS thing, I assume. Wouldn't we need to call it something else on android? :thinking:

vyvil: Good call :)
I could either change it to "Fingerprint unlock" for Android, or unify it for both platforms which would be in line with what we now have in settings - "Biometric Login".
I'll check with Dmitry what he would prefer.
Thanks :)

vyvil:
Changing to WIP - will change unlock type by platform on this new prompt screen and on settings as well.

vyvil:
Done with the changes - updated wording and images on new biometry prompt and settings screens based on supported biometry type.

And also added an info toast for an edge case when enabled biometrics can't be used due to a lack of connection keys (pin code is needed to generate them).

src/components/Layout/ContainerWithHeader.js
```diff
@@ -94,9 +95,16 @@ class ContainerWithHeader extends React.Component<Props> {
...
     const { children, keyboardAvoidFooter } = this.props;
-    if (!isShortScreenWithFooter) return children;
+    if (!isShortScreenWithFooter && !shouldRenderChildrenInScrollView) return children;
+    if (!isShortScreenWithFooter && shouldRenderChildrenInScrollView) {
+      return (
+        <ScrollView style={{ flex: 1 }}>
+          {children}
+        </ScrollView>
+      );
+    }
```
grilix: What do you think about having the first condition of both `if`s into its own `if`? I seems like it's claner to me, but not sure.

Something like:
```js
    if (!isShortScreenWithFooter) {
      if (!shouldRenderChildrenInScrollView) {
        return children;
      }

      return (
        <ScrollView style={{ flex: 1 }}>
          {children}
        </ScrollView>
      );
    }
```

vyvil: Sure :) Will adjust

PR #1584, opened 2019-12-10
Fix/bitcoin keys generation

alazarevski:
Remove BTC keypair and store address only, derive signing key per need


src/services/bitcoin.js
```diff
@@ -113,7 +113,7 @@ export const sendRawTransaction = async (rawTx: string): Promise<string> => {
 
 export const transactionFromPlan = (
   plan: BitcoinTransactionPlan,
-  inputSigner: (address: string) => ECPair,
+  keyPair: ECPair,
```
grilix: Why are we doing this?

grilix: Transactions can be made from multiple addresses, which will require different keys.

alazarevski: but the wallet currently has the keyPair derived from the only mnemonic of the pillarwallet, those inputs are coming from the unspent transactions to that address for sending.

grilix: Yes, but this service doesn't have to be aware of that limitation, IMO. We can delegate that to the caller.

```js
const { mnemonic, privateKey, path } = wallet;
let seed = privateKey;
if (mnemonic && mnemonic !== 'ENCRYPTED') {
  seed = mnemonic;
}

const root = await rootFromMnemonic(seed);

const rawTransaction = transactionFromPlan(plan, () => {
  return root.derivePath(path);
});
```

This would leave us some ground ready for multiple addresses, for example, we could have an object mapping addresses to paths, and it would be just changing it to:

```js
// const addressPathMap = { "<address>": <path> };
const rawTransaction = transactionFromPlan(plan, (address) => {
  return root.derivePath(addressPathMap[path]);
});
```

Having multiple addresses is gonna be mandatory, so I think we should have as much code ready for that as we can.  Do you think this make sense?

alazarevski: Makes sense, added a ToDo and reverted the inputSigner callback 18b5299

PR #1585, opened 2019-12-10
Add Bitcoin amount selection screen

grilix:
Move the individual screen for selecting the Bitcoin amount into the generic screen which will select the used component depending on the asset being sent.


src/screens/Assets/BTCView.js
```diff
@@ -108,10 +108,28 @@ class BTCView extends React.Component<Props, State> {
     this.refreshBalance();
   }
 
+  onPressSend = () => {
+    const { supportedAssets } = this.props;
+    const btcToken = supportedAssets.find(e => e.symbol === 'BTC');
```
jsidorenko:
```suggestion
    const btcToken = supportedAssets.find(e => e.symbol === BTC);
```

src/screens/Assets/BTCView.js
```diff
@@ -108,10 +110,28 @@ class BTCView extends React.Component<Props, State> {
...
+      decimals,
+    } = btcToken;
+
+    const assetData: AssetData = {
+      token,
+      icon,
+      decimals,
+    };
+
+    this.props.navigation.navigate(SEND_TOKEN_FROM_ASSET_FLOW, { assetData });
```
alazarevski: [SEND_BITCOIN_CONTACTS]: SendBitcoinContactsScreen,	because the one specified has ETH validation

grilix:
Good catch, @alazarevski. Updated.

PR #1587, opened 2019-12-11
do not perform smart wallet reset if sdk is not initialized

poocart:
Small fix that checks if smart wallet is initialized before calling SDK reset method.

### DIFF
--- a/src/services/smartWallet.js
+++ b/src/services/smartWallet.js
@@ -398,14 +398,12 @@ class SmartWallet {
   }
 
   async reset() {
+    if (!this.sdkInitialized) return;
     this.sdkInitialized = false;
     if (!this.sdk) return;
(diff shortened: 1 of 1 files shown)


src/services/smartWallet.js
```diff
@@ -398,14 +398,12 @@ class SmartWallet {
...
+    if (!this.sdkInitialized) return;
     this.sdkInitialized = false;
     if (!this.sdk) return;
     this.sdk.event$.next(null); // unsubscribes
     subscribedToEvents = false;
-    await this.sdk.reset({
-      device: true,
-      session: true,
-    }).catch(null);
+    await this.sdk.reset({ device: true, session: true }).catch(null);
```
jsidorenko: It was more readable before, IMO

PR #1589, opened 2019-12-12
change address to real sender name on synthetics settle screen

poocart:
Fixes issue where real sender address is not shown on "Settle Transactions" screen.

### DIFF
--- a/src/actions/smartWalletActions.js
+++ b/src/actions/smartWalletActions.js
@@ -152,8 +152,8 @@ import { getWalletsCreationEventsAction } from './userEventsActions';
 const storage = Storage.getInstance('db');
 
 const isValidSyntheticExchangePayment = (type: string, extra: any) => {
-  const smartWalletAccountPaymentTypes = sdkConstants.AccountPaymentTypes || {};
-  return type === smartWalletAccountPaymentTypes.SyntheticsExchange && !isEmpty(extra);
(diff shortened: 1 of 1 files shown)


src/actions/smartWalletActions.js
```diff
@@ -151,6 +151,11 @@ import { getWalletsCreationEventsAction } from './userEventsActions';
 
 const storage = Storage.getInstance('db');
 
+const isValidSyntheticExchangePayment = (type: string, extra: any) => {
+  const smartWalletAccountPaymentTypes = sdkConstants.AccountPaymentTypes || {};
```
jsidorenko:
```suggestion
  const syntheticsExchangeType = get(sdkConstants, 'AccountPaymentTypes.SyntheticsExchange');
  return type === syntheticsExchangeType && !isEmpty(extra);
```

src/actions/smartWalletActions.js
```diff
@@ -151,6 +151,11 @@ import { getWalletsCreationEventsAction } from './userEventsActions';
 
 const storage = Storage.getInstance('db');
 
+const isValidSyntheticExchangePayment = (type: string, extra: any) => {
+  const syntheticsExchangeType = get(sdkConstants, 'AccountPaymentTypes.SyntheticsExchange');
+  return !isEmpty(type) && !isEmpty(extra) && type === syntheticsExchangeType;
```
jsidorenko: 👍

PR #1592, opened 2019-12-16
Feature/text inputs

vyvil:
This PR:
1. Combines text input components into one reusable component (the only one needed according to [design system](https://app.zeplin.io/styleguide/5dc42334d688528346c12861/components?seid=5db9c95ed75e92126d5e455c)).
2. Adds themedColors to related screens (where text inputs where updated).
3. Add some other adjustments.

(Needed for Referrals)


src/screens/ImportWallet/ImportWallet.js
```diff
@@ -144,14 +144,14 @@ const ButtonInner = styled.View`
...
-  color: ${props => props.blue ? baseColors.electricBlue : baseColors.white};
+  color: ${({ blue, theme }) => blue ? theme.colors.primary : theme.colors.control};
   font-size: ${fontSizes.big}px;
   margin-bottom: 2px;
 `;
 
 const ButtonIconArrow = styled(Icon)`
   font-size: ${fontSizes.big}px;
-  color: ${props => props.blue ? baseColors.electricBlue : baseColors.white};
+  color: ${({ blue, theme }) => blue ? theme.colors.primary : theme.colors.control};
```
poocart: Just thinking here if `blue` prop is set we go for `primary` color which might change as well, maybe we should rename prop as `defaultColor` or `defaultTheme` or `primaryColor`?

src/screens/Exchange/Exchange.js
```diff
@@ -349,41 +349,53 @@ function SelectorInputTemplate(locals) {
       inputRef,
       onSelectorOpen,
     },
+    value,
   } = locals;
+  const { selector = {} } = value;
```
poocart: Just thinking maybe use lodash.get here for `value` and use `{}` fallback as form input tend to produce undefined / null values, what do you think?

PR #1593, opened 2019-12-16
added missing setUsername and tokens update callback during onboarding

poocart:
This PR includes fixes for 2 OAuth related issues which might happen only as edge case, but as we're having problems with OAuth tokens it's good to solve them all and those might even cover other cases which we're not even familiar yet. Fixes:
1. We were not setting tokens update callback method during onboarding.
2. If tokens expire after onboarding without re-login they were always failing to update as we were not setting username in SDK which is required on SDK's `refreshToken` method (ref – https://github.com/pillarwallet/pillarwallet-nodejs-sdk/blob/master/src/lib/register.ts#L202).


src/actions/onboardingActions.js
```diff
@@ -333,16 +337,22 @@ export const registerWalletAction = (enableBiometrics?: boolean) => {
...
       username: user.username,
       walletId: sdkWallet.walletId,
       ethAddress: wallet.address,
       fcmToken,
       ...oAuthTokens,
-    }));
+    };
 
-    if (!registrationSucceed) { return; }
+    await dispatch(signalInitAction(signalCredentials));
```
poocart: reorganized method priorities

PR #1600, opened 2019-12-17
remove firebase based feature flags and rely on userInfo feature flags

poocart:
This includes change of Firebase feature flags service to platform-core. From now on we're going to call SDK's `userInfo` endpoint which will provide feature flags enabled for user. As well this allows us to cut extra calls of fetching Firebase feature flags and setting Firebase Analytics data which results as slightly faster login and onboarding speeds.

One more note (especially for @alazarevski and @grilix): changed copies of Early Access programs to include Bitcoin feature.


ios/Podfile
```diff
@@ -38,18 +38,17 @@ target 'pillarwallet' do
...
   pod 'Firebase/Messaging'
   pod 'Firebase/DynamicLinks'
-  pod 'Firebase/RemoteConfig'
   pod 'Firebase/Storage'
   pod 'Firebase/Performance'
   pod 'ReactNativePermissions', :path => '../node_modules/react-native-permissions'
   pod 'react-native-fetch-blob', :path => '../node_modules/react-native-fetch-blob'
   pod 'RNKeychain', :path => '../node_modules/react-native-keychain'
 
+  pod 'Intercom', '~> 5.2.1'
```
poocart: This locks Intercom Pod version to the one that matches it's react native library javascript dependancy.

src/actions/featureFlagsActions.js
```diff
@@ -18,18 +18,32 @@
...
-    const {
-      session: { data: { isOnline } },
-    } = getState();
-    if (!isOnline) return;
-    const featureFlags = await getRemoteFeatureFlags();
+export const loadFeatureFlagsAction = (userInfo?: any) => {
+  return async (dispatch: Dispatch, getState: GetState, api: SDKWrapper) => {
+    if (!userInfo) {
+      const walletId = get(getState(), 'user.data.walletId');
+      userInfo = await api.userInfo(walletId);
```
jsidorenko: How about fall backing to the cached flags when the user is online?

poocart: Currently we use `loadFeatureFlagsAction` in 3 cases, 2 of them (login and onboarding) passes `userInfo` that is retrieved few calls before fetching feature flags and 3rd case that doesn't pass `userInfo` actually needs latest flags to be fetched. I think caching won't provide any improvement.

jsidorenko: makes sense 👍

PR #1610, opened 2019-12-19
fix/btc mainnet env

alazarevski:
Change BITCOIN_NETWORK to 'bitcoin' and switch mainnet on api calls

### DIFF
--- a/environments/.env.production
+++ b/environments/.env.production
@@ -28,6 +28,6 @@ SENDWYRE_ENVIRONMENT=prod
 SENDWYRE_RETURN_URL=https://offers-webapp-prod.prod.pillarproject.io/sendwyre
 MOONPAY_KEY=pk_live_NlrCbkw5fwVPsdyN0hGT4wsYXXDaFfjv
 WIDGET_SIGNATURE=_widget_signature_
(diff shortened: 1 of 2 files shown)


environments/.env.production
```diff
@@ -28,6 +28,6 @@ SENDWYRE_ENVIRONMENT=prod
 SENDWYRE_RETURN_URL=https://offers-webapp-prod.prod.pillarproject.io/sendwyre
 MOONPAY_KEY=pk_live_NlrCbkw5fwVPsdyN0hGT4wsYXXDaFfjv
 WIDGET_SIGNATURE=_widget_signature_
-BITCOIN_NETWORK=mainnet
+BITCOIN_NETWORK=bitcoin
```
jsidorenko: What's the point to set it to `bitcoin` as you're changing it to the `mainnet` here: https://github.com/pillarwallet/pillarwallet/pull/1610/files#diff-be672a51bbe0a61a33c68b5e5d4921a5R22

alazarevski: It's used in src/actions/bitcoinActions.js, as well as src/utils/validators.js . This was the minimal change that could be done

PR #1618, opened 2019-12-23
Add more types to screens

grilix:


src/screens/Settings/Settings.js
```diff
@@ -388,6 +389,7 @@ class Settings extends React.Component<Props, State> {
     const { enabled, privateKey } = setBiometrics;
     this.setState({ setBiometrics: null });
     resetIncorrectPassword();
+    // $FlowFixMe - privateKey could be null or undefined
```
jsidorenko: If the private key could be null or undefined why not to mark that param as optional?

grilix: Well yes, we could do that :sweat_smile:. It seems like the optional parameter could be `string` or `undefined` but not `null`, tho, so I updated the references. Please take a look at db87c7f9 to see how it looks like.

jsidorenko: It looks good!
Another option could be to type it as `privateKey?: ?string` - it's when you're totally unsure :D

PR #1620, opened 2019-12-26
Fixes DB race condition

jsidorenko:
This PR solves the race condition between the new and old AsyncStorage libraries. According to this [comment](https://github.com/react-native-community/async-storage/issues/118#issuecomment-500138053) usage of 2 libraries could cause a data loss.
In our case, we still have npm libraries that use an old AsyncStorage version, while the main app uses the new version. 
Here I've updated npm packages that were safe to update and manually patched others by using an amazing `npm-patch` utility.

Bugs that were caused by that race condition:
- auth tokens were not updating properly;
- when a user imports a new wallet after the refresh he gets the previous one.


jsidorenko:
One more thing we can do is to patch this file `node_modules/react-native/Libraries/react-native/react-native-implementation.js` and force to return the newest AsyncStorage version instead of the old one

alazarevski:
> One more thing we can do is to patch this file `node_modules/react-native/Libraries/react-native/react-native-implementation.js` [Line 223] and force to return the newest AsyncStorage version instead of the old one

Good Idea, can we also convert these fixes to patches : node ./android-release-gradle-fix.js && node ./ios-run-simulator-fix.js  ?

jsidorenko:
@grilix Sure thing, I wouldn't rush with this, it's better to test it thoroughly first

jsidorenko:
I tried to patch the RN library itself and replace a reference to the old AsyncStorage with the new one, but had no luck as there occurs a circular dependency between those libraries and I was unable to build the project

jsidorenko:
@alazarevski successfully converted `android-release-gradle-fix.js` and `ios-run-simulator-fix.js` to a new format

PR #1622, opened 2019-12-31
json.parse exception handlers throughout code and other cleanup

poocart:
Covers this Sentry issue https://sentry.io/organizations/pillar-project-worldwide-ltd/issues/1312068619/?project=1294444&query=is%3Aunresolved.

As well:
- Reported issue this to back-end to take a look why notification payload comes as undefined.
- Added JSON.parse exception handlers to other pieces in the code.
- Added missing typings and other cleanups that found along.

### DIFF
--- a/src/actions/notificationsActions.js
+++ b/src/actions/notificationsActions.js
@@ -24,6 +24,7 @@ import firebase from 'react-native-firebase';
 import Intercom from 'react-native-intercom';
 import { NavigationActions } from 'react-navigation';
 import { Alert } from 'react-native';
+import get from 'lodash.get';
 
 // actions
 import { fetchInviteNotificationsAction } from 'actions/invitationsActions';
@@ -164,7 +165,15 @@ export const startListeningNotificationsAction = () => {
     } = getState();
     if (SOCKET && SOCKET.socket && SOCKET.socket.readyState === 1) {
       SOCKET.onMessage(async response => {
-        const data = JSON.parse(response.data.msg);
+        const responseDataMessage = get(response, 'data.msg');
+        let data;
+        try {
+          data = JSON.parse(responseDataMessage);
+        } catch (e) {
+          return; // unable to parse data, do not proceed
+        }
+
+        const senderUserId = get(data, 'senderUserData.id');
 
         if (data.type === CONNECTION_REQUESTED_EVENT) {
           dispatch(fetchInviteNotificationsAction());
@@ -173,7 +182,7 @@ export const startListeningNotificationsAction = () => {
           data.type === CONNECTION_CANCELLED_EVENT ||
           data.type === CONNECTION_REJECTED_EVENT
         ) {
-          const updatedInvitations = invitations.filter(({ id }) => id !== data.senderUserData.id);
+          const updatedInvitations = invitations.filter(({ id }) => id !== senderUserId);
           dispatch({
             type: UPDATE_INVITATIONS,
             payload: updatedInvitations,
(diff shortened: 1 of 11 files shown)


src/actions/notificationsActions.js
```diff
@@ -163,8 +165,17 @@ export const startListeningNotificationsAction = () => {
...
+        let data;
+        try {
+          data = JSON.parse(response.data.msg);
+        } catch (e) {
+          // this shouldn't happen, but was reported to Sentry as issue, let's report with more details
+          Sentry.captureMessage('Platform WebSocket notification parse failed', { extra: { response } });
+          return; // unable to parse data, do not proceed
+        }
+
+        const senderUserId = get(data, 'senderUserData.id');
```
grilix: What would happen if `senderUserId` is `undefined`?

poocart: Hm, it has been like this, but will add additional check here 👌 thanks

poocart: Actually before making more changes here we should ask @alazarevski if it's ok to check whether `senderUserId` is present or those actions need to be called even if it's null / undefined?

src/utils/notifications.js
```diff
@@ -179,3 +182,14 @@ export const processNotification = (notification: Object, myEthAddress?: string)
 
   return result;
 };
+
+export const parseNotificationMessage = (notification: Object): Object => {
+  const createdAt = get(notification, 'createdAt');
```
grilix: Do we really need the `get` here?

Not sure if we expect  the parameter `notification` to be `null`/`undefined`. In that case, we should mark it as so. But then, if `notification` is `null`/`undefined` we can avoid the rest of the flow. Maybe we can make sure the caller passes in something instead of controlling the argument here (or else, just add an `if (!notification) return ...;` at the top of the function).

```js
const mappedInviteNotifications = inviteNotifications
  .filter(notification => !!notification)
  .map(parseNotificationMessage)
  // ...
```

Also, typing the parameter will make flow much helpful on letting us know what to check for:

```js
type RawNotificationMessage = {
  createdAt?: number, // number, I guess
  payload?: {
    msg?: string,
  }
};

export const parseNotificationMessage = (
  notification?: RawNotificationMessage,
): Object => {
  // Controlling the function flow with an argument
  // is normally something to be avoided, but if we
  // really want to make this optional:
  if (!notification) {
    return {};
  }

  const { createdAt, payload = {} } = notification;
  if (payload.msg) {
    try {
      const parsedMessage = JSON.parse(payload.msg);

      return { ...parsedMessage, createdAt };
    } catch (e) {
      //
    }
  }
  return { createdAt };
};
```

src/actions/notificationsActions.js
```diff
@@ -163,8 +165,17 @@ export const startListeningNotificationsAction = () => {
...
     if (SOCKET && SOCKET.socket && SOCKET.socket.readyState === 1) {
-      SOCKET.onMessage(async response => {
-        const data = JSON.parse(response.data.msg);
+      SOCKET.onMessage((response) => {
+        let data;
+        try {
+          data = JSON.parse(response.data.msg);
+        } catch (e) {
+          // this shouldn't happen, but was reported to Sentry as issue, let's report with more details
+          Sentry.captureMessage('Platform WebSocket notification parse failed', { extra: { response } });
```
jsidorenko: We could include `e` object into the `extra` param as well

jsidorenko: and maybe `msg: get(response, 'data.msg')` as Sentry could hit the recursion limit when displaying the raw `response` object

poocart: 👍

src/utils/notifications.js
poocart: I've added notification object as TODO, but will now go through back-end and will try to create a single model, thanks 👍

PR #1623, opened 2020-01-02
Prevent support chat impersonation

grilix:
Use support hmac generated by authentication api.

WIP: We need further changes on the backend, and the sdk needs to be updated as well.


src/actions/notificationsActions.js
```diff
@@ -82,6 +82,10 @@ import { SOCKET } from 'services/sockets';
 import { processNotification } from 'utils/notifications';
 import { STATUS_MUTED } from 'constants/connectionsConstants';
 
+// types
+import type { Dispatch, GetState } from 'reducers/rootReducer';
+import SDKWrapper from 'services/api';
```
poocart:
```suggestion
import type SDKWrapper from 'services/api';
```

PR #1636, opened 2020-01-07
Fix/api request timeout config

poocart:
Includes:
- Added `timeout` for all wallet http requests. This required a change from `fetch` to `axios` library and newer version of `axios` as well (older version was not supported with our version React Native native bridges/wrappers). With the change to `axios` most of the tests and request calls were changes as well (a note on this that `fetch` used `json()` to get results while `axios` already returns json formatted result in `data` key).
- Removed old `okhttp` and `okio` http requester libs from Android as React Native main package uses newer versions and those are redundant.
- As always – cleanup that found along.


src/actions/__tests__/exchangeActions.test.js
```diff
@@ -66,42 +67,32 @@ const sendWyreOfferMock = {
...
-  return {
-    provider: PROVIDER_MOONPAY,
-  };
-});
-pillarSdk.fetchSendWyreOffers = jest.fn(async () => {
-  return {
-    provider: PROVIDER_SENDWYRE,
-  };
-});
+pillarSdk.fetchMoonPayOffers = jest.fn(() => Promise.resolve(({ provider: PROVIDER_MOONPAY })));
```
jsidorenko:
```suggestion
pillarSdk.fetchMoonPayOffers = jest.fn(() => Promise.resolve({ provider: PROVIDER_MOONPAY }));
```

src/actions/__tests__/exchangeActions.test.js
```diff
@@ -66,42 +67,32 @@ const sendWyreOfferMock = {
...
+pillarSdk.fetchSendWyreOffers = jest.fn(() => Promise.resolve(({ provider: PROVIDER_SENDWYRE })));
```
jsidorenko:
```suggestion
pillarSdk.fetchSendWyreOffers = jest.fn(() => Promise.resolve({ provider: PROVIDER_SENDWYRE }));
```

src/actions/exchangeActions.js
```diff
@@ -242,10 +242,12 @@ export const searchOffersAction = (fromAssetCode: string, toAssetCode: string, f
...
       );
       // we're requesting although it will start delivering when connection is established
-      const { error } = await exchangeService.requestOffers(fromAddress, toAddress, fromAmount);
+      const response = await exchangeService.requestOffers(fromAddress, toAddress, fromAmount);
+      const responseError = get(response, 'error');
 
-      if (error) {
-        const message = error.message || 'Unable to connect';
+      if (responseError) {
+        const responseErrorMessage = get(responseError, 'response.data.error.message');
```
jsidorenko: In 2 lines above you're checking the 'response.error' while here you're grabbing the message from 'response.data.error'

poocart: Yes, this is correct because on call exception we return `{ error: axiosErrorResponse }` of which `axiosErrorResponse` is AxiosErrorResponse type from which `response.data` is axios built response and 'error.message' in 'data` is value returned from back-end.

jsidorenko: Oh, you're right! Didn't notice that you're taking it from the `responseError` object, sorry

PR #1638, opened 2020-01-08
Fix/synthetics send ui updates

poocart:
Updated UI for synthetic send flow. Main update – allow to calculate if there's enough PLR in tank on the fly while editing synthetic asset amount.


src/components/HeaderBlock/HeaderBlock.js
```diff
@@ -256,16 +250,20 @@ class HeaderBlock extends React.Component<Props> {
     if (item.user || item.userIcon) {
       return this.renderUser(!item.userIcon);
     }
-    if (item.title) {
+    if (item.title || item.customTitle) {
```
vyvil: Header block has already a way to have custom header - it was designed to accept various components in an array, so we could combine components in any way needed. 

Headers that need an icon in the middle of it, for example, could be created like this:
```headerProps={{ centerItems: [{ title: 'Send' }, { custom: (<ImageIcon source={lightningIcon} />) }, { title: symbol, color: baseColors.electricBlueIntense }] }}```

In such way we would not need an additional (almost single use) component.

File: src/screens/SendSynthetic/SendSyntheticAmount.js [hunk 2]
```diff
@@ -54,10 +55,11 @@ import { SEND_SYNTHETIC_CONFIRM, SEND_TOKEN_CONFIRM } from 'constants/navigation
 import { accountAssetsSelector } from 'selectors/assets';
 
 // models, types
-import type { Asset, Assets, Rates } from 'models/Asset';
+import type { Asset, AssetData, Assets, Rates, SyntheticAsset } from 'models/Asset';
 import type { SyntheticTransaction, TokenTransactionPayload } from 'models/Transaction';
 import type { Dispatch, RootReducerState } from 'reducers/rootReducer';
-import { fetchSingleAssetRatesAction } from 'actions/ratesActions';
+import HeaderSyntheticAssetTitle from 'components/HeaderBlock/HeaderSyntheticAssetTitle';
```
jsidorenko: Minor: pls move to the components section

jsidorenko:
I was unable to get the synthetics data/exchange rates locally due to the timeout setting :(

jsidorenko:
image
I think we needed the rates in order to display what is the maximum amount of other tokens I can send having xxx PLR. It could be more than we have in the liquidity pool, so on the next screen we can show info about liquidity pool and adjust the max amount user can send

jsidorenko:
Just found that we're using 5 seconds timeout, let's try to increase it to 10 seconds?

src/components/HeaderBlock/HeaderBlock.js
poocart: I'm sure I looked for this, but missed it somehow, thanks! 👍

PR #1643, opened 2020-01-14
Feature/theme toggle

vyvil:
This PR adds a way to toggle theme:

- dev way to toggle theme can now be toggled by adding value to `SHOW_THEME_TOGGLE` on env;
- there are tree options for app's theme: `Light`, `Dark` and `System's default`. If later one is selected, theme gets updated if user changes mode on his/her device.
- on first open, an alert is shown to those users that are set dark theme on their phones showing that dark mode is available on the wallet as well.

PR can be reviewed now. Yet it would be better to merge that in when all themes related PRs are in.


App.js
```diff
@@ -163,43 +192,49 @@ class App extends React.Component<Props, *> {
...
+        <ThemeProvider theme={theme}>
+          <React.Fragment>
+            <Root>
+              <RootNavigation
+                ref={(node) => {
+                  if (!node) return;
+                  setTopLevelNavigator(node);
+                }}
+              />
+              {!!SHOW_THEME_TOGGLE &&
```
grilix: :clap:

App.js
```diff
@@ -150,6 +159,26 @@ class App extends React.Component<Props, *> {
...
+        'Would you like to turn on Dark mode now? You can always switch between modes in settings.',
+        [
+          { text: 'Cancel', onPress: () => markThemeAlertAsShown() },
+          {
+            text: 'Turn on',
+            onPress: () => changeAppTheme(DARK_THEME),
+          },
+        ],
+      );
+    }
```
poocart: I know these kind of comments are annoying, but maybe just `if (defaultPreference !== DARK_PREFERENCE) return`? 😄

src/actions/appActions.js
```diff
@@ -224,12 +223,18 @@ export const initAppAndRedirectAction = (appState: string, platform: string) =>
 
       // check if current user has theme set and set it to default if
       const hasTheme = get(appSettings, 'themeType');
+      // check if theme is set to system's default
+      const isThemeSetAsSystemDefault = get(appSettings, 'isSetAsSystemPrefTheme');
```
poocart: How about moving this to line 234 where the logic happens?

src/actions/walletActions.js
```diff
@@ -39,19 +39,21 @@ import shuffle from 'shuffle-array';
...
-  return async (dispatch: Function, getState: () => Object, api: Object) => {
+  return async (dispatch: Dispatch, getState: GetState, api: SDKWrapper) => {
     try {
       const importedWallet = ethers.Wallet.fromMnemonic(tWordsPhrase);
 
       api.init();
       let apiUser = {};
       const addressValidationResponse = await api.validateAddress(importedWallet.address);
-      if (addressValidationResponse.walletId) {
+      if (!addressValidationResponse.error) {
```
poocart: I assume you carefully checked with this one as we had issues here.

src/screens/Settings/Settings.js
```diff
@@ -465,12 +527,13 @@ class Settings extends React.Component<Props, State> {
     );
   };
 
-  renderListItem = (field: string, onSelect: Function) => ({ item: { name } }: Object) => {
+  renderListItem = (field: string, onSelect: Function, currentValue) => ({ item: { name, value } }: Object) => {
```
poocart: maybe you could add `currentValue` type as well?

App.js
alazarevski: I'd say it's ok without it on functions with less than 20 lines of code

PR #1649, opened 2020-01-23
Feature/smart wallet migration bugfixes

jsidorenko:
Various bugfixes, probably it's better to review them going through each commit:

- Make Save button visible
- Reset user events on new wallet import
- Fix disappearance of connected smart account in reducer
- Store devices in the right format
- Fetch initial assets after wallet migration
- Addition to the previous commit
- Update transaction status for the key based wallet during wallet migration
- Fix smart wallets registration on the backend logic during wallet's migration

### DIFF
--- a/src/actions/smartWalletActions.js
+++ b/src/actions/smartWalletActions.js
@@ -103,7 +103,7 @@ import { accountHistorySelector } from 'selectors/history';
 
 // actions
 import {
-  addNewAccountAction,
+  addAccountAction,
   setActiveAccountAction,
   switchAccountAction,
 } from 'actions/accountsActions';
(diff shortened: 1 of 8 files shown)


src/actions/smartWalletActions.js
```diff
@@ -324,7 +331,7 @@ export const deploySmartWalletAction = () => {
     const account = await smartWalletService.fetchConnectedAccount();
     dispatch({
       type: SET_SMART_WALLET_CONNECTED_ACCOUNT,
-      account,
+      payload: account,
```
poocart: 👍

src/services/smartWallet.js
```diff
@@ -186,6 +189,31 @@ class SmartWallet {
...
+  async syncSmartAccountsWithBackend(
+    api: SDKWrapper,
+    smartAccounts: SmartWalletAccount[],
+    walletId: string,
+    privateKey: string,
+    fcmToken: string,
+  ) {
+    const backendAccounts = await api.listAccounts(walletId);
+    const registerOnBackendPromises = smartAccounts.map(async account => {
+      const backendAccount = backendAccounts.find(({ ethAddress }) => addressesEqual(ethAddress, account.address));
```
poocart:
```suggestion
      const backendAccount = backendAccounts.some(({ ethAddress }) => addressesEqual(ethAddress, account.address));
```

src/actions/smartWalletActions.js
grilix: Damn, good catch. This kind of things hurt my soul 😓.

PR #1650, opened 2020-01-25
Fix ERC20 transactions showing as collectibles

grilix:
OpenSea returns events for some transactions of some ERC20 assets. This filters them out so they don't show up on the transaction history as collectibles.
ERC20 transactions still show up as asset transfers.

### DIFF
--- a/src/services/api.js
+++ b/src/services/api.js
@@ -390,7 +390,7 @@ SDKWrapper.prototype.fetchCollectibles = function (walletAddress: string) {
 };
 
 SDKWrapper.prototype.fetchCollectiblesTransactionHistory = function (walletAddress: string) {
-  const url = `${OPEN_SEA_API}/events/?account_address=${walletAddress}&exclude_currencies=true&event_type=transfer`;
+  const url = `${OPEN_SEA_API}/events/?account_address=${walletAddress}&event_type=transfer`;
   return Promise.resolve()
     .then(() => axios.get(url, {
       ...defaultAxiosRequestConfig,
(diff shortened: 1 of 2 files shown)


src/services/api.js
```diff
@@ -390,7 +390,7 @@ SDKWrapper.prototype.fetchCollectibles = function (walletAddress: string) {
 };
 
 SDKWrapper.prototype.fetchCollectiblesTransactionHistory = function (walletAddress: string) {
-  const url = `${OPEN_SEA_API}/events/?account_address=${walletAddress}&exclude_currencies=true&event_type=transfer`;
+  const url = `${OPEN_SEA_API}/events/?account_address=${walletAddress}&event_type=transfer`;
```
jsidorenko: Why do you want to remove the `exclude_currencies=true&` part? Opensea guys asked us to add this param to the query url

grilix: Oh, what does this parameter do? I couldn't find anything on their documentation and removing it seemed to have no effect, maybe I'm not looking into the right place?

The name `exclude_currencies` sounds like it's not gonna return events for currencies (ERC20 tokens?), but it's not doing so. I wonder if this parameter is either something totally unrelated or deprecated 🤔, is there any place we can look for some details? Or maybe we just add it back?

jsidorenko: Let just add it back :) this param should exclude providing us information about user's ERC-20 tokens

grilix: Okay, added back

PR #1652, opened 2020-01-29
Fix double Choose Assets screen in smart wallet migration flow

graszka22:
To reproduce:
- enter the smart wallet migration flow
- select an asset
- click edit in the upper right corner
- go back to asset selection
- the previously selected asset is unselected
In fact when you go back once again, you are brought to the choose asset screen once again, with selected assets. That means that the choose asset screen is opened once again when we press Edit.
Why? When we go to the `appNavigation.js`we see two stack navigators:
```js
const smartWalletUpgradeFlow = createStackNavigator({
  [CHOOSE_ASSETS_TO_TRANSFER]: ChooseAssetsScreen,
...
  [EDIT_ASSET_AMOUNT_TO_TRANSFER]: EditAssetAmountScreen,
...
```
```js
const manageWalletsFlow = createStackNavigator({
  [ACCOUNTS]: AccountsScreen,
  ...
  [CHOOSE_ASSETS_TO_TRANSFER]: ChooseAssetsScreen,
}, StackNavigatorConfig);
```
So, in the smart wallet migration flow we firstly go to the `manageWalletsFlow` to the `AccountsScreen`, then we move on to the `ChooseAssetsScreen` in the same navigator, then it messes up as it goes to the `EditAssetAmountScreen` in another navigator that also contains `ChooseAssetsScreen`. 

So the solution might be moving explicitly to the `smartWalletUpgradeFlow` while moving to the `ChooseAssetsScreen` and it fixes this problem. We can also remove the `ChooseAssetsScreen` from one of the navigators but I guess it's there for a reason and we move to it from a couple of screens.

### DIFF
--- a/src/navigation/appNavigation.js
+++ b/src/navigation/appNavigation.js
@@ -567,7 +567,6 @@ const manageWalletsFlow = createStackNavigator({
   [ACCOUNTS]: AccountsScreen,
   [FUND_CONFIRM]: FundConfirmScreen,
   [RECOVERY_AGENTS]: RecoveryAgentsScreen,
-  [CHOOSE_ASSETS_TO_TRANSFER]: ChooseAssetsScreen,
 }, StackNavigatorConfig);
 
 manageWalletsFlow.navigationOptions = hideTabNavigatorOnChildView;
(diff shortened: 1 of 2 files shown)


jsidorenko:
Good spot! Will check tomorrow if we could remove `CHOOSE_ASSETS_TO_TRANSFER` from the `manageWalletsFlow`

jsidorenko:
Yes, we can safely remove `CHOOSE_ASSETS_TO_TRANSFER` from the `manageWalletsFlow`

PR #1654, opened 2020-01-29
Revert user settings transition speed on iOS

graszka22:


graszka22:
@poocart It was introduced in this commit: https://github.com/pillarwallet/pillarwallet/commit/59184d544c1b1c2bad50943476b49fef9a91d0f3. I think it was supposed to make the transition to this screen consistent with other transitions (on iOS sliding from the right side). But I don't know why there is a duration set to 500ms on iOS, especially when the boolean says that the animation should be faster and on all other screens it's 400ms. Probably @vyvil will know :) So I've changed it back to 400ms to be consistent with all other screens (on Android I've left 250ms because it really feels better than 400ms).

poocart:
@graszka22 thanks for providing insights! Only one thing to add is maybe the duration was increased to "hold" faster devices during transition, but totally agree with you that doesn't make any sense why from 400 to 500. Anyway, let's see how it will look in test cycle.

PR #1656, opened 2020-01-29
Disable vertical scroll when swiping item in People list

graszka22:


graszka22:
@poocart in the `screens/Assets/AssetsList` it's already fixed.

PR #1657, opened 2020-01-30
Avoid error on WC personal_sign

grilix:
There are scenarios where the message provided is not parseable, causing
the popup to break. This shows the raw message instead of failing.

We need to improve this later.


poocart:
What do you think of logging this to Sentry? I think "message" part is not sensitive.

grilix:
@poocart Well, the message is the part they are signing, it could contain their name, or email, for example, so I'd say we shouldn't save it anywhere.

I'm not entirely sure if having this workaround would even make sense, tbh. I'm looking for improvements to this approach, but at least this will avoid the exception to blow the app in the meantime.

PR #1658, opened 2020-01-31
PIL-244 Add buy tokens button in receive modal

graszka22:


vyvil:
With this additional section in the slide modal it gets too tall for smaller screens (iPhone SE, for example). (Small screens are not that common, but still :) )
Maybe updating all Receive modal UI as it is designed [here](https://app.zeplin.io/project/5db9b5d14c30ae56faaa3030/screen/5de535f8f29502786fbc7f28) would help to fit all items in. Or we might need scrollview.

PR #1659, opened 2020-01-31
PIL-245 Add services tab

graszka22:
About changes in the Exchange screen:
I've moved the initialization of the selection of assets from `componentDidUpdate` to `componentDidMount` because it was causing infinite loop of re-renders. I think it's not necessary anyway to check the navigation params every time the component updates since they're not changed anywhere now - they're just passed from the previous screen and that will be handled in `componentDidMount`.


src/screens/Services/Services.js
```diff
@@ -0,0 +1,123 @@
...
+import type { RootReducerState } from 'reducers/rootReducer';
+
+type Props = {
+    theme: Theme,
+    offers: Offer[],
+    navigation: NavigationScreenProp<*>,
+};
+
+class ServicesScreen extends React.Component<Props> {
+    getServices = () => {
```
jsidorenko: Pls check your IDE settings, we're using 2 spaces as default indentation

src/screens/Services/package.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Services.js",
```
jsidorenko: Same as above

src/screens/Services/Services.js
```diff
@@ -0,0 +1,123 @@
...
+            keyExtractor={(item) => item.key}
+            renderItem={this.renderServicesItem}
+            contentContainerStyle={{ width: '100%', padding: 20, paddingBottom: 40 }}
+          />
+        </ContainerWithHeader>
+      );
+    }
+}
+
+const mapStateToProps = ({ exchange: { data: { offers } } }: RootReducerState): $Shape<Props> => ({
```
jsidorenko:
```suggestion
const mapStateToProps = ({
  exchange: { data: { offers } },
}: RootReducerState): $Shape<Props> => ({
```

src/screens/Services/Services.js
```diff
@@ -0,0 +1,123 @@
...
+  offers,
+});
+
+export default withTheme(connect(mapStateToProps)(ServicesScreen));
```
jsidorenko: Missed a new line at the end of file

src/screens/Services/Services.js
```diff
@@ -0,0 +1,123 @@
...
+import type { NavigationScreenProp } from 'react-navigation';
+import Intercom from 'react-native-intercom';
+import { ListCard } from 'components/ListItem/ListCard';
+import ContainerWithHeader from 'components/Layout/ContainerWithHeader';
+import { EXCHANGE } from 'constants/navigationConstants';
+import { USD, ETH } from 'constants/assetsConstants';
+import { getThemeColors } from 'utils/themes';
+import type { Theme } from 'models/Theme';
+import type { Offer } from 'models/Offer';
+import { withTheme } from 'styled-components/native';
```
jsidorenko: @vyvil here and in Assets we're importing `withTheme` from 'styled-components/native' while on other screens importing it from 'styled-components' - which route is correct?

vyvil: There is no big difference - it's the same withTheme being exported from 'styled-components' and 'styled-components/native';

For a sake of consistency we could stick to 'styled-components/native' (as we get our configured `styled` from there as well.

vyvil:
As it is a quick one, I'll push a small addition that is kind of related to the changes - it will fix an navigation issue:
If user navigates to exchange from asset screen and then tries to get back, he/she is taken to Services tab rather than asset screen.

graszka22:
I've updated the code, fixing code style. As a side note, shouldn't such things be listed in the `.eslint.rc`? That way it'll be formatted and checked automatically. I tried to add it, but there are a lot of files that don't have correct indentation and I'd update half of the project 😜

jsidorenko:
We have the '.editorconfig' file for this

src/screens/Services/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Services.js",
+  "version": "0.0.0",
+  "private": true,
+  "main": "./Services.js"
+}
```
grilix: This one is also missing a newline at the end.

src/screens/Services/Services.js
```diff
@@ -0,0 +1,126 @@
...
+        title: 'Peer-to-peer trading',
+        body: 'Swap tokens directly with others. safe, secure, anonymous',
+        disabled: true,
+        label: 'soon',
+      },
+      {
+        key: 'buyCryptoWithFiat',
+        title: 'Buy crypto with fiat',
+        body: 'USD, GBP, EUR supported',
+        action: () => navigation.navigate(EXCHANGE, { fromAssetCode: USD, toAssetCode: ETH }),
```
vyvil: `fromAssetCode` could be default currency - it is either baseFiatCurrency (if user has set it), that is stored in appSettings, or defaultFiatCurrency (which is a constant set in assetsConstants.js).

src/screens/Services/Services.js
```diff
@@ -0,0 +1,126 @@
...
+          rightItems: [{ link: 'Support', onPress: () => Intercom.displayMessenger() }],
+          leftItems: [{ title: 'Services' }],
+        }}
+        inset={{ bottom: 'never' }}
+      >
+        <FlatList
+          data={services}
+          keyExtractor={(item) => item.key}
+          renderItem={this.renderServicesItem}
+          contentContainerStyle={{ width: '100%', padding: 20, paddingBottom: 40 }}
```
vyvil: I believe, horizontal paddings should be 16px - `spacing.layoutSides`

src/components/LabelBadge/LabelBadge.js
```diff
@@ -26,10 +26,11 @@ type Props = {
   label: string,
   containerStyle?: Object,
   labelStyle?: Object,
+  color?: string,
 }
 
 const BadgeWrapper = styled.View`
-  background-color: ${themedColors.positive};
+  background-color: ${props => props.color || themedColors.positive};
```
poocart:
```suggestion
  background-color: ${({ color }) => color || themedColors.positive};
```

src/components/ListItem/ListCard.js
```diff
@@ -120,6 +126,12 @@ export const ListCard = (props: Props) => {
             <TitleWrapper>
               <CardTitle style={titleStyle}>{title}</CardTitle>
               {!!label && <Label>{label}</Label>}
+              {!!labelBadge && (
+                <LabelBadge
+                  label={labelBadge.label}
+                  labelStyle={{ fontSize: fontSizes.regular }}
+                  color={labelBadge.color}
+                />)}
```
poocart:
```suggestion
                />
              )}
```

PR #1660, opened 2020-01-31
fix balance check between different bignumber libs

poocart:
Fixes issue where BigNumber libraries differs between calls during wallet deployment check if wallet has enough balance.


src/actions/smartWalletActions.js
```diff
@@ -296,7 +296,7 @@ export const deploySmartWalletAction = () => {
     const gasInfo = get(getState(), 'history.gasInfo', {});
     const deployEstimateFee = await smartWalletService.estimateAccountDeployment(gasInfo);
     const deployEstimateFeeBN = new BigNumber(utils.formatEther(deployEstimateFee.toString()));
-    const etherBalanceBN = smartWalletService.getAccountRealBalance();
+    const etherBalanceBN = new BigNumber(smartWalletService.getAccountRealBalance().toString());
```
jsidorenko: o_O good catch!

PR #1661, opened 2020-01-31
Fix walletconnect signing functions

grilix:


src/utils/wallet.js
```diff
@@ -76,7 +76,7 @@ export function catchTransactionError(e: Object, type: string, tx: Object) {
 
 // handle eth_signTransaction
 export function signTransaction(trx: Object, wallet: Object): Promise<string> {
-  wallet.connect(providers.getDefaultProvider(NETWORK_PROVIDER));
+  wallet.connect(ethers.getDefaultProvider(NETWORK_PROVIDER));
```
jsidorenko: How about using `getEthereumProvider` from 'utils/common'?

grilix: Great, I didn't realise we had that, it works as expected. Updated :+1:

PR #1662, opened 2020-02-03
PIL-25 Search assets locally if available

grilix:
When searching for assets, perform a local filter instead of going to
the API. If the local search returns no results, fallback to the API.


grilix:
I'm not reallly sure if going to the API makes sense even if we couldn't
find a match locally, tho. What do you think? :thinking:

src/actions/assetsActions.js
```diff
@@ -631,13 +631,30 @@ export const startAssetsSearchAction = () => ({
...
 export const searchAssetsAction = (query: string) => {
   return async (dispatch: Dispatch, getState: GetState, api: SDKWrapper) => {
+    const { assets: { supportedAssets } } = getState();
+    const search = query.toUpperCase();
+
+    const filteredAssets = supportedAssets.filter(({ name, symbol }) => {
+      return name.toUpperCase().includes(search) || symbol.toUpperCase().includes(search);
+    });
+
+    if (filteredAssets.length > 0) {
```
poocart: We had some issues with assets list object before, I would switch it to `lodash.isempty`. However, the issue was due API `fetch` requester failing to produce timeout as exception and after a switch to `axios` this shouldn't happen anymore, but since we're putting this to reducer I would go with `lodash.isempty` here.

jsidorenko: `supportedAssets.filter()` always return an array, so we could be safe here

jsidorenko: The problem could be if the `supportedAssets` itself is not an array, but we should fix that in reducer then and keep the state consistent

jsidorenko:
Very good idea!!
Initially, we've planned to store the huge list of supported assets on the backend and don't download it on every login. But since we keep the full and fresh list of all supported assets in memory - we don't even need to do any api calls

PR #1663, opened 2020-02-04
Feature/exchange ui 2point0 [PIL-256]

vyvil:
This PR adds first part of Exchange UI changes:
1. TextInput is updated so it accepts labels, UI for text inputs with selector item and selectors' modal were updated according latest design: https://zpl.io/VY7g4dn and https://zpl.io/bW0JwJM.
2. Exchange empty state is updated to either show "Promo" section (before user starts searching) or empty state message. Both of those states needs further improvements - animated providers logo carousel (promo block) and button to alert user when there are available offers (empty state).
3. Activity indicator has been updated by changing its animation and position.


src/screens/Exchange/ExchangeStatus.js
```diff
@@ -18,50 +18,112 @@
...
+  }
+
+  componentDidMount = () => {
+    const { isVisible } = this.props;
+    if (isVisible) this.startBlinking();
+  };
+
+  startBlinking = () => {
+    const { indicatorFadeValue } = this.state;
+    requestAnimationFrame(() => {
```
grilix: I'm not sure I understand why we want to call `requestAnimationFrame` here; The `requestAnimationFrame` will tell the browser we want to perform an animation, but we're not, we're gonna just setup a timer that's gonna be called on a later tick. Correct?

src/components/SearchBar/SearchBar.js
```diff
@@ -71,26 +65,40 @@ const InputIcon = styled(IconButton)`
...
 
-type Props = {
-  errorMessage?: string,
-  inputProps: inputPropsType,
+type CommonComponentsProps = {
+  inputProps: InputPropsType,
   placeholder?: string,
   backgroundColor?: string,
+  inputRef?: TextInput,
+}
```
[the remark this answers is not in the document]
vyvil: ;(

src/screens/Exchange/ExchangeStatus.js
vyvil: Oh, yeah, you are right. 
Since we have a lot going on on exchange screen, I want to make sure that this blinking animation won't be glitchy. But this request should be inside a loop as well to be effective. I'll update that part. Thanks :)

src/screens/Exchange/ExchangeStatus.js
```diff
@@ -18,50 +18,111 @@
...
+  }
+
+  componentDidMount = () => {
+    const { isVisible } = this.props;
+    if (isVisible) this.startBlinking();
+  };
+
+  startBlinking = () => {
+    const { indicatorFadeValue } = this.state;
+    this.blinkInterval = setInterval(() => {
```
graszka22: maybe we can use this: https://facebook.github.io/react-native/docs/animated#loop?

src/components/TextInput/TextInput.js
```diff
@@ -415,44 +423,43 @@ class TextInput extends React.Component<Props, State> {
...
+        key={name}
+        onPress={() => this.selectValue(item)}
+      >
+        <ProfileImage
+          uri={iconUri}
+          userName={name}
+          diameter={64}
+          textStyle={{ fontSize: fontSizes.medium }}
+          noShadow
+          borderWidth="0"
```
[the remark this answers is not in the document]
vyvil: Yeah, I'll update this. Thanks :)

src/components/TextInput/TextInput.js
```diff
@@ -667,47 +665,53 @@ class TextInput extends React.Component<Props, State> {
...
+                <SearchBar
+                  inputProps={{
+                    onChange: this.handleSearch,
+                    value: query,
+                    autoCapitalize: 'none',
+                  }}
+                  placeholder={optionsSearchPlaceholder}
+                  inputRef={ref => { this.searchInput = ref; }}
+                  noClose
+                  marginBottom="0"
```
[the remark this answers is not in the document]
vyvil: Here - no. As it gets ignored - 0 is understood as false here: 
`margin-bottom: ${props => props.marginBottom || 20}px;`

src/screens/Exchange/ExchangeStatus.js
vyvil: Yeah we can - will push a change with it.

src/screens/Exchange/ExchangeStatus.js
```diff
@@ -18,50 +18,120 @@
...
+            duration: 500,
+          },
+        ).start();
+      });
+      this.startBlinking();
+    }
+  }
+
+  componentWillUnmount = () => {
+    clearInterval(this.blinkInterval);
```
graszka22: `this.blinkInterval` is not needed anymore now :)

src/screens/Exchange/ExchangeStatus.js
```diff
@@ -18,50 +18,120 @@
...
+  }
+
+  componentDidMount = () => {
+    const { isVisible } = this.props;
+    if (isVisible) this.startBlinking();
+  };
+
+  startBlinking = () => {
+    const { indicatorFadeValue } = this.state;
+    requestAnimationFrame(() => {
```
graszka22: If we use native driver here, do we still need that `requestAnimationFrame`? The animation runs on native thread, without bothering the js thread.

src/screens/Exchange/ExchangeStatus.js
```diff
@@ -18,50 +18,120 @@
...
+    const { isVisible } = this.props;
+    const { statusFadeValue } = this.state;
+    const toValue = isVisible && !prevProps.isVisible ? 1 : 0;
+    if (isVisible !== prevProps.isVisible) {
+      requestAnimationFrame(() => {
+        Animated.timing(
+          statusFadeValue,
+          {
+            toValue,
+            duration: 500,
```
graszka22: we can use native driver here as well, we're just animating opacity

vyvil:
@graszka22, 
Keyboard issue is fixed.
I've also addressed later comments of yours.

In terms of exchange form reset - fair point. I'll talk with Dmitry.
But I'd leave this change out of this task's scope.

PR #1664, opened 2020-02-05
Feature/git tags and versioning

antonioned:
This commit will:

1. Use GitHub actions to fetch the latest tag, bump the patch number for develop merges and the minor number for master merges and push that new tag in the repo
1.1 For hockeyapp we will only create tags, we **won't** create releases and attach artifacts.

2. CircleCI will fetch the new tags and attach them in the corresponding jobs (hockey or prod) to version the builds (the apps in the stores will be versioned with it, e.g. v1.1.3)

3. The ghr tool will then use the created tag and create a release for it along with attaching the production ios and android artifacts to the release which will complete the process.


.github/workflows/tag-qa.yml
```diff
@@ -0,0 +1,17 @@
+name: Bump version QA
+on:
+  push:
+    branches:
+      - feature/git_tags_and_versioning
```
grilix: I don't understand why we have this branch name here. Are we supposed to point this to another branch from time to time? 🤔

antonioned: nice catch @grilix this is why PRs exist :D changing it now

grilix: Glad to hear, I was getting worried 😅

antonioned: @grilix fixed!

poocart:
P. S. Maybe I would try to leverage YAML file variables more for workspace filenames as it's easy to break things when changing/adding texts – just a room for improvement.

antonioned:
> P. S. Maybe I would try to leverage YAML file variables more while using workspace filenames as it's easy to break things when changing/adding manual text filenames throughout the file – just a room for improvement.

@poocart thanks for the comment, can you please just point to where and what exactly do you suggest here?

PR #1665, opened 2020-02-05
PIL-8 bFix/add trackid header

alazarevski:
Add trackid (lowercase) header param extracted from the offer.trackId on takeOffer call


src/models/Offer.js
```diff
@@ -54,6 +55,7 @@ export type OfferRequest = {
   provider: string,
   fromAssetAddress: string,
   toAssetAddress: string,
+  trackId: string,
```
poocart: Do we have it for fiat offers as well? 🤔

alazarevski: I don't think so. The changes show it as in Fiat, but the param is added to OfferRequest.

src/services/exchange.js
```diff
@@ -143,7 +143,7 @@ export default class ExchangeService {
     return axios.post(
       buildApiUrl('orders', '2.0'),
       JSON.stringify(order),
```
poocart: Remove `trackId` from payload here?

src/models/Offer.js
poocart: Yeah, sorry, mislooked! All good here 👍

src/services/exchange.js
alazarevski: 06824ea

PR #1666, opened 2020-02-06
Improve people search responsiveness

grilix:
When searching for people, local results (contacts, and in the future we
could even include cached results) are shown immediately instead of
waiting for the API response.


src/actions/contactsActions.js
```diff
@@ -48,22 +48,26 @@ import type { Dispatch, GetState } from 'reducers/rootReducer';
...
 export const searchContactsAction = (query: string) => {
   return async (dispatch: Dispatch, getState: GetState, api: Object) => {
-    const { user: { data: { walletId } }, contacts: { data: localContacts } } = getState();
+    const {
+      user: { data: { walletId } },
+      contacts: { data: localContacts },
+    } = getState();
+    const upperCaseQuery = query.toUpperCase();
+
+    const myContacts = localContacts.filter(({ username }) => {
```
poocart:
```suggestion
    const myContacts = localContacts.filter(({ username }) => username.toUpperCase().includes(upperCaseQuery));
```

src/actions/contactsActions.js
```diff
@@ -48,22 +48,26 @@ import type { Dispatch, GetState } from 'reducers/rootReducer';
...
+      return username.toUpperCase().includes(upperCaseQuery);
```
poocart:
```suggestion
```

src/actions/contactsActions.js
```diff
@@ -48,22 +48,26 @@ import type { Dispatch, GetState } from 'reducers/rootReducer';
...
+    });
```
poocart: Per suggestion above:
```suggestion
```

src/actions/contactsActions.js
poocart: Or you can even use `isCaseInsensitiveMatch` from utils.

grilix: Well actually, no, `isCaseInsensitiveMatch` doesn't provide the same behaviour :grimacing:

PR #1670, opened 2020-02-07
Validate identity keys related api responses

jsidorenko:
1) Validate get identity keys api response type. [Sentry report](https://sentry.io/organizations/pillar-project-worldwide-ltd/issues/1487473922/?environment=production&project=1289773&referrer=alert_email)

2) Validate patchIdentityKeys api response.  [Sentry report](https://sentry.io/organizations/pillar-project-worldwide-ltd/issues/1483651652/?environment=production&project=1289773&referrer=alert_email)


jsidorenko:
@poocart I think that validation should be done inside the SDK, so we could be sure that if the array should be returned by specification - it will be returned and nothing else.
I'm putting those Sentry reports in order to see what actually is being returned, later we could remove them and add response validation to the sdk

PR #1672, opened 2020-02-07
Validate wrong supported assets response

jsidorenko:
Fix for this [Sentry report](https://sentry.io/organizations/pillar-project-worldwide-ltd/issues/1422891164/?environment=production&project=1289773&referrer=alert_email)


poocart:
I guess e should remove duplicate of this from `assetsActions` then, what do you think?

Additionally I believe the report is already fixed as we're catching timeout error with `axios` (there's Sentry report that has timeout error message as result), but it's not yet publicly released.

jsidorenko:
yep, this is a temporary fix in order to see if the axios fix helped and don't crash the app if not.

jsidorenko:
Removed the old check we had

PR #1673, opened 2020-02-07
Handle missed icon in wallet connect ClientMeta

jsidorenko:
Closes this [report](https://sentry.io/organizations/pillar-project-worldwide-ltd/issues/1439666379/?environment=production&project=1294444&referrer=alert_email). As it turns out the 'icons' property could be not set in the ClientMeta

### DIFF
--- a/src/screens/ManageDetailsSessions/ManageDetailsSessions.js
+++ b/src/screens/ManageDetailsSessions/ManageDetailsSessions.js
@@ -20,6 +20,7 @@
 import * as React from 'react';
 import { FlatList } from 'react-native';
 import { connect } from 'react-redux';
+import get from 'lodash.get';
 import styled from 'styled-components/native';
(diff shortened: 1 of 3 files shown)


src/screens/ManageDetailsSessions/ManageDetailsSessions.js
```diff
@@ -212,7 +214,10 @@ class ManageDetailsSessions extends React.Component<Props, State> {
   }
 }
 
-const mapStateToProps = ({ user: { data: user }, walletConnect: { connectors, requests } }) => ({
+const mapStateToProps = ({
+  user: { data: user },
+  walletConnect: { connectors, requests },
+}) => ({
```
poocart: To continue on this change we can add Reducer type here.

jsidorenko: Good suggestion!

PR #1674, opened 2020-02-10
PIL-216, PIL-215 Update flow and eslint

alazarevski:
Enables eslint and flow without errors
TODO:  check packages.json for jest run.


jsidorenko:
What problem did you try to solve here? And why do you downgrading the flow?

poocart:
I think we can just close this branch.

PR #1675, opened 2020-02-10
PIL-269, PIL-270 Add storybook

graszka22:
There are a couple of patches and workarounds due to the bugs in the react-native storybook (for example https://github.com/storybookjs/react-native/issues/13, https://github.com/storybookjs/react-native/issues/19 but also issues that are not even reported).

I've put two stories just to see if storybook is working. We'll remove them once we put our first component in the storybook. You can also put stories in other directories, for example the `components` subdirectories. There is a script that treats every `stories.js` file as a story.

To start the storybook run: `yarn storybook`.
This should start a storybook server and open a new tab in your browser.
To enter storybook on your device go to Settings -> Storybook.
You can also set the env variable `SHOW_ONLY_STORYBOOK` - so you don't need to enter the settings every time you reload the app (with live reload this makes developing a lot quicker!). In the browser or on your device you can now switch between components. I've put these instructions in the readme.


grilix:
I found a weird behaviour when seeing the storybook on the device. When dragging from the lower left corner, the app shows the create account screen. After closing and restarting the app, it starts normally, tho. I have no idea how this happens, really. :thinking: 

ezgif-3-031c16bd1545

graszka22:
@grilix probably you just pressed back button by accident (the navigator tab is very close to it). I've fixed the navigation so that you're brought back to settings menu.

grilix:
@graszka22 you're right, that fixed it! :clap:

PR #1676, opened 2020-02-10
PIL-278 BugFix/BTC Minor Bugs update

alazarevski:
Fixes balance shown in accounts list with bitcoinBalances in storage
Fix BTC wallet initiation with privateKey imports from the eth wallet (generate/send functions)


grilix:
@alazarevski wouldn't this result in a different address being generated? :thinking:

alazarevski:
Yes, the address will be different. Pillarwallet support mainly mnemonic as the standard is to use mnemonic. When the users import via privateKey only the privateKey is used and no mnemonic can be generated. Therefore they can derive the address just from the privateKey, and will not use mnemonic at all

grilix:
Oh, but this should only affect generating addresses from private keys, right? :+1:

alazarevski:
> Oh, but this should only affect generating addresses from private keys, right? 👍

Yes. Because they will only use the privateKey, the addresses will always be derived from that privateKey

PR #1680, opened 2020-02-11
PIL-271 Add interaction tests

graszka22:
note: based on https://github.com/pillarwallet/pillarwallet/pull/1675, merge that one first

I've given up on specifications addon (https://github.com/mthuret/storybook-addon-specifications), it isn't maintained anymore, it requires many hacks and I'm not even sure if it's supposed to work on React Native (nothing pops up in the browser or in the addons section).

Instead I've added just usual Enzyme test and I've made sure that it works. We can simply put such tests in each component's folder. The only difference between this approach and specifications addon is that tests are not run in the storybook but after `yarn test` and the tests are in separate files (which is even nicer imho).


storybook/stories/Button/Button.test.js
```diff
@@ -0,0 +1,34 @@
...
+    );
+    mount(component);
+  });
+  it('clicks', () => {
+    const onPressCallback = jest.fn();
+    const component = (
+      <Button onPress={onPressCallback} />
+    );
+    const button = mount(component);
+    // yeah, I know it doesn't make any sense but enzyme for RN doesn't support .simulate('press')
```
grilix: Not sure about this, but if we can't simulate the click, this is not really testing anything, since we're just calling the function we just passed to it via props :thinking:

graszka22: Yeah it's not really testing, that's why I added this comment. Those tests are just to check if testing is working, these components will be removed anyway after I add real components.

PR #1686, opened 2020-02-12
Remove the unneeded tag bump for develop branch

antonioned:
Bumping patch version in develop branch does absolutely nothing for us so removing it. It just creates a tag that is not needed and used in any way and just clutters the releases page in the repo.

Keep in mind team, that this will always bump the minor version in master by default. If you think we need to change to patch number as default, please comment.


antonioned:
added a fix for the latest failed archive of ipa file in prod_ios job

PR #1689, opened 2020-02-12
Feature/ens support (PIL-200)

jsidorenko:
This PR adds support for ENS in our wallet.

The main changes:
1) for legacy users we auto-set the ens name on the background when they login;
2) for new users we're creating an ens subdomain for them by passing their username into `smartWalletService.createAccount` method;
3) we're allowing users to enter the recipient's ens name instead of the address;
4) new EnsRegistry was added to cache the `address=>ens` mappings. Additionally, we're using this registry in address rendering


src/actions/ensRegistryActions.js
```diff
@@ -0,0 +1,95 @@
...
+export const extractEnsInfoFromTransactionsAction = (transactions: Object[]) => {
+  return async (dispatch: Dispatch, getState: GetState) => {
+    const { ensRegistry: { data: ensRegistry } } = getState();
+    let dataUpdated = false;
+
+    transactions.forEach(transaction => {
+      const fromEnsName = get(transaction, 'from.account.ensName') || get(transaction, 'sender.account.ensName');
+      const fromAddress = get(transaction, 'from.account.address') || get(transaction, 'sender.account.address');
+      const toEnsName = get(transaction, 'to.account.ensName') || get(transaction, 'recipient.account.ensName');
+      const toAddress = get(transaction, 'to.account.address') || get(transaction, 'recipient.account.address');
```
grilix: What about something like:

```js
const sender = transaction.from || transaction.sender;
const recipient = transaction.to || transaction.recipient;
const {
  ensName: fromEnsName,
  address: fromAddress,
} = sender.account; // I think this is safe, but we could do: get(sender, 'account', {});
const {
  ensName: toEnsName,
  address: toAddress,
} = recipient.account; // I think this is also safe, or we could do: get(recipient, 'account', {});
```

I'm assuming we're always gonna get everything under either `transaction.from.account` or `transaction.sender.account`. If we receive data in both, it won't work. For example:

```json
{
  "transaction": {
    "from": {
      "account": {
        "ensName": "name"
      }
    },
    "sender": {
      "account": {
        "address": "0x..."
      }
    }
}
```

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -120,12 +123,12 @@ const EmptyStateWrapper = styled.View`
   justify-content: center;
 `;
 
-type EmptyState = {
+type EmptyState = {|
```
grilix: :clap:

src/actions/ensRegistryActions.js
jsidorenko: Unfortunately, the `account` property is optional, since you can get a transaction from non-registered address. In that case, as I remember, `from.account` will be null, and you get the `from.address` property as a string (without the ens value).

We could change this block to that one:
```javascript
      const sender = transaction.from || transaction.sender;
      const recipient = transaction.to || transaction.recipient;

      const fromEnsName = get(sender, 'account.ensName');
      const fromAddress = get(sender, 'account.address');
      const toEnsName = get(recipient, 'account.ensName');
      const toAddress = get(recipient, 'account.address');
```
what do you think?

PR #1692, opened 2020-02-13
PIL-234 Add email verification steps

grilix:
Allow users to verify their email on the referrals flow. Email verification is required before referrals emails can be sent.


src/actions/userActions.js
```diff
@@ -17,14 +17,34 @@
...
 import { logEventAction } from 'actions/analyticsActions';
 import type { Dispatch, GetState } from 'reducers/rootReducer';
 import SDKWrapper from 'services/api';
 import { saveDbAction } from './dbActions';
 
+const sendingOneTimePasswordAction = () => ({
+  type: SENDING_OTP,
+});
+
+const otpPasswordSent = () => ({
```
graszka22: all actions have "Action" suffix in the name, right? `otpPasswordSentAction` would be a better name

src/components/Verification/EmailVerified.js
```diff
@@ -0,0 +1,44 @@
...
+  onPressBack: () => void,
+};
+
+const EmailVerified = (props: Props) => {
+  const { onPressBack } = props;
+
+  return (
+    <React.Fragment>
+      <MediumText>
+        You have confirmed your email!
```
graszka22: Please add space between button and text. Also the text could be a bit bigger (on all verification screens).

Screenshot 2020-02-19 at 11 15 28

src/components/Verification/NoEmail.js
```diff
@@ -0,0 +1,43 @@
...
+  onPressSetEmail: () => void,
+};
+
+const NoEmail = (props: Props) => {
+  const { onPressSetEmail } = props;
+
+  return (
+    <React.Fragment>
+      <MediumText>
+        You don&apos;t have an email, please set your email first.
```
graszka22: Add space here too

src/components/Verification/SendEmailCode.js
```diff
@@ -0,0 +1,51 @@
...
+
+// components
+import Button from 'components/Button';
+import { MediumText } from 'components/Typography';
+
+// utils
+import { spacing } from 'utils/variables';
+
+const Label = styled(MediumText)`
+  padding-bottom: ${spacing.layoutSides}px;
```
graszka22: `layoutSides` is for layout sides not bottom margin, use `spacing.mediumLarge`

src/screens/Users/VerifyEmail.js
```diff
@@ -0,0 +1,182 @@
...
+  };
+
+  updateCode = (code: string) => {
+    this.setState({ code });
+  };
+
+  setEmail = () => {
+    const { navigation } = this.props;
+
+    navigation.navigate(ADD_EDIT_USER);
```
graszka22: It's a bit weird:
- I press the go to profile button
- I'm brought to user settings
- I go back
- and I'm on the main user settings page instead of email verification

src/actions/userActions.js
grilix: Good catch, updated.

src/components/Verification/EmailVerified.js
grilix: Updated. Created a `Label` subcomponent so we have all of them look the same and we can easily update it. Not sure if there's already a component that we could have used for this (The `Typography/Label` doesn't look like a fit, let me know what you think.

src/screens/Users/VerifyEmail.js
grilix: You're right, fixed.

src/components/Verification/EmailVerified.js
graszka22: It's a shame that this branch is not updated with develop, because on develop we have those new Typography components and you'd just do something like this:
`<MediumText big>...</MediumText>`. And creating a new separate component for some text on one screen is a bit overkill. And the name is too generic as well. 
Here is what'd do to: looking at the designs, such "descriptions" are usually regular font with size 16sp (see for example the recovery flow, there are a lot of similar screens with description and a button). And in the design system such font is marked as just "Text". So I'd create a new component in the Typography file called "Text" or "RegularText" (so it follows the design system) and use that. How does it sound?

src/actions/userActions.js
```diff
@@ -17,14 +17,34 @@
...
 import { logEventAction } from 'actions/analyticsActions';
 import type { Dispatch, GetState } from 'reducers/rootReducer';
 import SDKWrapper from 'services/api';
 import { saveDbAction } from './dbActions';
 
+const sendingOneTimePasswordAction = () => ({
+  type: SENDING_OTP,
+});
+
+const otpPasswordSentAction = () => ({
```
poocart: As other actions have relation to oneTimePassword and only this has "otp" short handle:
```suggestion
const oneTimePasswordSentAction = () => ({
```

src/actions/userActions.js
```diff
@@ -82,6 +109,43 @@ export type VerificationPhoneAction = {
...
+      dispatch({
+        type: ADD_NOTIFICATION,
+        payload: {
+          message: 'Email verification was successful',
+          title: 'Validation successful',
+          messageType: 'success',
+        },
+      });
+    } else {
+      dispatch(otpPasswordSentAction());
```
poocart: Can you explain the logic here? Why we mark it as sent only when error occurs?

src/components/Verification/ConfirmCode.js
```diff
@@ -0,0 +1,68 @@
...
+import styled from 'styled-components/native';
+
+// components
+import Button from 'components/Button';
+import TextInput from 'components/TextInput';
+import Label from 'components/Verification/Label';
+
+// utils
+import { spacing } from 'utils/variables';
+
```
poocart:
```suggestion


```

src/actions/userActions.js
grilix: This is so we go back to the state we had before calling `dispatch(sendingOneTimePasswordAction());`. The state chain, on failure scenario would be: `sent` -> `sending` -> `sent`.

grilix:
Thanks all for the comments! The UI for this feature changed a lot and it doesn't depend on the referrals anymore. So we extracted the code out and updated on #1701, please continue the thread there :heart:

PR #1696, opened 2020-02-19
PIL-243 Feature/hot swaps

graszka22:


grilix:
Found a weird behaviour on this: When selecting swaps, for some of them (most of them, actually) it seems like the "Sell" option get broken and cannot change the token afterwards (see how the arrows disappear from the dropdown).

hot-swaps

vyvil:
> Found a weird behaviour on this: When selecting swaps, for some of them (most of them, actually) it seems like the "Sell" option get broken and cannot change the token afterwards (see how the arrows disappear from the dropdown).
> 
> hot-swaps

Angelika pointed this out last week (or even earlier). Yet it's not a bug, but a feature. Our exchange options are being filtered based on supported tokens. 
I've brought this into Dmitry's attention and there was a solution to disable this filtering. Yet, since empty state will now be different than expected, I'm not sure if that's such a valid option 🤔.
Maybe we need more discussions around that. But either way, I'd treat it as out of scope for this particular PR.

src/screens/Exchange/HotSwapsList.js
```diff
@@ -0,0 +1,178 @@
...
+  render() {
+    const { swaps } = this.props;
+
+    // on small devices 3 columns will cause text lines to be broken
+    // on larger devices 2 columns will cause pills to be too wide
+    const columns = screenWidth >= threeColumnsMinScreenWidth ? 3 : 2;
+    const data = [...swaps];
+    while (data.length % columns !== 0) {
+      data.push({});
+    }
```
poocart: What about moving this to other lifecycle step rather than render? I. e. `componentDidUpdate` with check if data has changed?

graszka22: But why? It won't have much impact on performance imho.

poocart: The problem is that it will iterate the while on each render while it's only needed for first time, right?

jsidorenko: Actually, it's a very quick operation, the cycle will iterate +-2 times only.
If render function itself will be called too many time, then it's gonna be another problem :)

PR #1708, opened 2020-02-28
PIL-290 SW activation floating cards

graszka22:
I added a new component to the storybook by the way :)


src/screens/Assets/WalletView.js
```diff
@@ -436,20 +435,30 @@ class WalletView extends React.Component<Props, State> {
...
           title={insightsTitle}
           insightChecklist={insightList}
           onClose={() => { hideInsight(); }}
           wrapperStyle={{ borderBottomWidth: 1, borderBottomColor: colors.border }}
         />
-        {blockAssetsView &&
-        <DeploymentView
-          message={deploymentData.error ? getDeployErrorMessage(deploymentData.error) : sendingBlockedMessage}
-          buttonAction={deploymentData.error ? () => deploySmartWallet() : null}
-          buttonLabel="Retry"
```
jsidorenko: That component was helping users whose deployment failed. What should they do now?

src/screens/Assets/WalletView.js
```diff
@@ -494,12 +503,6 @@ class WalletView extends React.Component<Props, State> {
             />)}
           {!isInSearchMode && (!balance || !!showFinishSmartWalletActivation) &&
           <ActionsWrapper>
-            {!!showFinishSmartWalletActivation && smartWalletFeatureEnabled &&
-            <ListItemChevron
-              label={smartWalletDeployLabel}
-              onPress={() => navigation.navigate(SMART_WALLET_INTRO, { deploy: showDeploySmartWallet })}
```
jsidorenko: That was covering the case for legacy users when they created the sw account, but haven't deployed it yet

src/components/InsightWithButton/InsightWithButton.js
```diff
@@ -0,0 +1,104 @@
...
+    </ItemContainer>
+  );
+};
+
+const InsightWithButton = ({
+  title, itemsList, buttonTitle, description, buttonProps, onButtonPress,
+}: Props) => {
+  return (
+    <InsightWrapper>
+      <ShadowedCard>
```
vyvil: This component needs to be updated as well, since InsightCards should have bigger border radius - 30px;

jsidorenko:
It looks weird once you click on the 'Activate the SW' and choose that you have ETH, but it shows there are no assets to transfer, while I have ETH in my key based wallet.
I think the issue is that you're checking/displaying assets from the current account and not from the key based wallet

jsidorenko:
image
For the activation flow the submit button should be renamed to Activate Smart Wallet

src/screens/Assets/WalletView.js
graszka22: Do you mean the case like this:
- on the onboarding page, select "create account"
- you'll have the sw account created but not deployed yet
- you'll just see the insights "Why Smart Wallet knocks out your old private key wallet?" and then "Activate Smart Wallet" and you can deploy wallet from there which I think is correct

src/actions/smartWalletActions.js
```diff
@@ -364,6 +365,13 @@ export const dismissSmartWalletUpgradeAction = () => {
   };
 };
 
+export const dismissPPNInsightAction = () => {
+  return async (dispatch: Dispatch) => {
+    dispatch(saveDbAction('app_settings', { appSettings: { PPNInsightDismissed: true } }));
```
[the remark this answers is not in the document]
graszka22: I can't see any more insights like this in the designs that you can dismiss and these two are both related to smart wallet. It's hardly possible that there will be another one like this, but in that case we can refactor it easily.

graszka22:
@vyvil @jsidorenko 
I removed the DeploymentView completely and now those SWActivationCards can also show that wallet is being deployed and errors. I asked Dmitry about this but he hasn't answered me yet.

> It looks weird once you click on the 'Activate the SW' and choose that you have ETH, but it shows there are no assets to transfer, while I have ETH in my key based wallet.
I think the issue is that you're checking/displaying assets from the current account and not from the key based wallet

Now that's a good catch. There is kind of a flaw in the logic now.
As I understand there are two types of users:
- legacy users that has no SW account at all
- new users that has SW account created on onboarding, they just have to deploy it

And up till now it was implemented like this:
- legacy users enter the SW flow, select assets to transfer and the SW account is created and then SW is deployed
- new users have SW by default, they just deploy the SW, no need to transfer anything

Now, it changes the flow for legacy users like this:
- they enter the SW
- the account is created for them, but not deployed yet
- at this point they're just like new users and we cannot differentiate between them (!)
- now:
   - it doesn't make sense to transfer from SW to SW
   - we could try switching to legacy wallet and select to transfer but for new users this doesn't make sense
   - we can just treat both types equally and don't show them transfer assets screen and it's easy and  works fine, the only downside is that the legacy user would need to transfer assets by himself and that's what I'm doing now temporarily
  - we can try transferring assets while creating the account for legacy user but doesn't that mean that we need to deploy the SW at that time?
   - we can save the flag somewhere that the user is legacy user and check for that and show transfer assets screen if necessary
I also asked Dmitry what he thinks about this and I'm waiting for response

> Just to double check - PPN Insight has to be visible once for user (unless he/she hasn't dismissed it by pressing "Enable Pillar Network").
That would mean, that if user reimports wallet and had previously dismissed it, it should no longer be visible.
I believe this case is not yet covered.

Correct me if I'm wrong but if we reimport the wallet then all the data in the app gets wiped out. So how can we store information that user has dismissed the insight? Without backend?

src/actions/smartWalletActions.js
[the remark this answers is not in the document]
vyvil: In current designs and for this feature - yeah.
But I've checked with Dmitry, that we might have more of such cards.
I'd recommend discussing it a bit more with him.

And refactoring part wouldn't be just couple of lines - we would need migrations of some sort to keep previously set "true" values.

src/components/ShadowedCard/ShadowedCard.js
```diff
@@ -158,6 +160,7 @@ class ShadowedCard extends React.Component<Props, State> {
             style={upperContentWrapperStyle}
             isAnimated={isAnimated}
             opacity={contentOpacity}
+            borderRadius={borderRadius}
```
vyvil: Can we fallback to borderRadius of 6px here rather than in styled component declaration or change property name? It now overwrites border radius and sets it to 0 if borderRadius is not passed (check settings cards, for example).

src/components/ShadowedCard/ShadowedCard.js
```diff
@@ -120,6 +121,7 @@ class ShadowedCard extends React.Component<Props, State> {
       upperContentWrapperStyle = {},
       theme,
       isAnimated,
+      borderRadius,
```
vyvil: Since borderRadius can now be passed as prop, it should be also passed down to Shadow component ([L156](url)) to adjust Shadow's radius as well (not just the card's).
And it should be slightly smaller (by 2px) than the borderRadius that has been passed in.

src/actions/insightsActions.js
```diff
@@ -0,0 +1,41 @@
...
+import {
+  DISMISS_SMART_WALLET_INSIGHT,
+  DISMISS_PPN_INSIGHT,
+} from 'constants/insightsConstants';
+import { saveDbAction } from 'actions/dbActions';
+import type { Dispatch } from 'reducers/rootReducer';
+
+export const dismissSmartWalletInsightAction = () => {
+  return async (dispatch: Dispatch) => {
+    dispatch(saveDbAction('app_settings', { appSettings: { smartWalletInsightDismissed: true } }));
```
vyvil: Hm, insight are still being stored in app_settings here.

src/navigation/appNavigation.js
```diff
@@ -497,7 +495,6 @@ const sendBitcoinFromAssetFlow = createStackNavigator(
 
 // SEND ASSETS FROM CONTACT FLOW
 const sendTokenFromContactFlow = createStackNavigator({
-  [SEND_TOKEN_ASSETS]: SendTokenAssetsScreen,
```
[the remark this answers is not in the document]
graszka22: ok, I've just searched for `SEND_TOKEN_ASSETS`, haven't seen it used anywhere, so I've deduced it's safe to remove
I haven't noticed that we're navigating to `SEND_TOKEN_FROM_CONTACT_FLOW` so to the SendTokenAssets as it's the first screen in the navigator
I'll bring it back

src/components/SWActivationCard/SWActivationCard.js
```diff
@@ -0,0 +1,159 @@
...
+    const {
+      navigation, baseFiatCurrency, balances, rates, deploySmartWallet,
+    } = this.props;
+    const fiatCurrency = baseFiatCurrency || defaultFiatCurrency;
+    const ethBalance = getBalance(balances, ETH);
+    const balanceInFiat = ethBalance * getRate(rates, ETH, fiatCurrency);
+    const fiatAmount = formatFiat(balanceInFiat, baseFiatCurrency || defaultFiatCurrency);
+
+    return [
+      {
```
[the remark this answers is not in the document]
graszka22: this modal will be changed anyway

src/screens/Assets/WalletView.js
jsidorenko: You need to a) transfer eth to cover the deployment; b) deploy the account.
Since it's a 2 step operation - it could fail in between, that's why we've added the button to retry the deployment.
I think it could be possible if the user transfers too little eth, just to cover the gas and then the gas price spikes

graszka22:
@vy @jsidorenko 
I've added the new modal (https://app.zeplin.io/project/5db9b5d14c30ae56faaa3030/screen/5e687d908e2ebb14275d72ef).

- the icon for dark mode is missing, I'll add it when I get the asset
- there is a whole flow with choosing assets to transfer to SW and confirming that which now seems deprecated (we can't get to it). I've asked Dmitry about it and he said it's fine. But there is a possibility the flow will change once again therefore I've not removed the flow from the code. Let me know if you think it's better to remove it.
- I've added a new little cute `Spacing` component somewhere in the `Layout.js`. The reason behind it is that it's tedious to create new components with style just to add some spacing between items. Especially consider buttons or text components: now we have components like `BaseText` or `MediumText` and we can make formatted text in one line but nooo we must create a new component anyway just to create 20px spacing between two lines of text.   In my opinion this leads to shorter and more readable code. Also a couple of simple Views shouldn't affect the performance. Let me know what do you think about it.

src/components/SWActivationCard/SWActivationCard.js
```diff
@@ -0,0 +1,406 @@
...
+            <CachedImage
+              style={{ width: 64, height: 64, alignSelf: 'center' }}
+              source={themeType === DARK_THEME ? smartWalletIconDark : smartWalletIcon}
+            />
+            <Spacing h={20} />
+            <BaseText medium>
+              Enable better security and free instant transactions via Pillar Network. You will get a new badge too.
+            </BaseText>
+            <Spacing h={15} />
+            {!!ethBalanceInSmartWallet && (
```
jsidorenko: What if I just created my smart wallet and don't have any eth here, but have in my key based wallet?
We should display that case, disable activation from the smart wallet (since it's empty) and preselect the key based option

src/actions/assetsActions.js
```diff
@@ -517,21 +517,28 @@ function notifyAboutIncreasedBalance(newBalances: Balance[], oldBalances: Balanc
...
-      featureFlags: { data: { SMART_WALLET_ENABLED: smartWalletFeatureEnabled } },
-    } = getState();
+export const updateAccountBalancesAction = (accountId: string, balances: Balances) => {
+  return async (dispatch: Dispatch, getState: GetState) => {
+    const allBalances = getState().balances.data;
+    const updatedBalances = {
+      ...allBalances,
+      [accountId]: balances,
+    };
+    await dispatch(saveDbAction('balances', { balances: updatedBalances }, true));
```
poocart: Why wait here for DB to write if we're updating reducer state as well?
```suggestion
    dispatch(saveDbAction('balances', { balances: updatedBalances }, true));
```

PR #1713, opened 2020-03-02
PIL-292 Feature/update sw balance check ui

graszka22:
Based on https://github.com/pillarwallet/pillarwallet/pull/1708


src/components/ActionModal/ActionModal.js
```diff
@@ -0,0 +1,101 @@
...
+    } = this.props;
+    const colors = getThemeColors(theme);
+
+    return (
+      <SlideModal
+        isVisible={isVisible}
+        noClose
+        background={colors.card}
+        hideHeader
+        onModalHide={() => onModalClose(() => {})}
```
jsidorenko:
```suggestion
        onModalHide={() => onModalClose(noop)}
```

jsidorenko: From `src/utils/common.js`

src/components/ActionModal/ActionModal.js
```diff
@@ -0,0 +1,101 @@
...
+const ItemContainer = styled.TouchableOpacity`
+  flex-direction: row;
+  padding: 15px 0;
+  align-items: center;
+  justify-content: space-between;
+`;
+
+const ChevronIcon = styled(Icon)`
+  color: ${themedColors.secondaryText};
+  ${fontStyles.small}
```
jsidorenko:
```suggestion
  ${fontStyles.small};
```

src/components/ActionModal/ActionModal.js
```diff
@@ -0,0 +1,101 @@
...
+
+const MainContainer = styled.View`
+  padding: 15px ${spacing.layoutSides}px 40px;
+`;
+
+const ItemContainer = styled.TouchableOpacity`
+  flex-direction: row;
+  padding: 15px 0;
+  align-items: center;
+  justify-content: space-between;
```
vyvil: Based on designs, those items should be more spaced out. 
Could you please add appropriate amount of vertical padding / margin?
That should also help with current spacing on the top of modal - based on designs it should also be bigger - 30px, not 15px;

src/components/SWActivationCard/SWActivationCard.js
```diff
@@ -36,6 +50,38 @@ class SWActivationCard extends React.Component<Props, State> {
...
+        },
+        key: 'has ETH',
+      },
+      {
+        label: "I'd like to have some",
+        chevron: true,
+        onPress: () => {
+          navigation.navigate(EXCHANGE, {
+            fromAssetCode: fiatCurrency,
+            toAssetCode: 'ETH',
```
vyvil: There's ETH as a constant in assetsConstants (and it is already imported in this doc), it would be better to use it here.

src/components/ActionModal/ActionModal.js
vyvil: Or it could easily be optional (`onModalClose: (?() => void) => void`)
Since `this.setState({ isModalVisible: false }, undefined);` wouldn't break anything ([here](https://github.com/pillarwallet/pillarwallet/pull/1713/files#diff-ccb92037b12cf63dd541157c3d90ca7cR98))

src/components/ActionModal/ActionModal.js
vyvil: And it looks more like tiny (10px) rather than small to me :)

src/components/ActionModal/ActionModal.js
```diff
@@ -0,0 +1,101 @@
...
+  color: ${themedColors.secondaryText};
+  ${fontStyles.small}
+`;
+
+const Item = ({
+  label, onPress, money, chevron,
+}) => (
+  <ItemContainer onPress={onPress}>
+    <MediumText big>{label}</MediumText>
+    {money && <BaseText medium secondary>{money}</BaseText>}
```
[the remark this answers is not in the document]
jsidorenko:
```suggestion
    {!!money && <BaseText medium secondary>{money}</BaseText>}
```
if money would be equal to 0 it will crash on android :(

PR #1714, opened 2020-03-03
Feature/erc20 support note on receive modal [PIL-268 and PIL-282]

vyvil:
This PR:
1. Adds ERC20 support note on receive modal (PIL-268)
I've reused HeaderBlock on SlideModals in order to allow passing in custom components to SlideHeader. 
While at it, I've removed user component form HeaderBlock, since it is no longer used on almost each screen as it was initially planned. I've passed it as a custom prop on Home screen.
Also, I've slightly altered ConnectionConfirmationModal since it was the only modal that used subtitle. I've changed it into paragraph.

2. Increases back button's clickable area (PIL-282).

### DIFF
--- a/src/components/HeaderBlock/HeaderBlock.js
+++ b/src/components/HeaderBlock/HeaderBlock.js
@@ -224,7 +224,7 @@ class HeaderBlock extends React.Component<Props> {
                 <BackIcon
                   icon="back"
                   color={transparent ? colors.control : colors.text}
-                  onPress={customOnBack ? () => customOnBack() : () => { navigation.goBack(null); }}
+                  onPress={customOnBack ? customOnBack : () => navigation.goBack(null)}
                   fontSize={fontSizes.large}
                   horizontalAlign="flex-start"
                 />
--- a/src/components/HeaderBlock/HeaderBlock.js
+++ b/src/components/HeaderBlock/HeaderBlock.js
@@ -224,7 +224,7 @@ class HeaderBlock extends React.Component<Props> {
                 <BackIcon
                   icon="back"
                   color={transparent ? colors.control : colors.text}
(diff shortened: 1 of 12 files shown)


src/components/HeaderBlock/HeaderBlock.js
```diff
@@ -209,22 +201,35 @@ class HeaderBlock extends React.Component<Props> {
...
+                style={{
+                  marginTop: -20,
+                  marginBottom: -20,
+                }}
+                key="back"
+              >
+                <BackIcon
+                  icon="back"
+                  color={transparent ? colors.control : colors.text}
+                  onPress={customOnBack ? () => customOnBack() : () => { navigation.goBack(null); }}
```
jsidorenko:
```suggestion
                  onPress={customOnBack ? customOnBack : () => navigation.goBack(null)}
```

jsidorenko: or 
```suggestion
                  onPress={customOnBack || () => navigation.goBack(null)}
```
not sure if we support `??` it would be really cool to use it here

src/screens/Asset/ReceiveModal.js
```diff
@@ -119,34 +132,38 @@ export default class ReceiveModal extends React.Component<Props, *> {
...
             />
-            <IconsContainer>
-              <Image source={visaIcon} />
-              <IconsSpacing />
-              <Image source={mastercardIcon} />
-            </IconsContainer>
-          </BuyTokensWrapper>
-        )}
+          </ButtonsRow>
+          {showBuyTokensButton &&
```
jsidorenko:
```suggestion
          {showBuyTokensButton && (
```
to keep consistency with the changes above :)

PR #1715, opened 2020-03-03
PIL-301 Feature/include walletId in exchange requests

jdnmr:


src/actions/exchangeActions.js
```diff
@@ -367,10 +369,16 @@ export const setTokenAllowanceAction = (
 ) => {
   return async (dispatch: Dispatch, getState: GetState) => {
     connectExchangeService(getState());
+    const {
```
jsidorenko: It's better to grab the state after `await` since you can get different results if some state info will be changed while you're waiting for the response

jdnmr: Unless I called `getState` twice (which would be awkward), I can't do that, because I need `walletId` before I call `exchangeService.setTokenAllowance`

jsidorenko: the `walletId` value is kind of static, while the `supportedAssets` or `accounts` could get changed. I would call for the walletId here, and for other elements bellow.
We had a similar issue with transactions history before, and it was hard to detect why suddenly a component was rendering the wrong data

PR #1716, opened 2020-03-03
Fix ERC721 error handling

jsidorenko:
Fixes the crash which happens when ERC721 transaction fails. [Sentry report](https://sentry.io/organizations/pillar-project-worldwide-ltd/issues/1542702646/?environment=production&project=1294444&referrer=alert_email)

### DIFF
--- a/src/actions/assetsActions.js
+++ b/src/actions/assetsActions.js
@@ -332,6 +332,7 @@ export const sendAssetAction = (
           getState(),
         );
 
+        // $FlowFixMe
         if (tokenTx.hash) {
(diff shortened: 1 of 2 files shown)


src/actions/assetsActions.js
```diff
@@ -332,6 +332,7 @@ export const sendAssetAction = (
           getState(),
         );
 
+        // $FlowFixMe
```
poocart: 👹

jsidorenko: 🤷‍♂

PR #1729, opened 2020-03-06
PIL 211-225, 295: Expo Cleanup

poocart:
This includes:
- Fresh native projects for iOS and Android based on React Native 0.59.10.
- Updated all JS project files based on React Native 0.59.10, this includes such changes as ".babelrc" to "babel.config.js" and etc. This is because upgrading to newer React Native using React Native upgrade tools will be much easier when we keep file structure in same manner as original project source.
- Removal of Fabric (you can still find relations, but it's how they kept libs even though it's not connecting to Fabric anymore).
- Updated `react-native-firebase` to v6., From this version Push Notifications are excluded so added new `react-push-notifications` library, This comes without app badge methods for Android so added `react-native-android-badge` as well. Down part is that initial notification or "on push open" events are not working on Android (library's issue) and we try to update to latest React Native and see if it helps as it doesn't work because of native lifecycles not being triggered in module's native part. Upgrade to latest React Native is next step from this.
- Removed "reverse" option on `jetifier` as it's main purpose is to update all Android libs to AndroidX which helps to support more devices and edge cases that happen on Android, basically more modern ex Android Support lib.
- Updated Android "build flavours" as those were useless (changes can be seen that for builds we don't have "prod" or "dev" appendix, only "debug", "release" or "staging".
- Updated major libs: `react-native-camera`, `lottie-react-native`, `lottie-ios`, `@sentry/react-native`, `react-native-firebase`, `react-native-device-info`, `react-native-image-crop-picker`, `rn-threads`, `native-base`.
- Removed major libs: `flow-typed`, `reactotron`, `husky`.
- Removed `console.log` and switched to single log method (there's still `console.warn` and `console.error` in the code, those can be adjusted further on). 
- Switched Sentry to single method report.
- Many other cleanups and improvements.


App.js
```diff
@@ -161,6 +161,7 @@ class App extends React.Component<Props, *> {
     const { url: deepLink } = event;
     if (deepLink === undefined) return;
     executeDeepLink(deepLink);
+    executeDeepLink(deepLink);
```
jdnmr: I guess this doesn't need to be called twice?

ios/pillarwallet/Info.plist
```diff
@@ -114,8 +80,17 @@
...
+		<string>Roboto.ttf</string>
+		<string>rubicon-icon-font.ttf</string>
 		<string>SimpleLineIcons.ttf</string>
 		<string>Zocial.ttf</string>
+		<string>Fontisto.ttf</string>
+		<string>EuclidCircularB-Bold.otf</string>
+		<string>EuclidCircularB-Light.otf</string>
+		<string>EuclidCircularB-Medium.otf</string>
+		<string>EuclidCircularB-Regular.otf</string>
+		<string>PillarIcons.ttf</string>
```
jdnmr: While trying to build in xCode, these font files throw plenty of Multiple Command errors. This only occurs in new build system, works fine in legacy. Not sure if we need to care about this.

patches/@storybook+addons+5.3.12.patch
```diff
@@ -1,13 +0,0 @@
...
-+++ b/node_modules/@storybook/addons/dist/hooks.js
-@@ -264,7 +264,7 @@ var applyHooks = function applyHooks(applyDecorators) {
-       var hooks = context.hooks;
-       hooks.prevMountedDecorators = hooks.mountedDecorators;
-       hooks.mountedDecorators = new Set([getStory].concat(_toConsumableArray(decorators)));
--      hooks.currentContext = context;
-+      hooks.currentContext = Object.assign({}, context, { hooks: null });
-       hooks.hasUpdates = false;
-       var result = decorated(context);
-       numberOfRenders = 1;
```
jdnmr: Are you sure we don't need this patch? It was added together with storybook (#1675 )

poocart: Storybook seems to be working without it. Should be fine, tested last week with @graszka22.

ios/pillarwallet/Info.plist
poocart: Interesting, which XCode you're in? We can commit the legacy build setting to make it easier. Does it stop during build process or just throws error, but runs the app successfully?

jdnmr: 11.2.1, it stops the build. This might be something wrong with my local setup, so it would be great if someone else tried this too

alazarevski: Please check https://github.com/pillarwallet/pillarwallet/compare/fix/pure-react-native...fix/pure-rn-ios-further-work
More specifically included changes in commits 7b6a55f 780abd1

src/utils/common.js
```diff
@@ -46,12 +46,36 @@ import {
...
 
 const WWW_URL_PATTERN = /^www\./i;
 const supportedAddressPrefixes = new RegExp(
   `^(?:${ETHEREUM_ADDRESS_PREFIX}|${BITCOIN_ADDRESS_PREFIX}):`, 'gi',
 );
 
+export const printLog = (...params: any) => {
+  if ((isProdEnv && !__DEV__) || isTest) return;
+  console.log(...params);
+};
```
grilix: Just curious of what would be the benefit of having this. Is that bad to call `console.log` on other environment than dev or test?

poocart: Exactly, it is known that console.log produces huge lags in real devices.

https://reactnative.dev/docs/performance.html#using-consolelog-statements

android/app/src/main/java/com/pillarproject/wallet/MainApplication.java
```diff
@@ -1,115 +1,115 @@
...
-        new IntercomPackage(),
-        new LinearGradientPackage(),
-        new RNFirebaseNotificationsPackage(),
-        new RandomBytesPackage(),
-        new PickerPackage(),
-        new NativeShadowPackage(),
-        new RNThreadPackage(mReactNativeHost),
-        new KeychainPackage()
+          new MainReactPackage(),
+            new BadgePackage(),
```
vyvil: Indentation :)

src/utils/common.js
grilix: Wouldn't it be better to just remove them by using the suggested plugin?

.gitignore
```diff
@@ -1,30 +1,9 @@
-# See https://help.github.com/ignore-files/ for more about ignoring files.
-
-# dependencies
-/node_modules
-
-# misc
-.env.local
```
graszka22: there are some .env variables that we don't want to commit by accident and we're keeping them in `.env.development` files
we were talking about this recently: https://pillarwallet.slack.com/archives/CEEV48867/p1582121613105600

package.json
```diff
@@ -2,138 +2,46 @@
...
-    "metro-react-native-babel-preset": "^0.51.1",
-    "mock-socket": "^9.0.3",
-    "npm-run-all": "^4.1.5",
-    "prop-types": "^15.6.1",
-    "react-dom": "^16.3.0-alpha.1",
-    "react-native-debugger-open": "^0.3.19",
-    "react-native-dotenv": "^0.2.0",
-    "react-native-storybook-loader": "^1.8.1",
-    "react-test-renderer": "16.6.3",
-    "reactotron-react-native": "^3.2.2",
```
graszka22: out of curiosity, now that Reactotron is removed what do we use to inspect the Redux actions and state?

src/utils/common.js
poocart: Well, tbh both results the same, but we can add it. However with this approach it will just remove one line that we have now.

package.json
poocart: Redux dev tools are added as enhancer which allows us to inspect it. Reactotron was never used.

.gitignore
poocart: Added `.env.*`, what do you think?

.gitignore
```diff
@@ -1,30 +1,9 @@
...
 .DS_Store
 
-npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
-
-.idea
-
-# output of jest tests+ coverage etc.
-/reports/
```
graszka22: why removing that?

poocart: Restored.

PR #1731, opened 2020-03-06
Fix/avoid threads debugging crash android

alazarevski:
Avoid running threads if on emulator debug since non refactored code in the threads module crashes the application when the threads are invoked


poocart:
Not sure if we need it as it's not crashing already.

jsidorenko:
@alazarevski I think we can close this since there will be no rn-threads support in the upcoming version

PR #1734, opened 2020-03-09
[PIL-309] Fix/Disappearing assets in smart account

alazarevski:
Avoid removal of existing assets if smart account already has them linked.


jsidorenko:
`setSmartWalletUpgradeStatusAction` is called when we finish the SW migration as far as I remember. But the assets' problem appears on every user login

poocart:
In addition to what Jegor replied – why we're not using existing methods such as `accountAssetsSelector` for selecting account asstets? Why we can't use lodash's is empty for checking assets?

alazarevski:
@jsidorenko that's true, but following the call stack while debugging on the simulator, the current code executes on every login : 
src/actions/accountsActions.js 
function : `setActiveAccountAction`

`    if (state === sdkConstants.AccountStates.Deployed) {
      dispatch(setSmartWalletUpgradeStatusAction(SMART_WALLET_UPGRADE_STATUSES.DEPLOYMENT_COMPLETE));
      return;
    }`
Without the condition not to replace initial assets, the assets will be set to the default (ETH, PLR)

PR #1747, opened 2020-03-13
[PIL-371] Fix/add ens domain placeholder

alazarevski:
Add rightPlaceholder feature for textInput like '.pillar.eth'
Show ens for new pillarwallet, do not show for externally imported one

### DIFF
--- a/src/screens/NewProfile/NewProfile.js
+++ b/src/screens/NewProfile/NewProfile.js
@@ -119,6 +119,7 @@ const getDefaultFormOptions = (inputDisabled: boolean, isLoading?: boolean) => (
         statusIcon: null,
         statusIconColor: null,
         inputType: 'bigText',
+        rightPlaceholder: '.pillar.eth',
       },
     },
   },
--- a/src/screens/NewProfile/NewProfile.js
+++ b/src/screens/NewProfile/NewProfile.js
@@ -102,7 +102,7 @@ const formStructure = t.struct({
 
 const PROFILE_IMAGE_WIDTH = 144;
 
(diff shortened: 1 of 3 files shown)


src/screens/NewProfile/NewProfile.js
```diff
@@ -156,12 +157,13 @@ class NewProfile extends React.Component<Props, State> {
 
   constructor(props: Props) {
     super(props);
-    const { apiUser } = props;
+    const { apiUser, importedWallet } = props;
     const value = apiUser && apiUser.username ? { username: apiUser.username } : null;
     const inputDisabled = !!(apiUser && apiUser.id);
+    const showRightPlaceholder = !importedWallet;
```
jsidorenko: If I imported my old mnemonic that has no user associated on the Pillar platform, then it will show my the 'Pick a username' field without placeholder.

alazarevski:
There was a discussion about this whether to show it to new pillarwallet users only who will get a smart account and ens upon registration. We can discuss this further on Monday

src/screens/NewProfile/NewProfile.js
alazarevski: There was a discussion about this whether to show it to new pillarwallet users only who will get a smart account and ens upon registration. We can discuss this further on Monday

alazarevski: Just confirmed that imported wallets from external source have no smartwallet feature yet (they need to init & deploy)

jsidorenko: Right! Then everything is correct :)

PR #1749, opened 2020-03-15
PIL-307, PIL-308: Upgrade to latest React Native

poocart:
Inlcudes React Native version update to latest stable version of 0.61.5.

Notes:
- React Native auto linking was added.
- React Native fast refresh was added and in favor `redux` and `react-redux` had to be updated to latest.
- React Native `hermes` added for Android.
- Prettier lint is added, but ignored as we will have to go through all of our code and this is not part of update, but we can apply this later.
- Signal module for iOS is disabled for auto linking and added manually.
- Before running iOS make sure you have updated your local XCode and pod cli to latest version.
- Other libraries were updated to support auto linking.
- Partially removed connections key generation and `rn-threads` lib.
- Added `proxy-polyfill` as `hermes` JS engine for Android does not support `Proxy` calls.
- Updated `react-native-vector-icons` and applies some changes as icons were off the design with latest RN version.


.circleci/config.yml
```diff
@@ -382,7 +382,7 @@ jobs:
           name: Fetch CocoaPods Specs
           command: |
             Y | sudo gem uninstall cocoapods
-            sudo gem install cocoapods -v 1.5.3
+            sudo gem install cocoapods -v 1.9.1
```
antonioned: has this line been tested in our CircleCI pipeline?

graszka22:
After building the app and entering the home screen on iOS the following error appears:

Screenshot 2020-03-16 at 12 25 26

I've removed the `node_modules` completely, run `yarn`, started RN bundler, updated pod cli, run `pod install` and build the app in the xcode.

graszka22:
On Android:
- the fingerprint login is a bit broken:
Screenshot 2020-03-16 at 12 58 34

- the back arrow in the navbar is too high
Screenshot 2020-03-16 at 13 04 14

- the services tab is a bit broken:
Screenshot 2020-03-16 at 12 59 30

.circleci/config.yml
poocart: @antonioned will try to hit it once merged to this target branch of pure RN, but exactly same pod and XCode versions were used locally as defined in CircleCI config

poocart:
Updated per comments above. Biometrics request is due partial connections removal.

PR #1751, opened 2020-03-17
Fix/modal scroll (proposition)

vyvil:
This PR is an attempt to fix scroll issue on country select modal.
As of my understanding, scroll freezes due to multiple pan responders:
- SlideModal (modal in it) has a pan responder that captures scrolling to bottom (in order to dismiss modal)
- SelectList (FlatList in it) has a pan responder to scroll list
- SettingsItem (SelectList item) (TouchableOpacity / TouchableNativeFeedback in it) has a pan responder to capture touches (to execute onPress)

I've added couple of methods to manage modal's dismiss action when user interacts with SelectList (while scrolling it). Yet having touchable items in it, messes with Modal's pan responders. To disable capturing onMove, I've added custom touchable (view with custom pan responder).

This solution still has a latency when trying to close modal when scrolling down. 
So I've added another PR (#1752) that disables this scroll to dismiss feature in order to have smooth scrolling experience on modal's content.

Let me know what you think. Any optimisation ideas are more than welcome :)

### DIFF
--- a/src/components/Input/InputWithSwitch.js
+++ b/src/components/Input/InputWithSwitch.js
@@ -82,7 +82,7 @@ const StyledItemView = styled.View`
   padding: 0 ${spacing.layoutSides}px 0;
   border-bottom-color: ${({ hasErrors }) => hasErrors ? themedColors.negative : themedColors.tertiary};
   border-bottom-width: 1px;
-  height: 60px;
+  height: 65px;
 `;
 
 const Wrapper = styled.View`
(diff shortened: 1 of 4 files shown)


src/components/Input/InputWithSwitch.js
```diff
@@ -124,9 +127,18 @@ const ModalTitle = styled(MediumText)`
...
 
 export default class InputWithSwitch extends React.Component<Props, State> {
-  state = {
-    showModal: false,
-  };
+  scrollViewRef: React.ElementRef<FlatList>;
+
+  constructor(props: Props) {
+    super(props);
+    this.scrollViewRef = React.createRef();
```
poocart: This not needed as it's not actual ref.

PR #1752, opened 2020-03-17
Fix/modal scroll fix

vyvil:
Due to scroll to dismiss latency on attempt to fix this issue (on #1751), I'm adding this PR that
fixes scrolling issue on county select modal by disabling scroll to dismiss feature.


src/components/Input/SelectList.js
```diff
@@ -90,6 +90,17 @@ export default class SelectList extends React.Component<Props, State> {
           renderItem={this.renderListItem}
           keyExtractor={({ name }) => name}
           keyboardShouldPersistTaps="handled"
+          initialNumToRender={10}
+          viewabilityConfig={{
+            minimumViewTime: 300,
```
graszka22: do we need this? on Android it causes content to disappear when scrolling fast and I don't think rendering a country name is so costly that we need such drastic optimisations

vyvil: yeah, makes sense, I'll make some adjustments

PR #1754, opened 2020-03-17
Feature/images for theme modes

vyvil:
This PR:
- adds method to get image by theme and key;
- changes pattern's fallback image on dark mode;
- changes generic token image (for a sake of simplicity, I've added a `fallbackToGenericToken` boolean to `ListItemWithImage` so other components that uses it won't be needed to be wrapped in `withTheme` to pass theme in this method to get fallback image by current theme).


src/components/TextInput/TextInput.js
```diff
@@ -523,6 +526,9 @@ class TextInput extends React.Component<Props, State> {
     const { value = '', selectorValue = {}, label } = inputProps;
     const { selector = {}, input: inputValue } = selectorValue;
     const textInputValue = inputValue || value;
+    const { genericToken: _genericToken } = images(theme);
+    const genericToken = fallbackToGenericToken && _genericToken;
+    const fallbackSource = genericToken || _fallbackSource;
```
poocart: Naming looks very complex, maybe this would make it read simpler (plus suggesested change above)?
```suggestion
    if (fallbackToGenericToken) {
        const { genericToken: defaultGenericToken } = images(theme);
        fallbackSource = defaultGenericToken;
    }
```

src/components/TextInput/TextInput.js
```diff
@@ -508,7 +510,8 @@ class TextInput extends React.Component<Props, State> {
       getInputRef,
       errorMessageStyle,
       innerImageURI,
-      fallbackSource,
+      fallbackSource: _fallbackSource,
```
poocart: Per suggestion that is below:
```suggestion
```

PR #1756, opened 2020-03-19
Exchange fixes

graszka22:
This PR solves the following issues:
- Popular swaps should be updated to first show these pairs: ETH -> PLR, ETH -> DAI, DAI -> ETH, PLR -> ETH.
- Popular swaps grid should only show 6 pairs.
- Popular swaps grid should not be on keyboard avoid view.
- Popular tokens should be visible on both, Sell and Buy, selectors. Fiat tokens should be shown as first when entering exchange via “Fiat exchange’ menu item on Services tab.
- Exchange search input borders should be updated as per new designs.
- Do not show “No offers” message this early. If not possible to determine if there def. won’t be any offers, add a time out for empty state message of 5 seconds. (Disable “Updated on real time” blinking message if we def. know if there are no offers coming in.
- Options should not be filtered out based on exchange availability (this is feature revert, yet it is needed to allow smoother and more understandable search experience.
- Advanced options should be renamed into speed settings and x button should be removed from it’s modal.
- Exchange count indicators on cards on Service screen should show count of providers not offers.


src/components/TextInput/TextInput.js
```diff
@@ -718,21 +744,14 @@ class TextInput extends React.Component<Props, State> {
                   marginBottom="0"
                 />
               </SearchBarWrapper>
+              {!!displayFiatOptionsFirst && !!fiatOptions.length &&
+                this.renderHorizontalOptions(fiatOptions, fiatOptionsTitle)
+              }
```
poocart: To avoid DRY maybe you could add `data.length` check within `renderHorizontalOptions` itself and return null if no data, then another suggest would be to call `this.renderHorizontalOptions(fiatOptions, fiatOptionsTitle)` before as `const renderedHorizontalOptions` and in then end it would just be `{!!displayFiatOptionsFirst && renderedHorizontalOptions}` or below – `{!displayFiatOptionsFirst && renderedHorizontalOptions}`. What do you think?

src/components/TextInput/TextInput.js
```diff
@@ -718,21 +744,14 @@ class TextInput extends React.Component<Props, State> {
...
-                  renderItem={this.renderHorizontalOption}
-                  horizontal
-                  contentContainerStyle={{ paddingHorizontal: spacing.layoutSides, paddingVertical: spacing.medium }}
-                  ItemSeparatorComponent={() => <View style={{ width: 26, height: 1 }} />}
-                />
-              </HorizontalOptions>
+                this.renderHorizontalOptions(filteredHorizontalListData, horizontalOptionsTitle)
+              }
+              {!displayFiatOptionsFirst && !!fiatOptions.length &&
+                this.renderHorizontalOptions(fiatOptions, fiatOptionsTitle)
```
poocart: Suggesting include change per comment above.

src/components/TextInput/TextInput.js
```diff
@@ -496,6 +499,26 @@ class TextInput extends React.Component<Props, State> {
...
+          keyExtractor={({ name }) => name}
+          keyboardShouldPersistTaps="always"
+          renderItem={this.renderHorizontalOption}
+          horizontal
+          contentContainerStyle={{ paddingHorizontal: spacing.layoutSides, paddingVertical: spacing.medium }}
+          ItemSeparatorComponent={() => <View style={{ width: 26, height: 1 }} />}
+        />
+      </HorizontalOptions>
+    );
+  }
```
poocart:
```suggestion
  };
```

PR #1757, opened 2020-03-19
[PIL-444] Feature/smart wallet exchange

alazarevski:
Smart Wallet Id is the default recipient of the exchange transaction (does not matter if user uses the legacy or smart wallet in assets)
If there is no smart wallet (edge case) prompt user to initiate a smart wallet account although enabling exchange for legacy wallet as well


src/utils/smartWallet.js
```diff
@@ -91,6 +91,15 @@ export const userHasSmartWallet = (accounts: Accounts = []): boolean => {
   return accounts.some(acc => acc.type === ACCOUNT_TYPES.SMART_WALLET);
 };
 
+export const getPreferredWalletId = (accounts: Accounts = []): string => {
+  if (userHasSmartWallet(accounts)) {
+    const smartWallet = accounts.find(acc => acc.type === ACCOUNT_TYPES.SMART_WALLET);
+    return smartWallet ? smartWallet.walletId : '';
+  }
```
grilix: I think we could avoid the extra search, while making the code just a bit simpler at the same time:

```suggestion
  const smartWallet = accounts.find(({ type }) =>
    type === ACCOUNT_TYPES.SMART_WALLET);
  if (smartWallet) {
    return smartWallet.walletId;
  }
```

src/screens/Exchange/ExchangeConfirm.js
```diff
@@ -345,12 +351,19 @@ class ExchangeConfirmScreen extends React.Component<Props, State> {
...
               buttonText="Legacy Wallet"
               rightIconProps={{ name: 'selector', style: { fontSize: 16 } }}
               onPress={() => navigation.navigate(EXCHANGE_RECEIVE_EXPLAINED)}
               wrapperStyle={{ marginTop: 0 }}
-            />
+            />}
+            {!!hasSmartWallet &&
+              <SettingsWrapper>
+                <BaseText secondary regular center style={{ marginBottom: 0 }}>
+                  The assets will be transferred to your Smart Wallet.
```
jsidorenko: Shouldn't this phrase depend on the wallet you selected to receive to?

src/screens/Exchange/ExchangeReceiveExplained.js
```diff
@@ -23,6 +23,9 @@ import type { NavigationScreenProp } from 'react-navigation';
 import { Wrapper } from 'components/Layout';
 import ContainerWithHeader from 'components/Layout/ContainerWithHeader';
 import { Paragraph } from 'components/Typography';
+import Button from 'components/Button';
+
+import { SMART_WALLET_INTRO } from 'constants/navigationConstants';
```
jsidorenko:
```suggestion
// constants
import { SMART_WALLET_INTRO } from 'constants/navigationConstants';
```

src/utils/smartWallet.js
```diff
@@ -91,6 +91,16 @@ export const userHasSmartWallet = (accounts: Accounts = []): boolean => {
   return accounts.some(acc => acc.type === ACCOUNT_TYPES.SMART_WALLET);
 };
 
+export const getPreferredWalletId = (accounts: Accounts = []): string => {
```
jsidorenko: you can just use the `getActiveAccountWalletId()` function instead

src/screens/Exchange/ExchangeConfirm.js
```diff
@@ -345,12 +351,19 @@ class ExchangeConfirmScreen extends React.Component<Props, State> {
 
               </AllowanceWrapper>
             }
-            <ButtonText
+            {!hasSmartWallet && <ButtonText
               buttonText="Legacy Wallet"
               rightIconProps={{ name: 'selector', style: { fontSize: 16 } }}
```
jsidorenko: Shouldn't that button allow to select where do you want to receive your assets?

src/utils/smartWallet.js
poocart: @jsidorenko not sure about this, looking at whole implementation this seems that this method and whole logic describes that this returns "default wallet address to receive exchange result" and based on PR we see that we can either receive to our Smart Wallet if it's deployed and otherwise – Key Based. But then there's no switch as per designs so not sure if the logic in this PR follows the design @alazarevski?

alazarevski: Per discussion with @paragramm and @vitorpy, we will use the SmartWallet walletId as the default receiver, even though the active account might be legacy. The only exception in this method is if there is no smart-wallet at all (edge case) is when we return the legacy walletId.

src/screens/Exchange/ExchangeConfirm.js
alazarevski: With the latest changes the smart wallet will be the default receiver (image1). We agreed to have this button only to guide users (edge-case legacy only) to create a smart wallet account. (images 2, 3)

src/screens/Exchange/ExchangeConfirm.js
alazarevski: With the latest changes the smart wallet will be the default receiver (image1). We agreed to have this button only to guide users (edge-case legacy only) to create a smart wallet account. (images 2, 3)

jsidorenko:
@alazarevski just got an update from Dmitry regarding latest decisions. Most of my comments became irrelevant, sorry for the noise :D going to quickly review this PR once again now

src/utils/smartWallet.js
```diff
@@ -91,6 +91,16 @@ export const userHasSmartWallet = (accounts: Accounts = []): boolean => {
...
+  const smartWallet = accounts.find(({ type }) =>
+    type === ACCOUNT_TYPES.SMART_WALLET);
+  if (smartWallet) {
+    return smartWallet.walletId;
+  }
+  const legacyWallet = accounts.find(acc => acc.type === ACCOUNT_TYPES.KEY_BASED);
```
jsidorenko:
```suggestion
  const legacyWallet = findKeyBasedAccount(accounts);
```

src/utils/smartWallet.js
```diff
@@ -91,6 +91,16 @@ export const userHasSmartWallet = (accounts: Accounts = []): boolean => {
   return accounts.some(acc => acc.type === ACCOUNT_TYPES.SMART_WALLET);
 };
 
+export const getPreferredWalletId = (accounts: Accounts = []): string => {
+  const smartWallet = accounts.find(({ type }) =>
```
jsidorenko: How about creating a new helper `findFirstSmartAccount` ?
I had the same need for this method in another task, so I think it's worth having it

PR #1762, opened 2020-03-20
Feature/referral reward screen UI [PIL-233]

vyvil:
This PR adds "Thanks for using Pillar modal". Just UI part.
To see it, set `showRewardModal` to true on Home.js.


src/components/ReferralRewardModal/ReferralModalReward.js
```diff
@@ -0,0 +1,133 @@
...
+            <HeaderBlock
+              centerItems={[{ title: 'Thank you for using Pillar' }]}
+              rightItems={[{ close: true }]}
+              onClose={onModalHide}
+              noBack
+            />
+            <StyledScrollView contentContainerStyle={{ flexGrow: 1 }}>
+              <ContentWrapper>
+                <Confetti source={confettiTop} resizeMode="cover" top height={screenWidth * 0.73} />
+                <Confetti source={confettiBottom} resizeMode="contain" height={screenWidth * 0.48} />
```
poocart: 🎊 🎉 🎈

src/screens/Home/Home.js
```diff
@@ -558,6 +553,11 @@ class HomeScreen extends React.Component<Props, State> {
             contentContainerStyle={{ flexGrow: 1 }}
           />
         </ScrollView>
+        <ReferralModalReward
+          isVisible={showRewardModal}
+          onModalHide={(callback) => this.setState({ showRewardModal: false },
```
poocart: What do you think if just move this navigation `callback` from component level to screen and avoid that `callback` in general?

vyvil: Initially I did so,  but then figured that it's better not to add state to Modal component and keep it on parent component. 
And by adding state to ReferralModalReward, additional methods are needed to track isVisible prop changes. So I thought callback solves modal closing in cleaner way.

If those arguments are not that sound, I can adjust this part :)

PR #1764, opened 2020-03-23
Restructure notifications subscribe actions

jsidorenko:
The `startListeningNotificationsAction()` method got hard to read, so I've split it into 2 new actions


src/actions/notificationsActions.js
```diff
@@ -70,7 +70,7 @@ import {
   CONNECTION_DISCONNECTED_EVENT,
   CONNECTION_REJECTED_EVENT,
   CONNECTION_REQUESTED_EVENT,
-  CONNECTION_COLLECTIBLE_EVENT,
+  COLLECTIBLE_EVENT,
```
poocart: 👍

PR #1766, opened 2020-03-23
[PIL-448, PIL-450] Feature/btc fiat exchange

alazarevski:
Enabled Fiat Exchange to BTC if the user has activated the BTC address
Fixed a bug showing all supported assets in the update form options function


src/screens/Exchange/Exchange.js
```diff
@@ -979,6 +995,32 @@ class ExchangeScreen extends React.Component<Props, State> {
...
 
+  generateBTCAssetOption = () => {
+    const symbol = 'BTC';
+    const {
+      btcAddresses,
+      btcBalances,
+      baseFiatCurrency,
+      rates,
+    } = this.props;
+    const { address } = btcAddresses[0];
```
poocart:
```suggestion
    const [{ address }] = btcAddresses;
```

src/screens/Exchange/Exchange.js
```diff
@@ -979,6 +995,32 @@ class ExchangeScreen extends React.Component<Props, State> {
...
+    const assetBalance = rawAssetBalance ? formatAmount(rawAssetBalance) : null;
+    const formattedBalanceInFiat = getFormattedBalanceInFiat(baseFiatCurrency, assetBalance, rates, symbol);
+    const btcAsset = initialAssets.find(e => e.symbol === symbol);
+    return {
+      key: symbol,
+      value: symbol,
+      icon: btcAsset ? btcAsset.iconUrl : '',
+      iconUrl: btcAsset ? btcAsset.iconUrl : '',
+      symbol,
+      ...btcAsset,
```
poocart: What if no `btcAsset`?

src/screens/Exchange/Exchange.js
```diff
@@ -979,6 +995,32 @@ class ExchangeScreen extends React.Component<Props, State> {
...
+    const addressBalance = btcBalances[address];
+    const rawAssetBalance = addressBalance ? satoshisToBtc(addressBalance.confirmed) : 0;
+    const assetBalance = rawAssetBalance ? formatAmount(rawAssetBalance) : null;
+    const formattedBalanceInFiat = getFormattedBalanceInFiat(baseFiatCurrency, assetBalance, rates, symbol);
+    const btcAsset = initialAssets.find(e => e.symbol === symbol);
+    return {
+      key: symbol,
+      value: symbol,
+      icon: btcAsset ? btcAsset.iconUrl : '',
+      iconUrl: btcAsset ? btcAsset.iconUrl : '',
```
poocart: `btcAsset` is always present in `initialAssets` so we can just deconstruct `iconUrl` in above and avoid code duplicate.

alazarevski: flow acted up on that case since find can return undefined

PR #1767, opened 2020-03-23
PIL-338 Feature/update services

jdnmr:
Note: this also slightly changes the Menu screen according to the new design.


src/screens/Services/Services.js
```diff
@@ -70,7 +74,13 @@ class ServicesScreen extends React.Component<Props> {
...
       {
         key: 'buyCryptoWithFiat',
-        title: 'Buy crypto with fiat',
+        // hack to avoid inline images because of iOS13 issue. Likely can be dropped in RN 0.62
+        title: [
+          'Buy crypto with ',
+          <Image source={mastercardIcon} style={{ marginBottom: 1 }} height={11} />,
+          ' & ',
+          <Image source={visaIcon} height={11} />,
+        ],
```
grilix: Maybe we can have a custom component for this? For example:

```jsx
const Wrapper = styled.View``; // or whatever
const BuyTitle = () => (
  <Wrapper>
    <Text>Buy crypto with </Text>
    <Image source={mastercardIcon} style={{ marginBottom: 1 }} height={11} />
    <Text> &amp; </Text>
    <Image source={visaIcon} height={11} />
  <Wrapper>
);
```

Then we could do just something like:

```jsx
  title: <BuyTitle />
```

jdnmr: That's basically what we're doing, but inside `ListCard`. TBH I think it makes more sense to do this internally in `ListCard` than to set styles for child components here in a screen component

PR #1769, opened 2020-03-24
[PIL-445] Connections v3

jsidorenko:
I know, it's massive, sorry :(
The main idea is to remove everything identityKeys/accessToken related.

P. S. The backend should be merged on 25/03


src/actions/authActions.js
```diff
@@ -144,7 +136,7 @@ export const loginAction = (
 
       if (pin) {
         const saltedPin = await getSaltedPin(pin, dispatch);
-        const decryptionOptions = generateNewConnKeys ? { mnemonic: true } : {};
+        const decryptionOptions = {};
         wallet = await decryptWallet(encryptedWallet, saltedPin, decryptionOptions);
```
poocart: Options of `decryptWallet` is optional param, right?
```suggestion
        wallet = await decryptWallet(encryptedWallet, saltedPin);
```

jsidorenko: According to spec - yes :)

src/actions/contactsActions.js
```diff
@@ -87,26 +86,9 @@ export const syncContactAction = (userId: string) => {
...
-    } else {
-      const accessToken = accessTokens.find(token => token.userId === userId);
-      if (!accessToken) return;
-      userInfo = await api.userInfoById(userId, {
-        walletId,
-        userAccessKey: accessToken.myAccessToken,
-        targetUserAccessKey: accessToken.userAccessToken,
-      });
-    }
+    const userInfo = await api.userInfoById(userId, walletId);
```
poocart: `userInfoById` requires objet payload:
```suggestion
    const userInfo = await api.userInfoById(userId, { walletId });
```

src/screens/PinCodeUnlock/PinCodeUnlock.js
```diff
@@ -122,10 +121,10 @@ class PinCodeUnlock extends React.Component<Props, State> {
   };
 
   showBiometricLogin() {
-    const { loginWithPrivateKey, connectionKeyPairs: { data: connKeys, lastConnectionKeyIndex } } = this.props;
+    const { loginWithPrivateKey } = this.props;
     const { biometricsShown, supportedBiometryType } = this.state;
 
-    if (biometricsShown || connKeys.length <= PRE_KEY_THRESHOLD || lastConnectionKeyIndex === -1) {
+    if (biometricsShown) {
       Toast.show({
```
alazarevski: Not sure if we need this toast here

alazarevski:
Could you also remove the android/app/src/main/assets/threads directory and contents
Also ios unlink : ios/index.thread.jsbundle and ios/pillarwallet.xcodeproj/project.pbxproj lines where index.thread.jsbundle is referenced like 57, 606, 728, 1701

src/screens/PinCodeUnlock/PinCodeUnlock.js
jsidorenko: Not sure as well :D But since we're going to change the auth flow in the next PR, I think we could leave this

poocart:
@alazarevski re cleanup – this is done on new RN branches, I would appreciate if this PR would not include those changes as there might be huge conflicts. As well my branch contains other changes for threads such as removal of main index thread, run scripts, etc.

PR #1770, opened 2020-03-25
Feature/refer friends flow [PIL-313]

vyvil:
This PR extends refer friends flow:
1. Updates phone contacts list, by filtering out repeated phone numbers / emails.
2. Allows referral flow only if phone or email is added and verified, and allows to select only allowed contacts to refer (only emails if only email is verified and only phone number if only phone is verified).
3. Adds loaders, alerts and success messages where required for phone contacts fetching and invitations sending.
4. Allows to send only 10 invitations per day (front end covers invitations from one device, we do not get already invited count from backend)
5. Simplifies Checkbox component (we do not need state there, as there are no more animation used on this component)
6. Adds closable pill component to show selected contacts to invite.
7. Stores already invited contacts and daily invitations count into local storage.


src/reducers/referralsReducer.js
```diff
@@ -32,30 +53,155 @@ export type ReferralsSendingInviteAction = {|
...
+      return setAlreadySentInvites(state, action);
+
+    case SET_CONTACTS_FOR_REFERRAL:
+      return setContacts(state, action);
+
+    case REMOVE_CONTACT_FOR_REFERRAL:
+      return removeContact(state, action);
+
+    case ALLOW_ACCESS_PHONE_CONTACTS:
+      return { ...state, hasAllowedToAccessContacts: true };
```
grilix: What's this for? It looks like we don't use it.

src/actions/referralsActions.js
```diff
@@ -82,17 +118,37 @@ export const sendReferralInvitationsAction = (invitations: ReferralInvitation[])
...
+          let updatedInvitationCount = count + invitationContacts.length;
+          const currentDate = new Date().toJSON().slice(0, 10);
+          if (date !== currentDate) updatedInvitationCount = invitationContacts.length;
+          const updatedAlreadyInvitedContacts = [...alreadyInvitedContacts, ...invitationContacts];
+          await dispatch(saveDbAction('referralData', {
+            referrals: {
+              alreadyInvitedContacts: updatedAlreadyInvitedContacts,
+              sentInvitationsCount: { count: updatedInvitationCount, date: currentDate },
+            },
+          }));
```
grilix: I think we can just store this on the redux state, and the persist it.

src/reducers/referralsReducer.js
vyvil: I use it here:
https://github.com/pillarwallet/pillarwallet/pull/1770/files/555baa96dce2e76c67392bda120abc3d7f5b905f#diff-3849a1b9cacfa664dfb332b2993fda6bR95
To check where to navigate user - to contacts list or to this screen explaining our need of accessing user's address book

src/actions/referralsActions.js
vyvil: Hm, I've followed our usual way of persisting data - storing it to local db using saveDbAction.
Or I got you wrong?

Also, this part should be reworked once we will get available invitation count form backend :)

PR #1771, opened 2020-03-25
[PIL-451] Fix/Collectible Send Navigation crashFix

alazarevski:
Fix Collectible navigation crash on re-auth and navigate forward and back.


vyvil:
Hm, after those changes there's something not right with the stack on Collectible send flow. When user clicks on "Magic" when transaction is done, we are calling navigation.dismiss() (that should bring us to the parent stack) and user ends up at "

jsidorenko:
Agree, those changes also look suspicious to me

PR #1772, opened 2020-03-25
PIL-339 Feature/exchange intro

jdnmr:


src/components/OverlayModal/OverlayModal.js
```diff
@@ -0,0 +1,110 @@
...
+  height: ${({ height }) => height};
+  justify-content: center;
+`;
+
+const Spacer = styled.View`
+  height: 30%;
+`;
+
+export default class OverlayModal extends React.Component<Props, State> {
+    state: State = {
```
jsidorenko: Pls fix the indentation

src/actions/appActions.js
```diff
@@ -192,6 +193,9 @@ export const initAppAndRedirectAction = (appState: string, platform: string) =>
       const { insights = {} } = await storage.get('insights');
       dispatch({ type: SET_INSIGHTS_STATE, payload: insights });
 
+      const { hasSeenExchangeIntro = {} } = await storage.get('hasSeenExchangeIntro');
```
jsidorenko: What do you think about re-using appSettings for that?
I assume we're going to have more screens those should be seen only once, so it's better to re-use some existing storage table rather than flooding the new ones

jsidorenko: We have insights, userSettings and appSettings those could be re-used

vyvil: yeah... and those intro modals were intended for this very reason ;/
I guess there was a huge miscommunication when assigning the tasks.
Jan, we can discuss it in slack if you like. We have UI for those modals and a way to store the state of seen or not seen already

src/constants/exchangeConstants.js
```diff
@@ -44,3 +44,4 @@ export const MARK_NOTIFICATION_SEEN = 'MARK_NOTIFICATION_SEEN';
 export const MIN_MOONPAY_FIAT_VALUE = 20;
 
 export const EXCHANGE = 'EXCHANGE';
+
```
grilix, line 47:
```suggestion
```

PR #1775, opened 2020-03-26
HotFix RC PR [PIL-451] Fix/Collectible Send Navigation crashFix

alazarevski:
Use unique stacks and stack navigator keys for different flows


poocart:
Added my changes, please re-review.

alazarevski:
@poocart Looks Good with b877f61. There is one more issue related but it is for erc20 sending, will create a separate linear issue for that one

PR #1776, opened 2020-03-26
PIL-294 Activity feed updates

graszka22:
Note for reviewers:
- most of the feed items with their respective designs are listed here: https://www.notion.so/pillarproject/6c33dd0a614f465a83999d05ea34bdc3?v=8479107243f2427b9a406bd54b71dcfa
- some are missing on the list (for example ENS registration) and I've kept them in the app
- some on the list (for example exchanges, referrals) are not implemented yet or there is no backend right now
- the PN settlement item will have updated design soon
- there will be new icons for key and smart wallets  

- I hope the refactoring will make easier to add new items - now the activity feed items have been extracted to another component, every type has their own function with logic for them, the items are rendered in one place so they're consistent
- you don't need to generate every event just to see if it looks alright: that's where storybook shines! just run storybook server `yarn storybook`, find storybook in the settings in the app and boom, you have most of interesting cases (some cases are very redundant) within one click in the storybook navigator!
- but you'll have to generate event if you want to test the logic, sorry 😜 

tl;dr: optimisation stuff, but not implemented yet, not in the scope of this PR
One interesting thing that I've noticed that this component can cause a pretty significant performance lag (I'm not sure yet, I haven't measured it, but it is likely that it's causing a lag especially on entering the Home screen on app startup). On my account I have about 100-200 events. It doesn't seem to be much. If you place the console log or something in the render function of the `ActivityFeedItem` component, it turned out that on the app startup (on the home screen) it was called over 3000 (yep, 3000 😄) times (and no, the refactoring has not much to do with it, it's going in thousands on develop too 😉). The main culprit here is that the FlatList is nested inside the ScrollView. FlatList has an optimisation that it doesn't render everything, but only a visible portion of the items + some more for scrolling. But when it's in the scroll view, it loses the ability to see what items are visible (basically it just spans through entire scroll view and it's not scrolling at all). So it just renders all of them. Simply deleting the ScrollView from the home screen reduces the amount of calls about three times on my account, but we have messed up scrolling 😅 However I think it's solvable. Another issue is that it's probably re-rendering unnecessarily. Things like `shouldComponentUpdate`, not passing objects created every render and not passing inline arrow functions through props will probably help. But the optimisations mentioned here are not in the scope of this PR and I didn't run any measurements etc and this should be probably carded up to do later 😄


src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -0,0 +1,530 @@
...
+        onPress={eventData && (() => selectEvent(eventData, event.type, event.status))}
+        badge={badge}
+        actionLabel={status}
+        actionLabelColor={this.getColor(statusColor)}
+        labelAsButton={statusAsButton}
+        rejectInvitation={rejectInvitation}
+        acceptInvitation={acceptInvitation}
+        itemImageSource={icon}
+        avatarUrl={avatarUrl}
+        itemImageUrl={imageUrl}
```
vyvil: It would be good to add `fallbackToGenericToken` for transaction events, so if asset or collectible icon is missing, a generic token would be rendered (rather than "initials").

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -0,0 +1,530 @@
...
+  render() {
+    const { event, selectEvent } = this.props;
+
+    if (event.type === TRANSACTION_EVENT && event.tag === PAYMENT_NETWORK_TX_SETTLEMENT) {
+      return this.renderSettlement(event);
+    }
+
+    const itemData = this.getEventData(event);
+
+    if (!itemData) return null;
```
vyvil: This might cause us some problems as we have a sectioned list and we are still returning something. So this "null" might get "visible" empty section with header will be rendered.
I caught a case (but I believe it's happening because we have an issue with connections), when this happens:
A and B users. B user invites A. A rejects it. A invites B. (It is now registered as "connectionDisconnectedEvent" and not as invitation - it must be an error coming from connections). And if user has no more event on that day, empty section is being rendered for that day.
Maybe we can first filter all events by the type that we should show?

src/components/ListItem/ListItemWithImage.js
```diff
@@ -506,15 +508,15 @@ const Addon = (props: AddonProps) => {
           color={colors.secondaryText}
           margin={0}
           icon="close"
-          fontSize={fontSizes.regular}
+          fontSize={fontSizes.big}
```
vyvil: Hm, there's something wrong with those icons. They get cut and neither X nor tick is shown properly

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -0,0 +1,530 @@
...
+      name: username,
+      status,
+      avatarUrl: profileImage,
+      eventData: { ...event },
+      statusAsButton: type === TYPE_SENT,
+    };
+
+    if (type === TYPE_RECEIVED) {
+      data.information = 'Connection request';
+      data.rejectInvitation = () => createAlert(TYPE_REJECTED, event, () => onRejectInvitation(event));
```
vyvil: `rejectInvitation` and `acceptInvitation` are not being passed as functions, hence trying to accept or reject invite results in error: `rejectInvitation is not a function` or `acceptInvitation is not a function`

src/components/ListItem/ListItemWithImage.js
```diff
@@ -126,7 +129,7 @@ const InnerWrapper = styled.TouchableOpacity`
   flex-direction: row;
   align-items: ${props => props.horizontalAlign || 'center'};
   justify-content: center;
-  padding: ${spacing.medium}px ${spacing.layoutSides}px;
+  padding: 14px 20px;
```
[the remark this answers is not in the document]
graszka22: Dmitry explicitly said that the spacings here should be 14px and 20px. `spacing.layoutSides` is 16px.

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -0,0 +1,530 @@
...
+  type?: string,
+  asset?: string,
+  isPending?: boolean,
+  supportedAssets: Asset[],
+  selectEvent: Function,
+  contacts: ApiUser[],
+  contactsSmartAddresses: ContactSmartAddressData[],
+  ensRegistry: EnsRegistry,
+  theme: Theme,
+  event: Object,
```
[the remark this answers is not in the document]
graszka22: Yeah, it would be nice, but we have different types of events (the transaction event is very different than for example user event). Also we'd need to refactor the types all over the place which is out of scope of this PR.

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -0,0 +1,530 @@
...
+        trxData.hideAmount = true;
+        trxData.hideSender = true;
+        break;
+      default:
+        const address = isReceived ? event.from : event.to;
+        const usernameOrAddress = event.username
+            || ensRegistry[address]
+            || elipsizeAddress(address);
+        const isPPNTransaction = get(event, 'isPPNTransaction', false);
+        const information = event.accountType === ACCOUNT_TYPES.KEY_BASED ? 'Key wallet' : 'Smart wallet';
```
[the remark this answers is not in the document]
graszka22: I asked Dmitry about those specifically. Every smart wallet and key wallet transaction has sender and recipient. So those (for example https://zpl.io/agdXRGM) without any adress or contact are actually outdated and instead we'll be having for example these https://zpl.io/VQ8plG4 instead.

src/components/ActivityFeed/ActivityFeedItem.js
graszka22: Can't reproduce that, but added filtering anyway

PR #1777, opened 2020-03-27
Feature/input for verification code

vyvil:
This PR adds Code input for verification code


vyvil:
> * On smaller devices (even on iPhone X) it wraps to a second line (so we have 4 digits in one line, 1 digit in second line), is it intentional? Couldn't we just make these squares smaller?
> * On iOS the caret goes all over the input randomly, I even managed to type two numbers in one square o.o just for a short moment though)
> 
> Screenshot 2020-03-27 at 14 20 45
> 
> * can we add a bit of space around "That didn't work..." label?

Thanks, @graszka22!
Addressed them all. Could you take a second look? :)

graszka22:
Maybe we can remove the caret completely? https://reactnative.dev/docs/textinput#carethidden And leave just focused squares? 
on Android the caret is green and it looks weird too

vyvil:
Yeah, I thought so too :). I hid the whole input from the very beginning, but then thought it might look weird without caret blinking :)) 
But I've added border on the square that's focused. So there's that :)
I'll hide inputs in an upcoming commit.

PR #1778, opened 2020-03-30
PIL-464 Feature/Remove Connections V1

alazarevski:
Remove oldInvitations and Contact build system
Add direction for pending contacts to see which one is sent and which one is received


src/actions/connectionsActions.js
```diff
@@ -67,7 +67,8 @@ export const updateConnectionsAction = (theWalletId?: ?string = null) => {
     const removeContacts = [];
     const removeInvitations = [];
 
-    connections.forEach(connection => {
+    connections.forEach((connection: ApiContact) => {
```
jsidorenko: What do you think about returning from this function if no connections were received from the backend?
`if (isEmpty(connections)) return;`

I'm afraid that you could lose all your contacts if you have a shitty network connection.

src/actions/connectionsActions.js
```diff
@@ -106,23 +98,25 @@ export const updateConnectionsAction = (theWalletId?: ?string = null) => {
...
+    dispatch(saveDbAction('contacts', { contacts: updatedContacts }, true));
+
+    await dispatch({
       type: UPDATE_INVITATIONS,
       payload: updatedInvitations,
     });
-    dispatch(saveDbAction('invitations', { invitations: updatedInvitations }, true));
 
-    dispatch({
+    await dispatch({
```
jsidorenko: what the point of using awaits here?

alazarevski: would it matter if they are async but in order ?   I thought this  await dispatch(getExistingChatsAction()); was needed after state is set

jsidorenko: As I remember there is no need to use `await` if you simply dispatch the object.
But you can easily check this by dispatching some change and checking if it was changed immediately after the call:
```javascript
dispatch({ type: UPDATE_SOMETHING });
const { wasUpdated } = getState().someReducer;
if (!wasUpdate) console.log('I was right!');
```

PR #1781, opened 2020-04-01
Fix/App state changes auth flow

jdnmr:
- Fix white screen of death that appeared when going background => active;
- Disable the 20 seconds background timeout.


src/navigation/appNavigation.js
```diff
@@ -843,26 +843,26 @@ class AppFlow extends React.Component<Props, State> {
...
-        updateSignalInitiatedState(false);
-        stopListeningForBalanceChange();
-      }, SLEEP_TIMEOUT);
+      // lockTimer = BackgroundTimer.setTimeout(() => {
+      const { navigation } = this.props;
+      const pathAndParams = navigation.router.getPathAndParamsForState(navigation.state);
+      const lastActiveScreen = pathAndParams.path.split('/').slice(-1)[0];
+      const lastActiveScreenParams = pathAndParams.params;
+      updateNavigationLastScreenState({ lastActiveScreen, lastActiveScreenParams });
+      navigation.navigate(AUTH_FLOW);
```
poocart, line 860: Just thinking that removing background timer in general might cause issues if app is continuously being switched between background and foreground. I would suggest leave the timer, but remove this part and it will only stop listening for notifications if app has been in background for over 20s while maintaining logged in state.

jsidorenko: agree

PR #1789, opened 2020-04-03
[PIL-468] Missing label in 'Connect'

alazarevski:
Add Connect label in QR button in WalletConnect Screen


src/screens/WalletConnect/QRCodeScanButton.js
```diff
@@ -45,6 +45,7 @@ type State = {
 
 const Container = styled.View`
   align-items: center;
+  margin-top: -15px;
```
graszka22, line 48: why negative margin here? o.o

alazarevski: iPhone 6/7/8 and some similar resolutions just barely cut the Connect label, this looks better for all those models and didn't affect the bigger ones, just for the sake of it

graszka22: Approved, but I'd rather just make existing margins smaller rather than sticking negative margins somewhere else. IMHO it's better to use negative margins only when you really need it, they make things messier.

PR #1791, opened 2020-04-03
[PIL-462] Incative PPN screen

alazarevski:
Hide Settled/Unsettled Tabs when PPN is not activated/toped up


src/screens/Assets/PPNView.js
```diff
@@ -347,11 +347,14 @@ class PPNView extends React.Component<Props, State> {
             color={colors.text}
             bordered
           />}
+          {!!this.props.PPNTransactions.length &&
```
jsidorenko, line 350:
```suggestion
          {!!PPNTransactionsMapped.length &&
```

src/screens/Assets/PPNView.js
```diff
@@ -347,11 +347,14 @@ class PPNView extends React.Component<Props, State> {
...
           <Tabs
             tabs={historyTabs}
             wrapperStyle={{ paddingTop: 16 }}
             activeTab={activeTab}
           />
+          }
+          {!!this.props.PPNTransactions.length &&
```
jsidorenko, line 357:
```suggestion
          {!!PPNTransactionsMapped.length &&
```

poocart:
Can't we just use the same logic that sits in `paymentNetwork` reducer as `isPLRTankInitialised `? It's also populated with the right value on `managePPNInitFlagAction` during import and etc.

src/screens/Assets/PPNView.js
jsidorenko:
```suggestion
          {(!!PPNTransactionsMapped.length || availableStake > 0) &&
```

src/screens/Assets/PPNView.js
jsidorenko:
```suggestion
          {(!!PPNTransactionsMapped.length || availableStake > 0) &&
```

PR #1803, opened 2020-04-07
Fix/ Remove path from wallet state, hardcode instead

alazarevski:
Remove path from state and hardcode the correct eth path

### DIFF
--- a/src/actions/bitcoinActions.js
+++ b/src/actions/bitcoinActions.js
@@ -229,10 +229,6 @@ const updateBTCTransactions = (
   transactions: transactions.filter(tx => !!tx.details),
 });
 
-const bitcoinWalletCreationFailed = (): BitcoinWalletCreationFailedAction => ({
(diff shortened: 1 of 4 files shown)


src/actions/bitcoinActions.js
```diff
@@ -234,7 +234,7 @@ const bitcoinWalletCreationFailed = (): BitcoinWalletCreationFailedAction => ({
 
 export const initializeBitcoinWalletAction = (wallet: EthereumWallet) => {
```
jsidorenko, line 235: Should we change this type as well? Maybe something like `KeychainWalletData = { mnemonic?: string, privateKey: string };`

alazarevski: it will be in a different set of work, maybe an optimization Friday ? :)

jsidorenko: actually there is no need to change this type :) I thought you're getting the data straight from the keychain, but was wrong as it's the same `EthereumWallet` object as it was before

PR #1805, opened 2020-04-07
Ui/offer card

vyvil:
This PR updates offer card by addressing those comments:

1. Offer cards should have bigger and new icons on them.
2. All additional actions on exchange cards (like “Connect” for ShapeShift, “Enable” for asset allowances and min/max required values) should be showed as green button on the left side of the card (instead of “Available” amount).
3. Card itself should not be see-through (as disabled) if user can’t take an offer. Only main card button should look like “disabled”.

(On dark mode crypto exchange icons are not being returned for now - not yet in the server)


src/components/DynamicSizeImage/DynamicSizeImage.js
```diff
@@ -0,0 +1,91 @@
...
+    const { imageSource } = this.props;
+
+    if (!!imageSource && typeof imageSource === 'object' && imageSource.uri) {
+      Image.getSize(imageSource.uri, (width, height) => {
+        if (width && height) {
+          this.setState({
+            width: width / 3,
+            height: height / 3,
+          });
+        }
```
jdnmr, line 64: You're using exact same code a few lines below, so please make this a helper function

src/screens/Exchange/Exchange.js
```diff
@@ -383,7 +384,7 @@ function SelectorInputTemplate(locals) {
   );
 }
 
-function getCardTopButtonData(topButtonProps) {
+function getCardAdditionalButtonData(additionalData) {
```
jdnmr, line 387: nit: this file is way too long already, and most of these helper functions could be moved to utils. Also it wouldn't hurt to check type here. All this should be done as a separate task though I guess. Any thoughts on this?

grilix: If the function returns data that's presentation (like titles, press handlers, etc), it shouldn't be on utls, IMHO. I'd try to separate the code into components. Not entirely sure if this specific case could become its own component, but if not, there are parts that can.

vyvil: Yeah, you are both right :)
File is in deed long, I'd also be up for leaving helper functions next to the rest of related code when it is not globally used and I was thinking about dividing the file into two parts - one for the form and selectors and one for the offers list.

Dmitry has asked me to fix scrolling layout here, so I could divide this file when doing so :)

src/components/DynamicSizeImage/DynamicSizeImage.js
```diff
@@ -51,26 +51,23 @@ class DynamicSizeImage extends React.Component<Props, State> {
...
       });
     } else {
       const { width, height } = Image.resolveAssetSource(imageSource) || {};
-      if (width && height) {
-        this.setState({
-          width: width / 3,
-          height: height / 3,
-        });
-      }
+      if (width && height) this.setImageSize(width, height);
```
jdnmr, line 70: `if (width && height)` can also be included inside `setImageSize` :)

vyvil: Yeah, I thought about it :)
But I preferred to keep method clean and leave logic outside of it.

PR #1807, opened 2020-04-07
PIL-344 Update transaction sent screen

jdnmr:


jdnmr:
> Ok, but I suppose this should conform to the designs here: https://app.zeplin.io/project/5db9b5d14c30ae56faaa3030/screen/5e1f0cf499c3e9a7a2806ca2. And we have social media buttons and "Share the love text" which I don't see in the app. The title looks bigger on the designs. Or is the design outdated?

Good catch with the font sizes, corrected 👍 
As of the social media, it should stay commented out - confirmed with Dmitry.

jdnmr:
> * no margin between the title and paragraph - it should be 7px
> * you can use the `BasicText` and `MediumText` components from `components/Typography` - they have nice props for setting size and color

Margin fixed. As of `BasicText` and `MediumText`, the component already uses `Title` and `Paragraph` components that deal with styles too, so I don't want to mix approaches.

PR #1808, opened 2020-04-07
PIL-470 Activity feed performance improvements

graszka22:
#### What I've done
- removed ScrollView from HomeScreen: instead we pass header and tabs components to activity feed list and we're scrolling the list, not the ScrollView
- because we have sticky tabs we can't just put everything in a ListHeader prop. So here is a trick: I turned SectionList into FlatList. Header, tabs, section headers are just elements in the FlatList. I just render them accordingly to their type. And we make tabs sticky using `stickyHeaderIndices` prop.
- I only did that on HomeScreen for now, on other screens it's probably not worth it, there are not many items in the list.  But if we need it somewhere else, let me know, it's easy to do on other screens now.

- another thing that I've noticed is that we're keeping calculated result of `generateFeedSections` function in the state. And update it in `componentDidUpdate`. So now we have the second re-render after we change the state, resulting in two times more re-renders of the list. I realize that this function can be costly and probably it's in the state for cache. So I got rid of that in the state and memoized the result of the function (so it's called on every render but only when the props change). That way we have exactly the same behaviour but without unecessary re-renders! More here: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

- I took a look also at the `ActivityFeedItem`. It's connected to much of the redux state. And it's frequently re-rendered because redux state references change but in reality redux state doesn't change. So I added `shouldComponentUpdate` method. Yay, but now we run it many times. And it can be costly again. So additionally I added some more specific selectors. Now we don't have to compare the whole `assets` object but just what we need: one asset decimals. We don't have to compare the whole `smartWallet` object but just whether we have wallet activated. It also helps with re-renders if we don't have `shouldComponentUpdate`, with specific selectors we don't have whole objects that can change references - just primitive values like int, boolean or string that don't change so much. Bonus: it helps reduce code duplication and they have memoization built in!

#### Lessions learned
✅ don't nest ScrollViews
✅ don't keep cached values in state and update them in componentDidUpdate - use for example memoize
✅ use selectors - **select** what you need, just primitive values are better than whole objects

#### Results
I've counted all renders of `ActivityFeedItem` in Home screen before and after refactor:
before: ~3900
after: 83

Also, to check the real gain I've measured time to switch the tabs in Home screen (all, transactions, social), on Android device with DEV=false. 
first switch, just after opening screen:
before: 12s (!)
after: 4s
next switches:
before: 4.5s
after: 1s  🚀


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -18,15 +18,13 @@
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 import * as React from 'react';
-import { connect } from 'react-redux';
 import isEqual from 'lodash.isequal';
 import type { NavigationScreenProp } from 'react-navigation';
 import styled, { withTheme } from 'styled-components/native';
-import { createStructuredSelector } from 'reselect';
+import memoize from 'memoize-one';
```
jsidorenko, line 24: 😍

jsidorenko:
This is f*king awesome!!! You rock!!

PR #1815, opened 2020-04-08
Activity feed details updates

graszka22:
- Update smart wallet activate event layout: https://zpl.io/anR0lrv
- Update settle event layout: https://zpl.io/brQ7REL
- Update top up event layout: https://zpl.io/boz5nNg (when pillar network selected), 

Screenshot 2020-04-08 at 18 30 22

(when smart wallet selected)

- Update status colors (no green ones, just gray)
- Update SW -> KW, KW -> SW transactions edge cases, for example: if we're on KW and we send money from KW to SW, then we should show the subtext "to Smart Wallet": currently we show "Key Wallet"


src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -186,29 +198,75 @@ export class ActivityFeedItem extends React.Component<Props> {
     return (smartWalletStatus.status !== SMART_WALLET_UPGRADE_STATUSES.DEPLOYMENT_COMPLETE);
   }
 
+  findAccount = (address: string) => {
```
jsidorenko, line 201: We have the `findAccountById` util for that

jsidorenko: And it could be worth adding the `addressesEqual` comparison into the `findAccountById()`

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -186,29 +198,75 @@ export class ActivityFeedItem extends React.Component<Props> {
...
+    const { accounts } = this.props;
+    return accounts.find(account => addressesEqual(account.id, address));
+  }
+
+  isSWAddress = (address: string) => {
+    const account = this.findAccount(address);
+    return (account && account.type === ACCOUNT_TYPES.SMART_WALLET);
```
jsidorenko, line 208: we have `checkIfSmartWalletAccount` for that

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -186,29 +198,75 @@ export class ActivityFeedItem extends React.Component<Props> {
...
+  }
+
+  isKWAddress = (address: string) => {
+    const account = this.findAccount(address);
+    return (account && account.type === ACCOUNT_TYPES.KEY_BASED);
```
jsidorenko, line 213: we have `checkIfKeyBasedAccount` for that

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -186,29 +198,75 @@ export class ActivityFeedItem extends React.Component<Props> {
...
+      const value = new BigNumber(rawValue);
+      if (!valueByAsset[symbol]) {
+        valueByAsset[symbol] = { ...trx, value, decimals };
+      } else {
+        const { value: currentValue } = valueByAsset[symbol];
+        valueByAsset[symbol].value = currentValue.plus(value);
+      }
+    });
+
+    const valuesArray = Object.keys(valueByAsset).map((key) => valueByAsset[key]);
```
jsidorenko, line 242: What do you think about that approach?
```suggestion
    const valuesArray = (Object.values(valueByAsset): any);
```

src/components/ActivityFeed/ActivityFeedItem.js
jsidorenko: Or you could use `findAccountByAddress()`

PR #1819, opened 2020-04-09
Feature/wallet toggle and Home screen touch-ups [PIL-494]

vyvil:
This PR adds Wallet switcher to Home screen and addresses comments in PIL-494.


src/selectors/wallets.js
```diff
@@ -0,0 +1,61 @@
...
+    const { SMART_WALLET_ENABLED: smartWalletFeatureEnabled, BITCOIN_ENABLED: bitcoinFeatureEnabled } = featureFlags;
+    const keyWallet = accounts.find(({ type }) => type === ACCOUNT_TYPES.KEY_BASED) || {};
+    const availableWallets = [{ ...keyWallet, isActive: isBitcoinActive ? false : keyWallet.isActive }];
+
+    if (smartWalletFeatureEnabled) {
+      const smartWallet = accounts.find(({ type }) => type === ACCOUNT_TYPES.SMART_WALLET);
+      if (smartWallet) {
+        availableWallets.unshift({
+          ...smartWallet,
+          isActive: isBitcoinActive ? false : smartWallet.isActive,
```
grilix, line 51: I feel like without the ternary is easier to read, but not entirely sure (same for the key wallet). Your call:

```suggestion
          isActive: !isBitcoinActive && smartWallet.isActive,
```

src/screens/Home/Home.js
```diff
@@ -452,7 +456,7 @@ class HomeScreen extends React.Component<Props, State> {
             <React.Fragment>
               <ScrollView
                 style={{ width: '100%', flex: 1 }}
-                stickyHeaderIndices={referralsFeatureEnabled ? [4] : [3]}
+                stickyHeaderIndices={referralsFeatureEnabled ? [3] : [2]}
```
grilix, line 459: What do these numbers mean exactly? Can we put a name on them?

vyvil: It means an index of items that should become sticky. This has been changed after ActivityFeed improvements

src/selectors/wallets.js
```diff
@@ -0,0 +1,63 @@
+// @flow
```
jsidorenko, line 1: We need to add the copyright/disclaimer here

PR #1820, opened 2020-04-09
Allow referring users outside contacts list

grilix:
This will allow users to type an email or phone into the search bar and invite it directly without the need of having that person in the contacts list.


src/screens/ReferFriends/ReferralContacts.js
```diff
@@ -234,9 +232,17 @@ class ReferralContacts extends React.PureComponent<Props, State> {
     } = this.props;
 
     const showConfirmButton = !!(selectedContacts.length || addedContactsToInvite.length);
-    const missingType = getInfoTypes(isPhoneVerified, isEmailVerified);
-    const allowedContacts = !missingType ? phoneContacts : phoneContacts.filter((contact) => !contact[missingType]);
-    const filteredContacts = getFilteredContacts(allowedContacts, query);
+    const allowedContacts = filterAllowedContacts(phoneContacts, isPhoneVerified, isEmailVerified);
+    const isSearching = query && query.length > MIN_QUERY_LENGTH;
```
poocart, line 236: Per search block settings below:
```suggestion
    const isSearching = query && query.length >= MIN_QUERY_LENGTH;
```

src/screens/ReferFriends/ReferralContacts.js
```diff
@@ -88,30 +90,27 @@ const ButtonWrapper = styled.View`
...
-
 const MIN_QUERY_LENGTH = 3;
 
-const getInfoTypes = (isPhoneVerified, isEmailVerified) => {
-  // one or another should be true, otherwise feature is not available at all
-  if (!isPhoneVerified) {
-    return 'phone';
-  } else if (!isEmailVerified) {
-    return 'email';
+const createCustomContact = (
```
poocart, line 95: It has only 1 return?

src/screens/ReferFriends/ReferralContacts.js
```diff
@@ -88,30 +90,27 @@ const ButtonWrapper = styled.View`
...
+  isPhoneVerified: boolean,
+  isEmailVerified: boolean,
+): ?ReferralContact => {
+  const contact = {
+    id: 'custom-contact',
+    name: query,
+  };
+
+  if (isPhoneVerified && isValidPhone(query)) {
+    return { ...contact, phone: query };
```
poocart, line 106: It has only 1 return?

src/screens/ReferFriends/ReferralContacts.js
grilix: You're right! :+1:

PR #1824, opened 2020-04-13
Fixes on referral flow

grilix:
- Update test app link
- Update reducer so the token is kept after claiming reward
- Add notices


src/actions/referralsActions.js
```diff
@@ -41,12 +41,15 @@ import {
...
+  CLAIM_REWARD,
 } from 'constants/referralsConstants';
 
+// components
+import Toast from 'components/Toast';
+
 // services
 import { logEvent, getUserReferralLink } from 'services/branchIo';
-import { saveDbAction } from './dbActions';
+import { saveDbAction } from 'actions/dbActions';
```
poocart, line 52:
```suggestion

// actions
import { saveDbAction } from 'actions/dbActions';
```

PR #1825, opened 2020-04-13
feature/switch_to_sw_method_update

vyvil:
This PR allows to switch between wallets without a need to provide biometrics or enter pin code.


src/actions/accountsActions.js
```diff
@@ -238,27 +243,53 @@ export const setActiveAccountAction = (accountId: string) => {
...
-export const switchAccountAction = (accountId: string, privateKey?: string) => {
+/* eslint-disable */
+export const getPkAndInitSWSdkAction = (switchToAcc?: string) => {
+  return async (dispatch: Dispatch) => {
+    const currentScreenAndProps = getNavigationPathAndParamsState();
+    if (!currentScreenAndProps) return;
+    const { path, params } = currentScreenAndProps;
+    const returnToCurrentScreenAction = getActionForPathAndParams(path, params);
+    navigate(PIN_CODE, {
+      customUnlockAction: async (privateKey: string) => {
```
grilix, line 254: Isn't it a bit weird to pass a function to the `navigate`? :thinking:

src/screens/PinCodeUnlock/PinCodeUnlock.js
```diff
@@ -198,9 +222,12 @@ class PinCodeUnlock extends React.Component<Props, State> {
     }
   };
 
-  handlePinSubmit = (pin: string) => {
+  handlePinSubmit = async (pin: string) => {
```
poocart, line 225:
```suggestion
  handlePinSubmit = (pin: string) => {
```

src/actions/accountsActions.js
poocart: Agree, does look weird, but cannot think of any other pragmatic way. Other approach would be to play with this in reducer and putting a flag that this action should be exectured?

src/actions/accountsActions.js
```diff
@@ -269,23 +300,13 @@ export const initOnLoginSmartWalletAccountAction = (privateKey: string) => {
...
     const activeAccountType = getActiveAccountType(accounts);
 
     if (activeAccountType !== ACCOUNT_TYPES.SMART_WALLET) {
-      if ([
-        SMART_WALLET_UPGRADE_STATUSES.TRANSFERRING_ASSETS,
-        SMART_WALLET_UPGRADE_STATUSES.DEPLOYING,
-      ].includes(upgradeStatus)) {
-        await dispatch(initSmartWalletSdkAction(privateKey));
-      }
+      await dispatch(initSmartWalletSdkAction(privateKey));
```
poocart, line 309: I guess this will now force app to switch to Smart Wallet even if user left the app in legacy wallet?

src/actions/authActions.js
```diff
@@ -102,6 +102,16 @@ export const updateFcmTokenAction = (walletId: string) => {
...
 
+export const proceedWithPkAction = (pin: string, action: (pk: string) => void) => {
+  return async (dispatch: Dispatch) => {
+    const { wallet: encryptedWallet } = await storage.get('wallet');
+    const saltedPin = await getSaltedPin(pin, dispatch);
+    const wallet = await decryptWallet(encryptedWallet, saltedPin);
+    const pk = get(wallet, 'signingKey.privateKey');
+    action(pk);
+  };
+};
```
poocart, line 113: Shouldn't this retrieve not from storage, but from keychain (if it's not saved with biometric flag)?

src/screens/PinCodeUnlock/PinCodeUnlock.js
```diff
@@ -115,7 +122,21 @@ class PinCodeUnlock extends React.Component<Props, State> {
...
+    if (customUnlockAction && typeof customUnlockAction === 'function') {
+      if (privateKey) {
+        customUnlockAction(privateKey);
+      } else if (pin) {
+        proceedWithPk(pin, customUnlockAction);
+      }
+    } else {
+      defaultAction();
+    }
+  };
```
jdnmr, line 139: Can we move this logic to `authActions`? This screen has been crazy enough...
Also see my prev comment about `customUnlockAction`. Ideally, we should just pass a boolean value to `handleUnlockAction` and handle all the logic outside of this screen.

src/actions/authActions.js
jdnmr, line 113: I think this logic is too complex. This should be basically a helper function that returns pk basing on pin (something like `getPrivateKeyFromPin = (pin: string) => { ...`). Then the `action` should be called outside of this function.

src/screens/PinCodeUnlock/PinCodeUnlock.js
```diff
@@ -115,7 +122,21 @@ class PinCodeUnlock extends React.Component<Props, State> {
...
+
+  handleUnlockAction = ({ pin, privateKey, defaultAction }: HandleUnlockActionProps) => {
+    const { navigation, proceedWithPk } = this.props;
+    const customUnlockAction = navigation.getParam('customUnlockAction');
+    if (customUnlockAction && typeof customUnlockAction === 'function') {
+      if (privateKey) {
+        customUnlockAction(privateKey);
+      } else if (pin) {
+        proceedWithPk(pin, customUnlockAction);
```
jdnmr, line 134: See my previous comment about `proceedWithPk`. This should look something like
```
const pk = getPrivateKeyFromPin(pin);
fooAction(pk);
```

src/actions/accountsActions.js
vyvil: True, I've passed two variables now - one to check if PinUnlock screen action should open app or init Smart Wallet SDK and second one to check if it needed to switch to an account by doing so.

src/screens/PinCodeUnlock/PinCodeUnlock.js
vyvil: It should be async now - waiting on this.handleUnlockAction to be finished.

src/actions/accountsActions.js
vyvil: It does not. This method only initialises sdk and sets it as init on redux store.
After doing so we return (so connectSmartWalletAccountAction is not called if previously used wallet wasn't smart wallet).

src/actions/authActions.js
vyvil: It's the only place we know for sure that wallet exists, so I assumed it's the safest way to go.
And faster than waiting it to be fetched from keychain. No?

src/screens/PinCodeUnlock/PinCodeUnlock.js
vyvil: Moved :)

src/actions/authActions.js
vyvil: Agree - moved this to utilities

src/screens/Accounts/Accounts.js
```diff
@@ -518,14 +481,16 @@ class AccountsScreen extends React.Component<Props, State> {
 
     const accountsList = [...walletsInList, ...networksToShow];
 
+    const showLoader = changingAccount || isChanging;
```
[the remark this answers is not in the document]
vyvil: Oh, I wish. But if I remove it, is seems like it lagging - it still waits for wallet to to be switched to navigate to wallet view. And if we wouldn't wait for it, user would be returned to wallet view to early and would see previous wallet for a brief moment before it changes to Smart Wallet.

src/actions/smartWalletActions.js
```diff
@@ -1678,3 +1679,15 @@ export const setSmartWalletEnsNameAction = (username: string) => {
...
+
+export const initSmartWalletSdkWithPrivateKeyOrPinAction = ({ privateKey: _privateKey, pin }: InitSmartWalletProps) => {
+  return async (dispatch: Dispatch) => {
+    let privateKey = _privateKey;
+    if (!_privateKey && pin) {
+      privateKey = await getPrivateKeyFromPin(pin, dispatch);
+    } else {
+      return;
+    }
+    await dispatch(initSmartWalletSdkAction(privateKey));
```
jdnmr, line 1691: I don't get this part. So if privateKey is provided in the function argument, we don't want to dispatch `initSmartWalletSdkAction` and instead we just return null? Are you sure you want to keep this?:
```
else {
  return;
}
```
In this scenario `dispatch(initSmartWalletSdkAction(privateKey));` will only be invoked if the user authenticates with pin, and not if they authenticate with privateKey. Is this intended?

vyvil: Ahh, darn it, must have overlooked something. Thanks! Will fix it

PR #1831, opened 2020-04-14
Feature/people empty states [PIL-459]

vyvil:
This PR adds Invite Friends banner to People screen (as empty state - when user does not have contacts yet or search provides 0 results) if referral feature is enabled.

### DIFF
--- a/src/screens/People/People.js
+++ b/src/screens/People/People.js
@@ -38,6 +38,7 @@ import {
 } from 'actions/contactsActions';
 import { fetchInviteNotificationsAction } from 'actions/invitationsActions';
 import { logScreenViewAction } from 'actions/analyticsActions';
+import { goToInvitationFlowAction } from 'actions/referralsActions';
 
 // components
 import Icon from 'components/Icon';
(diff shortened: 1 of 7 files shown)


src/screens/People/People.js
```diff
@@ -399,7 +373,7 @@ class PeopleScreen extends React.Component<Props, State> {
     }, 1000);
   };
 
-  handleReferralBannerPress = () => {
+  handleInvitePress = () => {
     const { navigation, user } = this.props;
```
graszka22, line 377: one nitpick: we have this code already in 3 different places and there is a high possibility we'll need it in another places, is there any possibility to move it to utils? or at least to the InviteBanner component?

src/screens/People/People.js
```diff
@@ -409,6 +383,31 @@ class PeopleScreen extends React.Component<Props, State> {
...
+                bodyText={esBody}
+              />
+            </EmptyStateWrapper>
+          )
+          : (
+            <InviteBanner
+              title={inviteTitle}
+              onInvitePress={this.handleInvitePress}
+            />
+          )
```
poocart, line 405: idk I would move this ternary handling before return and add the result var here instead, but as this piece is temporary (referrals going to be enabled anyway) – not sure if it makes sense 🤔

vyvil: Yeah, didn't want to repeat Wrapper twice :)) and yeah, this flag will be removed eventually so there will be just one type of empty state :)

src/screens/People/People.js
vyvil: Hm, this could be made into an action. I'll take a look :)

vyvil: Done :)

PR #1835, opened 2020-04-15
Feature/collapsible badges section [PIL-356]

vyvil:
This PR adds a collapsible section for badges on Home screen


src/screens/Home/Home.js
```diff
@@ -454,27 +448,31 @@ class HomeScreen extends React.Component<Props, State> {
...
+                          <EmptyStateWrapper>
+                            <EmptyStateParagraph
+                              title="No badges"
+                              bodyText="You do not have badges yet"
+                            />
+                          </EmptyStateWrapper>
+                        )}
+                      />
+                    }
+                    onPress={() => { this.setState({ isBadgesCollapsed: !isBadgesCollapsed }); }}
```
jsidorenko, line 473: Maybe we should store that setting in storage?

vyvil: Good call - will update. Thanks :)

PR #1837, opened 2020-04-15
Add checks for referral flow

grilix:
Make sure the user verified the correct information before triggering
the reward claims, by storing the invited information on the reducer.

This will enable us to:
- Allow the user to change their email/phone and still trigger the
  reward after using an incorrect value.
- In the future, tell/hint the user about the correct value that they
  should use for obtaining the reward.

This also moves the referral data from the storage into redux persist.


src/actions/referralsActions.js
```diff
@@ -147,63 +152,62 @@ export const sendReferralInvitationsAction = (invitationContacts: ReferralContac
...
+        inviteErrorAction(dispatch, errorMessage);
+        return;
+      }
+
+      let updatedInvitationCount = count + invitationContacts.length;
+      const currentDate = formatDate(new Date(), 'YYYY-MM-DD');
+      if (date !== currentDate) {
+        updatedInvitationCount = invitationContacts.length;
       }
+      inviteSentAction(dispatch, {
```
poocart, line 185: This looks a bit weird. All actions have dispatcher within them, idk if this action can be built the way others are?

src/actions/referralsActions.js
```diff
@@ -147,63 +152,62 @@ export const sendReferralInvitationsAction = (invitationContacts: ReferralContac
...
 };
 
 export const startReferralsListenerAction = () => {
   return (dispatch: Dispatch) => {
     if (branchIoSubscription) return;
+
     branchIoSubscription = branch.subscribe(({ error, params }) => {
       if (!isEmpty(error)) return;
+
       console.log('params: ', params);
```
poocart, line 200:
```suggestion
```

PR #1838, opened 2020-04-15
PIL-485 Event cards update

graszka22:


src/components/EventDetails/EventDetails.js
```diff
@@ -18,862 +18,1089 @@
...
-            value={transactionNote}
-          />
-          }
-        </EventBody>
-      );
+    if (isReceived) {
+      eventData = {
+        ...eventData,
+        actionTitle: 'Received',
+        actionSubtitle: `from ${usernameOrAddress}`,
```
vyvil, line 767: Based on designs, this line should start with "Collectible".
The one in L780 as well :)

src/components/EventDetails/EventDetails.js
```diff
@@ -18,862 +18,1089 @@
...
-      lookupAddress(relatedAddress);
+    if (gasUsed) {
+      const fee = gasUsed && gasPrice ? Math.round(gasUsed * gasPrice) : 0;
+      formattedFee = parseFloat(utils.formatEther(fee.toString()));
+    } else {
+      formattedFee = parseFloat(formatUnits(btcFee, assetDecimals));
     }
+    const rate = getRate(rates, asset, fiatCurrency);
+    const formattedFiatValue = formatFiat(formattedFee * rate, fiatCurrency);
+    const feeLabel = `Fee ${formattedFee} ${asset} (${formattedFiatValue})`;
```
vyvil, line 269: This is not right - fee (at least for now) is in ETH, not in the asset that was sent.
(And the way its fiat value is calculated is wrong - it should take ETH as an asset to get proper rate)

src/components/EventDetails/EventDetails.js
```diff
@@ -18,862 +18,1089 @@
...
+            },
+            {
+              title: 'Refer friends',
+              onPress: this.referFriends,
+              squarePrimary: true,
+            },
+          ],
+        };
+      case PAYMENT_NETWORK_ACCOUNT_TOPUP:
+        if (activeBlockchainNetwork === BLOCKCHAIN_NETWORK_TYPES.PILLAR_NETWORK) {
```
vyvil, line 530: Hm, I feel like this is not fair to the user to show this info only if PPN is active. We could change active network if user pick any of those actions (just like we do with other events).

Now, options (buttons) availability is another question. Send availability should be based on PPN's balance and "Top up more", on PLR balance in Smart Wallet.

src/components/EventDetails/EventDetails.js
```diff
@@ -18,862 +18,1089 @@
...
+          buttons: [
+            {
+              title: 'View on the blockchain',
+              onPress: this.viewOnTheBlockchain,
+              secondary: true,
+            },
+          ],
+        };
+      case PAYMENT_NETWORK_ACCOUNT_WITHDRAWAL:
+        return null;
```
vyvil, line 567: It means that there is no modal for withdrawal?

src/components/EventDetails/EventDetails.js
```diff
@@ -18,862 +18,1089 @@
...
+      </React.Fragment>
     );
   }
 }
 
 const mapStateToProps = ({
+  rates: { data: rates },
+  appSettings: { data: { baseFiatCurrency } },
   contacts: { data: contacts, contactsSmartAddresses: { addresses: contactsSmartAddresses } },
-  txNotes: { data: txNotes },
```
vyvil, line 849: Why we are no longer showing transaction notes?

src/components/EventDetails/EventDetails.js
```diff
@@ -18,862 +18,1089 @@
...
+      case WALLET_CREATE_EVENT:
+        return this.getWalletCreatedEventData(event);
+      case PPN_INIT_EVENT:
+        return {
+          name: 'Pillar Network',
+          itemImageSource: PPNIcon,
+          actionTitle: 'Created',
+          buttons: [
+            {
+              title: 'Activate',
```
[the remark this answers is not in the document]
graszka22: Dmitry told me to show [this](https://zpl.io/brZmmB3) modal instead.

src/components/EventDetails/EventDetails.js
```diff
@@ -18,862 +18,1089 @@
...
-              block
-              title="View on the blockchain"
-              primaryInverted
-              onPress={() => viewTransactionOnBlockchain(hash, asset)}
-            />
-          </ButtonsWrapper>
-        );
-      }
-      return null;
+  getSocialEventData = (event: Object): ?EventData => {
```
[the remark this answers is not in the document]
graszka22: There is no design for TYPE_SENT, so I assume we don't show any modal.

src/components/EventDetails/EventDetails.js
```diff
@@ -18,862 +18,1089 @@
...
+    switch (event.eventTitle) {
+      case 'Wallet created':
+        return {
+          name: 'Key wallet',
+          itemImageSource: keyWalletIcon,
+          actionTitle: 'Created',
+          buttons: keyWalletButtons,
+        };
+      case 'Smart wallet created':
+        if (this.props.isSmartWalletActivated) return null;
```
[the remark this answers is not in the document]
graszka22: In Activity Feed it's like this: if we're on PillarNetwork, we display that it's a top up with Pillar Network. If we're on SW, we display it as a transaction from SW to PN. So I made it similar now: if we're on PN, I show this modal, if we're on SW, I show the usual transaction modal.

src/components/EventDetails/EventDetails.js
graszka22: In Activity Feed it's like this: if we're on PillarNetwork, we display that it's a top up with Pillar Network. If we're on SW, we display it as a transaction from SW to PN. So I made it similar now: if we're on PN, I show this modal, if we're on SW, I show the usual transaction modal.

src/components/EventDetails/EventDetails.js
graszka22: There is no design for withdrawal modal.

src/components/EventDetails/EventDetails.js
graszka22: There is no place for transaction notes on design.

graszka22:
@vyvil I'll ask Dmitry about the missing modals, tx notes and edge cases just to be sure and update the PR if needed

graszka22:
Ok, after talking with Dmitry: 
- I added a withdraw event modal
-`TYPE_SENT` modal is not required, this event is temporary
- tx notes will be added later
-  https://zpl.io/2vpm5GP if SW is activated we don't show activate button, and top up button just shows receive modal - that way it doesn't matter if user has any KW assets
- https://zpl.io/a3Bx04l if PN is activated, we don't show activate button

PR #1839, opened 2020-04-16
Fix/list item with image on dark mode [PIL-507]

vyvil:
This PR fixes ListItemWithImage component so items with icons would have correct background color.
It also removes background color from ProfileImage if item has url or displays fallback, so there won't be any pixelated "border" visible.


src/components/ListItem/ListItemWithImage.js
```diff
@@ -184,12 +184,13 @@ const IconCircle = styled.View`
...
-  background-color: ${props => props.backgroundColor || themedColors.card};
+  background-color: ${props => props.backgroundColor || themedColors.tertiary};
   align-items: center;
   justify-content: center;
   text-align: center;
-  border-color: ${props => props.border ? themedColors.border : 'transparent'};
-  border-width: 1px;
+  ${({ border, theme }) => border &&
+    `border-color: ${theme.colors.border};
+    border-width: 1px;`};
```
poocart, line 193:
```suggestion
  ${({ border, theme }) => border && `
    border-color: ${theme.colors.border};
    border-width: 1px;
  `};
```

PR #1841, opened 2020-04-17
[PIL-460] Show legacy wallet if it has balance

alazarevski:
Show legacy wallet if it has balance


src/utils/accounts.js
```diff
@@ -72,6 +72,14 @@ export const findKeyBasedAccount = (accounts: Accounts): ?Account => {
   return accounts.find(({ type }) => type === ACCOUNT_TYPES.KEY_BASED);
 };
 
+export const hasLegacyAccountBalance = (accounts: Accounts, balances: BalancesStore): boolean => {
+  const account = findKeyBasedAccount(accounts);
+  if (!account || Object.keys(balances[account.id]).length === 0) {
```
jsidorenko, line 77:
```suggestion
  if (!account || isEmpty(balances[account.id])) {
```

src/utils/accounts.js
```diff
@@ -72,6 +72,14 @@ export const findKeyBasedAccount = (accounts: Accounts): ?Account => {
...
+    return false;
+  }
+  const legacyBalances: Balances = balances[account.id];
+  return Object.keys(legacyBalances).some(token => legacyBalances[token].balance !== '0.0');
```
jsidorenko, line 81:
```suggestion
  const accountBalances: Balances = balances[account.id];
  return Object.keys(accountBalances).some(token => getBalance(accountBalances, token) > 0);
```

PR #1844, opened 2020-04-20
PIL-499 Performance improvements in Contacts tab

graszka22:
No nested FlatList and no sticky search bar 🎉 * magic *

- I extracted the Overlay component so that I can put the search bar into the list header without z-index issues
- I render just one FlatList for both contacts and search results. And I put the search bar in the header of that FlatList. That way the search bar doesn't re-mount and keeps its state, focus etc. since the components tree stays the same for contacts and search mode.
- And I added a bit of minor performance fixes, such as `getItemLayout`.
- But I doubt this will gain us any massive performance boost. The number of re-renders is about two times smaller with one hundred of contacts on Android (about 200 renders). If the problem persists I'd search the cause somewhere else.


src/screens/People/People.js
```diff
@@ -151,6 +195,10 @@ type Props = {
   theme: Theme,
   referralsFeatureEnabled: boolean,
   goToInvitationFlow: () => void,
+  sendInvitation: Function,
+  acceptInvitation: Function,
+  cancelInvitation: Function,
+  rejectInvitation: Function,
```
jdnmr, line 201: Can we have something more descriptive than `Function` in these? `(user: ApiUser) => void` / `(invitation: Object) => void`.

src/screens/People/People.js
```diff
@@ -396,104 +471,216 @@ class PeopleScreen extends React.Component<Props, State> {
...
+        scrollEventThrottle={16}
+        keyboardShouldPersistTaps="never"
+        contentContainerStyle={{
+          flexGrow: 1,
+        }}
+        extraData={chats}
+        refreshControl={
+          <RefreshControl
+            refreshing={false}
+            onRefresh={() => { fetchInviteNotifications(); }}
```
jdnmr, line 635:
```suggestion
            onRefresh={fetchInviteNotifications}
```

src/screens/People/People.js
```diff
@@ -574,6 +749,10 @@ const mapDispatchToProps = (dispatch: Dispatch): $Shape<Props> => ({
   blockContact: (contactId: string, block: boolean) => dispatch(blockContactAction(contactId, block)),
   logScreenView: (view: string, screen: string) => dispatch(logScreenViewAction(view, screen)),
   goToInvitationFlow: () => dispatch(goToInvitationFlowAction()),
+  sendInvitation: (user) => dispatch(sendInvitationAction(user)),
+  acceptInvitation: (user) => dispatch(acceptInvitationAction(user)),
+  cancelInvitation: (user) => dispatch(cancelInvitationAction(user)),
+  rejectInvitation: (user) => dispatch(rejectInvitationAction(user)),
```
jdnmr, line 755: These 3 actions take `invitation` as an argument, not `user`

PR #1845, opened 2020-04-20
PIL-466 Feature/SW activation view

jdnmr:


src/screens/Assets/WalletActivation.js
```diff
@@ -0,0 +1,120 @@
...
+        <Title>{title}</Title>
+        <Text>{text}</Text>
+        <ButtonsWrapper>
+          <Button
+            height={48}
+            title="Smart Wallet FAQ"
+            onPress={this.handleFaq}
+            textStyle={fontStyles.medium}
+          />
+          <View style={{ height: 4 }} />
```
graszka22, line 90: use `Spacing` component from`components/Layout`?

src/screens/Assets/WalletActivation.js
```diff
@@ -0,0 +1,120 @@
...
+  ${fontStyles.medium};
+  margin: 35px 0;
+`;
+
+const ButtonsWrapper = styled.View`
+  width: 100%;
+  margin-bottom: 32px;
+`;
+
+const title = 'Smart Wallet is being activated';
```
graszka22, line 48: why is the text in constants? throughout the whole app we use text in components without constants

PR #1847, opened 2020-04-21
Feature/ppn insights [PIL-511]

vyvil:
This PR adds new Insights to PPN screen based on Smart Wallet status and PLR amount in PPN and in Smart Wallet.

It also moves Smart Wallet activation modal to a separate component so it could be reused in PPN screen.


src/components/SWActivationCard/SWActivationCard.js
```diff
@@ -370,7 +74,9 @@ class SWActivationCard extends React.Component<Props, State> {
       smartWalletState,
       onButtonPress,
```
jdnmr, line 75: As far as I've noticed, we're never passing `onButtonPress` to this component anymore, so I guess the prop can be deleted?

PR #1848, opened 2020-04-21
PIL-467 Feature/SW activated screen

jdnmr:


src/actions/smartWalletActions.js
```diff
@@ -770,12 +771,7 @@ export const onSmartWalletSdkEventAction = (event: Object) => {
       const deployedAccountState = get(sdkConstants, 'AccountStates.Deployed', '');
       if (newAccountState === deployedAccountState && accountState !== deployedAccountState) {
         dispatch(setSmartWalletUpgradeStatusAction(SMART_WALLET_UPGRADE_STATUSES.DEPLOYMENT_COMPLETE));
-        Toast.show({
-          message: 'Your Smart Wallet has been deployed',
-          type: 'success',
-          title: 'Success',
-          autoClose: true,
-        });
```
jdnmr, line 778: Deleted this Toast because it doesn't make much sense now that we have a whole screen informing about the same.

src/screens/WalletActivated/WalletActivated.js
```diff
@@ -0,0 +1,93 @@
...
+  padding: 0 30px;
+`;
+
+const imageLight = require('assets/images/swActivatedLight.png');
+const imageDark = require('assets/images/swActivatedLight.png');
+
+
+const WalletActivated = (props: Props) => {
+  const isDarkTheme = props.theme === DARK_THEME;
+  const imgSource = isDarkTheme ? imageDark : imageLight;
```
vyvil, line 59: We have a method to get img based on theme. It's in `utils/images.js`. Would be good to use it here as well :)

src/screens/WalletActivated/WalletActivated.js
```diff
@@ -0,0 +1,93 @@
...
+        <Title>Your Smart Wallet is now activated</Title>
+        <Image source={imgSource} style={{ height: 137, width: '100%' }} resizeMode="stretch" />
+        <Text>
+          You can now access your new Smart Wallet and the new Pillar Payment Network from the upper right
+          hand side of the Assets screen
+        </Text>
+        <ButtonWrapper>
+          <Button
+            title="Go to Smart Wallet"
+            onPress={() => props.navigation.navigate(ASSETS)}
```
vyvil, line 82: Might be an edge case, but what if I wasn't on Smart Wallet when it was deployed? It would open Assets screen with the last active wallet.

jdnmr: Great catch! Handled now

PR #1853, opened 2020-04-22
PIL-532 Fix/Sending assets on iOS

jdnmr:
Also refactored `components/TextInput` to make the component shorter


src/screens/SendToken/SendTokenContacts.js
```diff
@@ -465,6 +470,28 @@ class SendTokenContacts extends React.Component<Props, State> {
...
+              regularText
+              textStyle={fontStyles.medium}
+              disabled={submitDisabled}
+              title="Next"
+              onPress={this.handleFormSubmit}
+            />
+          </ButtonWrapper>
+        )}
+      </>
+    );
```
jsidorenko, line 492:
```suggestion
    if (!isSearchQueryProvided) return null;
    return (
      <ButtonWrapper>
        <Button
          height={48}
          regularText
          textStyle={fontStyles.medium}
          disabled={submitDisabled}
          title="Next"
          onPress={this.handleFormSubmit}
        />
      </ButtonWrapper>
    );
```

jdnmr: I tried this at first and this doesn't do the job. But it does when you replace `null` with `<></>`, so committing that. RN never fails to surprise me...

PR #1858, opened 2020-04-23
PIL-472 Return instance of ether.Wallet in checkAuthAction

graszka22:
This is related to clovers network WC sign in this [bug](https://linear.app/issue/PIL-472/cloversnetwork-wc-request-cards-freeze). I couldn't reproduce the freezing but certainly I couldn't sign in to clovers.network.
Approving of the WC request calls `signPersonalMessage` which in turn calls `wallet.connect`. The `wallet` variable should be an instance of `ethers.Wallet` but for some reason it was just a plain object and the `connect` function was undefined. I debugged it and in `checkAuthAction` we return just a plain object if we ask for a private key. So I changed it into `ethers.Wallet` (it should have the `privateKey` and `address` fields set up correctly) so it signs in correctly now and we can search for clovers again! 🍀 🍀 🍀


src/actions/authActions.js
```diff
@@ -346,8 +347,7 @@ export const checkAuthAction = (
         const saltedPin = await getSaltedPin(pin, dispatch);
         wallet = await decryptWallet(encryptedWallet, saltedPin, options);
       } else if (privateKey) {
-        const walletAddress = normalizeWalletAddress(encryptedWallet.address);
-        wallet = { ...encryptedWallet, privateKey, address: walletAddress };
+        wallet = new ethers.Wallet(privateKey);
```
jsidorenko, line 350: 1) why we didn't change this before? :D 
2) `wallet = await decryptWallet(encryptedWallet, saltedPin, options);` additionally connects to the provider, while your approach will simply return the wallet instance. Do you think it's enough?
As I remember we're checking for the provider in other places, so it could be ok

graszka22: 1. No idea 🤷‍♀️ I just found out that it was introduced a few months ago in this [PR](https://github.com/pillarwallet/pillarwallet/pull/1366) by @poocart. Maybe back then it was making sense and over the time the usage of these methods changed.
2. Yes, we're always connecting to provider first.

jsidorenko: 1) Right, before it was enough to have this version `wallet = { ...encryptedWallet, privateKey, address: walletAddress };` since we just needed to actually know the privateKey and the wallet's address in the code below. Things have changed since then :)
2) we should fix the way we connect to the provider, it could be done in a separate PR. I've checked all the occurrences of that provider's connection and only 1 out of 7 is correct.
The problem is that `wallet.connect(provider)` will return a new instance which we don't store anywhere

PR #1860, opened 2020-04-24
Feature/referral already invited [PIL-248]

vyvil:
This PR uses API endpoint to fetch already sent referral invites and also adds some UI improvements on referral screen


src/components/Note/Note.js
```diff
@@ -22,26 +22,32 @@ import { Text } from 'react-native';
...
 import { BaseText } from 'components/Typography';
-import { fontStyles } from 'utils/variables';
+import { fontStyles, spacing } from 'utils/variables';
 import { themedColors } from 'utils/themes';
 
 type Props = {
   note: string | React.Node,
   emoji?: string,
   containerStyle?: Object,
+  childrenOnRight?: React.Node,
```
jdnmr, line 32: There aren't any other `children` for now, so I guess you can just name this `children` :)

src/services/api.js
```diff
@@ -316,6 +316,13 @@ class SDKWrapper {
       });
   }
 
+  getSentReferralInvites(walletId: string) {
+    return Promise.resolve()
+      .then(() => this.pillarWalletSdk.referral.list({ walletId }))
+      .then(({ data }) => data.data)
+      .catch(() => ([]));
```
jsidorenko, line 323:
```suggestion
      .catch(() => []);
```

src/screens/ReferFriends/MissingInfoNote.js
```diff
@@ -56,19 +70,22 @@ const MissingInfoNote = (props: Props) => {
...
-      note={
-        <React.Fragment>
-          <BaseText>
-            {`To show your ${missingType} contacts, please add and verify your ${missingType}. `}
-          </BaseText>
-          <StyledTextLink onPress={onPressAdd}>Add</StyledTextLink>
-        </React.Fragment>
-      }
-    />
+    <TouchableOpacity onPress={onPressAdd}>
```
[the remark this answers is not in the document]
vyvil: Yeah, me neither, so I've added external wrapper for now :)
That would be easy to change if it becomes a "trend" to add touchable actions to notes :)

src/components/Note/Note.js
vyvil: Yeah, was thinking that we might wanna add children in footer, so I've named this prop according to position. Since those are the first ones I'll rename them to "children".

PR #1863, opened 2020-04-24
[PIL-542] Feature/Allow users to copy address when pressed

alazarevski:
Allow users to copy address when pressed


src/screens/Asset/ReceiveModal.js
```diff
@@ -124,7 +124,9 @@ export default class ReceiveModal extends React.Component<Props, *> {
         <ContentWrapper forceInset={{ top: 'never', bottom: 'always' }}>
           <WarningBanner rounded small />
           <QRCodeWrapper>
-            <WalletAddress>{address}</WalletAddress>
+            <WalletAddress onPress={() => Clipboard.setString(address)}>
```
vyvil, line 127: It's not clear when text has been copied - no info is shown to the user. Maybe we can add a toast?
(And it would also be nice to have a visual touch representation here)

alazarevski: i can just add a copy symbol but it might jump over a line

alazarevski: Made font smaller and added clipboard symbol

alazarevski: Actually changed the symbol and reverted the address. Since users are expected to tap to select the string, they'll be notified that the address has been copied successfully and they'll know no more effort is needed to get the address :)

PR #1864, opened 2020-04-24
Fix/referral ui fix [PIL-548]

vyvil:
This PR adds a modal to notify that invitations have been sent and that referral reward is pending.
It also fixes phone contact's list empty state.
(the other issue on the the ticket was fixed before, in my previous PR).


src/screens/ReferFriends/ReferralSent.js
```diff
@@ -0,0 +1,107 @@
...
+          statusbarColor={{
+            [LIGHT_THEME]: DARK_CONTENT,
+          }}
+        >
+          <Wrapper flex={1} center>
+            <RewardBadge source={rewardBadge} />
+            <Title>Your reward is on the way</Title>
+            <Paragraph>
+              Thank you for spreading the word about Pillar.
+              You will receive 25 PLR for each friend installed the app with your referral link.
```
jsidorenko, line 83: Maybe we should extract those 25 PLR into some config?

src/actions/referralsActions.js
```diff
@@ -191,6 +188,10 @@ export const sendReferralInvitationsAction = (invitationContacts: ReferralContac
         sentInvitationsCount: { count: updatedInvitationCount, date: currentDate },
       }));
     }));
+    if (unsentInvitations.length < invitations.length) {
+      navigate(REFERRAL_SENT);
+    }
+    dispatch(inviteErrorAction(errorMessage, unsentInvitations.length === invitations.length));
```
graszka22, line 194: You're dispatching an error even though all invitations are sent?

src/screens/ReferFriends/ReferralSent.js
```diff
@@ -0,0 +1,107 @@
...
+`;
+
+const rewardBadge = require('assets/images/referralBadge.png');
+
+class ReferralSent extends React.PureComponent<Props> {
+  render() {
+    const { navigation } = this.props;
+    return (
+      <ConfettiBackground>
+        <ContainerWithHeader
```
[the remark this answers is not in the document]
vyvil: I think this feels not that right because of component name. In the end - it's a container component that holds some useful methods. In this case, yes - the method for status bar color. And safe area view. 
I'd rather leave it like this instead of copying the code for status bar management elsewhere.
I could make some adjustments to Container so it would not render header if no header props are passed.

src/screens/ReferFriends/ReferralSent.js
vyvil: Good point - fixed :)

src/actions/referralsActions.js
vyvil: Yeah, missed that one. Thanks! Fixed :)

src/components/ConfettiBackground/ConfettiBackground.js
```diff
@@ -0,0 +1,75 @@
...
+
+import { themedColors } from 'utils/themes';
+import { spacing } from 'utils/variables';
+
+
+type Props = {
+  children: React.Node,
+};
+
+const { width: screenWidth } = Dimensions.get('window');
```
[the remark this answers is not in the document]
vyvil: Added :)

PR #1866, opened 2020-04-27
Replace errors with warnings for local builds [PIL-526]

alazarevski:
Replace console.error with console.warn (Yellow screen) for various non-fatal errors.
Also as PIL-526 couldn't be reproduced with the latest develop branch changes, would suggest errors to be handled with a warning instead of crashing the app


jsidorenko:
What do you think about replacing all those ^^ `console.warn/error` with Sentry reporting?

alazarevski:
> What do you think about replacing all those ^^ `console.warn/error` with Sentry reporting?

I think these were added for developers to see, since they are pretty local build and developer related.

jsidorenko:
Not really, I def want to be aware if some of those errors happened to prod users. Especially regarding the smart wallet

jsidorenko:
Since we have `reportLog()` function, I would suggest to modify it by adding: `if (__DEV__) { console.error(); return; }`. In that case, locally we would see the red screen and in prod there will be a Sentry report sent

PR #1869, opened 2020-04-27
PIL-531 PPN history UI updates

graszka22:


src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -316,6 +318,11 @@ class ActivityFeed extends React.Component<Props, State> {
             activeTab={activeTab || firstTab}
           />
         }
+        {formattedFeedData.length === 2 && (
```
vyvil, line 321: Hm, why we show Empty State message if feed is not empty (has 2 list items in it)?

graszka22: It has always the header and the tabs. Maybe I can add a comment to make it clearer?

graszka22:
@vyvil empty state fixed, no more weird ifs

src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -123,6 +122,7 @@ type Props = {
   tabsComponent?: React.Node,
   headerComponent?: React.Node,
   flatListProps?: FlatList,
+  isPPNView: boolean,
```
jdnmr, line 125:
```suggestion
  isPPNView?: boolean,
```

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -114,6 +114,7 @@ type Props = {
   isSmartWalletActivated: boolean,
   assetDecimals: number,
   bitcoinAddresses: BitcoinAddress[],
+  isPPNView: boolean,
```
jdnmr, line 117:
```suggestion
  isPPNView?: boolean,
```

PR #1872, opened 2020-04-27
PIL-293 Update People UI

graszka22:


src/screens/People/People.js
```diff
@@ -681,15 +629,35 @@ class PeopleScreen extends React.Component<Props, State> {
...
+
     return (
       <ContainerWithHeader
-        headerProps={{ noBack: true, leftItems: [{ title: 'People' }] }}
+        headerProps={{
+          noBack: true,
+          leftItems: [{ title: 'People' }],
+          rightItems: [{
+            custom: (
+              <IconButton
```
vyvil, line 649: This should be wrapped with Referrals feature flag, I suppose.

src/screens/People/People.js
```diff
@@ -681,15 +629,35 @@ class PeopleScreen extends React.Component<Props, State> {
...
+                icon="present"
+                color={colors.positive}
+                iconText="Invite friends"
+                style={{ flexDirection: 'row' }}
```
vyvil, line 653: Oh and one more tiny thing: 
seems like horizontal alignment is a tiny bit off - this button should to be 2px or so lower to be horizontally centered with screen's title

src/screens/People/People.js
```diff
@@ -681,15 +629,35 @@ class PeopleScreen extends React.Component<Props, State> {
...
       <ContainerWithHeader
-        headerProps={{ noBack: true, leftItems: [{ title: 'People' }] }}
+        headerProps={{
+          noBack: true,
+          leftItems: [{ title: 'People' }],
+          rightItems: [{
+            custom: (
+              <IconButton
+                icon="present"
+                color={colors.positive}
```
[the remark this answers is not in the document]
graszka22: I don't think it's worth the effort. In case of buttons it makes sense as we have a set of button styles definied in style guide (primary, secondary etc). In case of icon buttons I can't see such styles and probably they won't be reused.

PR #1877, opened 2020-04-28
Ui/username [PIL-523]

vyvil:
This PR adds username next to avatar on Home screen


src/components/UserNameAndImage/UserNameAndImage.js
```diff
@@ -0,0 +1,82 @@
...
+};
+
+
+const UserWrapper = styled.TouchableOpacity`
+  padding: 0 ${spacing.medium}px;
+  flex-direction: row;
+  align-items: center;
+`;
+
+const UserName = styled(BaseText)`
```
graszka22, line 51: it should have medium weight

vyvil: Ahh, busted :)) Thanks. Will adjust

PR #1880, opened 2020-04-28
[PIL-529] Fix/badges imagecache change

alazarevski:
Image caching on badges loads the first badge, but keeps the others to the default image.
Changed the library used to react-native-fast-image which improves speed of image loading
TODO: Review and Change all CachedImage to FastImage components


package.json
```diff
@@ -105,6 +105,7 @@
     "react-native-dotenv": "^0.2.0",
     "react-native-emoji": "^1.6.0",
     "react-native-extra-dimensions-android": "^1.2.5",
+    "react-native-fast-image": "^8.1.5",
```
graszka22, line 108: We're probably using proguard for release builds: https://github.com/DylanVann/react-native-fast-image#are-you-using-proguard

alazarevski: yup 7723c36

src/components/BadgeImage/BadgeImage.js
```diff
@@ -16,13 +16,13 @@ const BadgeImage = ({ data: badge, size = 96 }: Props) => {
...
+                height: size,
+                width: size,
+            }}
       source={{
-              uri: badgeUri,
-              priority: FastImage.priority.normal,
-          }}
+                uri: badgeUri,
+                priority: FastImage.priority.normal,
+            }}
```
graszka22: too much indent :D imho it should go like this:
```js
style={{
    height: size.
    width: size,
}}
source={{
    uri: badgeUri,
    priority: FastImage.priority.normal,
}}
```

alazarevski: weird, webstorm ide reformats it to this with eslint -fix

graszka22: vscode too, I feel like eslint has a bug or something's up with configuration, because it doesn't care about indenting objects passed as props, you can make any indent and it will pass :D

PR #1884, opened 2020-04-29
Fix android vibrate permission

graszka22:
To reproduce try to scan the QR code when connecting with wallet connect. You'll get red screen saying that VIBRATE permission is missing. I don't know how it was overlooked :D


jsidorenko:
We have an extra list of permissions those we ask user during registration. Pls check that new permission is here

graszka22:
@jsidorenko yes, it's there, although the rendering of those permissions is broken :p

PR #1886, opened 2020-04-29
Feature/fast image replace

alazarevski:
Reviewing possible inability to control the cache https://github.com/DylanVann/react-native-fast-image/blob/master/docs/how-is-caching-handled.md


jsidorenko:
How the FastImage caches the image and how we could control that?

alazarevski:
> How the FastImage caches the image and how we could control that?

the thing is, we don't, since they cache the image and once full, the cache capacity clears the least accessed keys first

PR #1898, opened 2020-04-30
Fix/show import in forgot pin screen

alazarevski:
The button was invisible and the wording removes private key mentions


src/screens/ForgotPin/ForgotPin.js
```diff
@@ -67,12 +67,9 @@ class ForgotPin extends React.Component<Props, {}> {
         footer={(
           <FooterWrapper>
             <FooterParagraph>
-              It is impossible to restore your wallet without the backup, be careful.
+              It is impossible to restore your wallet without the backup phrase, be careful.
             </FooterParagraph>
             <Button
-              leftIconName="down-arrow"
-              block
-              danger
```
vyvil, line 75: Why are you changing UI?
The issue was Button danger theme. I've fixed it here:
https://github.com/pillarwallet/pillarwallet/pull/1897/commits/7d50b8a59be6fffac4d5f4bf0625655767d6fad4

alazarevski: done 92a767c

PR #1908, opened 2020-05-04
Feature/sw deploying state

vyvil:
This PR adds deployment state to PPN screen and also applies some fixes on deployment process


src/screens/Assets/WalletView.js
```diff
@@ -417,70 +412,71 @@ class WalletView extends React.Component<Props, State> {
...
+          <>
+            <SearchBlock
+              hideSearch={blockAssetsView}
+              searchInputPlaceholder={activeTab === TOKENS ? 'Search asset' : 'Search collectible'}
+              onSearchChange={this.handleSearchChange}
+              wrapperStyle={{
+              paddingHorizontal: spacing.layoutSides,
+              paddingVertical: spacing.mediumLarge,
+              marginBottom: searchMarginBottom,
+            }}
```
jdnmr, line 464: please fix indentation

src/screens/Assets/WalletView.js
```diff
@@ -313,11 +313,6 @@ class WalletView extends React.Component<Props, State> {
     });
   };
 
-  getStickyIndices = () => {
-    const { showDeploySmartWallet } = this.props;
-    return !this.isInSearchAndFocus() && !this.shouldBlockAssetsView() && showDeploySmartWallet ? [2] : [1];
```
jdnmr, line 318: Are you sure it's safe to delete the `? [2] : [1]` and replace it with `[1]` for all cases?

vyvil: Yes, I've wrapped upper components so they are all now considered to be first child ([0]), so no matter what, we can always make second one ([1]) sticky (a group of SearchBar and Tabs)

PR #1910, opened 2020-05-04
Event details modals fixes

graszka22:
- adds back ens sync
- adds back transaction pending sync, so that we show in real time when the transaction gets confirmed

- to reproduce:
  - top up PN
  - go to activity feed
  - a new event appears, click on it
  - it's not pending as it should be


src/components/EventDetails/EventDetails.js
```diff
@@ -243,14 +249,82 @@ const Divider = styled.View`
...
+
   state = {
     isReceiveModalVisible: false,
     SWActivationModalVisible: false,
   };
 
-  isReceived = ({ to: address }: Object): boolean => {
+  componentDidMount() {
+    if (this.props.event.type !== TRANSACTION_EVENT) return;
+    const txInfo = this.findTxInfo() || {};
```
jsidorenko, line 263: How about moving `|| {}` into `findTxInfo()` method?

graszka22: We need to know that we can't find transaction info in order to know whether we should use cachedTxInfo.

PR #1917, opened 2020-05-05
[PIL-519] Fix/exchange buy speedup

alazarevski:
Remove ScrollView so it is faster, workaround the rest of the components to show correctly.


alazarevski:
> Codewise looks good, but I'm not sure if it solves the real performance issue here. When testing on Android the pop up animation lags a bit, like on the video in the issue.

iOS is significantly faster

PR #1926, opened 2020-05-06
fix/relogin oauth refresh

alazarevski:
relogin flow when oauth fails on startup
This will work on startup, but not when the auth tokens expire in-app (which rarely happens and is recoverable by relaunching the app)


poocart:
@alazarevski this will prevent from sending PK from wallet to SDK which will end up as no access tokens at all. Correct me if I'm wrong?

jsidorenko:
Sprinteris

alazarevski:
Testing more in Exchange for the refresh since it's not loading assets on first login, if opened again it works

PR #1937, opened 2020-05-07
fix/show eth gastoken keybased

alazarevski:
Set feeByGasToken false when it's not smartwallet
This will set ETH for KeyBased wallets and fix the bug which may be the cause of TransactionFailed in Exchanges

### DIFF
--- a/src/screens/Exchange/ExchangeConfirm.js
+++ b/src/screens/Exchange/ExchangeConfirm.js
@@ -277,7 +277,6 @@ class ExchangeConfirmScreen extends React.Component<Props, State> {
     if (activeAccount && checkIfSmartWalletAccount(activeAccount)) {
       return this.getSmartWalletTxFeeInWei();
     }
-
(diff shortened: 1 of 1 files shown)


src/screens/Exchange/ExchangeConfirm.js
```diff
@@ -274,9 +274,16 @@ class ExchangeConfirmScreen extends React.Component<Props, State> {
 
   getTxFeeInWei = (txSpeed?: string, gasLimit?: number): BigNumber => {
     const { gasInfo, activeAccount } = this.props;
+    const { feeByGasToken } = this.state;
     if (activeAccount && checkIfSmartWalletAccount(activeAccount)) {
       return this.getSmartWalletTxFeeInWei();
     }
+    if (feeByGasToken) {
+      this.setState({
+        feeByGasToken: false,
```
jsidorenko, line 283: Why not set this in constructor?

PR #1953, opened 2020-05-11
PIL-538 Feature/smooth wallet transitions

jdnmr:


src/screens/Accounts/Accounts.js
```diff
@@ -248,12 +235,18 @@ class AccountsScreen extends React.Component<Props, State> {
...
     } = item;
     return (
       <SettingsItemCarded
+        isSwitching={id === this.state.switchingToWalletId}
         title={title}
         subtitle={balance}
-        onMainPress={isInitialised ? mainAction : initialiseAction}
+        onMainPress={() => {
+          this.setState({ switchingToWalletId: id }, isInitialised ?
+            mainAction || (() => {}) :
```
jsidorenko, line 247:
```suggestion
            mainAction || noop :
```

src/screens/Accounts/Accounts.js
```diff
@@ -248,12 +235,18 @@ class AccountsScreen extends React.Component<Props, State> {
...
+            initialiseAction || (() => {}));
```
jsidorenko, line 248:
```suggestion
            initialiseAction || noop);
```

src/screens/Accounts/Accounts.js
```diff
@@ -180,9 +167,9 @@ class AccountsScreen extends React.Component<Props, State> {
         (!user.isLegacyUser && isActiveKeyWallet) || hasLegacyBalance;
     this.state = {
       showPinModal: false,
-      changingAccount: false,
       isLegacyWalletVisible: forceShowLegacyWallet,
       onPinValidAction: null,
+      switchingToWalletId: '',
```
jsidorenko, line 172: What do you think about making the default value as null?

PR #1955, opened 2020-05-12
Feature/sticky confirm buttons [PIL-547]

vyvil:
This PR adds reusable ReviewAndConfirm layout
and updates Exchange review screen to make confirmation button floating.


src/components/ReviewAndConfirm/ReviewAndConfirm.js
```diff
@@ -0,0 +1,147 @@
...
+
+const WarningMessage = styled(Paragraph)`
+  text-align: center;
+  color: ${themedColors.negative};
+  padding-bottom: ${spacing.small}px;
+`;
+
+const ValueHolder = ({ value, valueArray, isLoading }) => {
+  if (isLoading) return <Spinner width={20} height={20} />;
+  if (valueArray && valueArray.length) return valueArray.map((val) => (<Value key={val}>{val}</Value>));
```
jsidorenko, line 78:
```suggestion
  if (!isEmpty(valueArray)) return valueArray.map(val => <Value key={val}>{val}</Value>);
```

vyvil:
Can you specify Android device, @graszka22?

graszka22:
@vyvil https://www.gsmarena.com/sony_xperia_xa2-8986.php with android 9

graszka22:
it's kinda like it's scrolling to the top itself?

vyvil:
@graszka22,
Fixed TextInput focus on iOS.
Could not replicate this issue on Android, that you've reported;/.
I've attempted to fix it by disabling this Keyboard avoidance scroll wrapper adjustments on Android (as we do have windowSoftInputMode set to adjustResize on Android). Could you, please, take a look if this fixes screen jumping on your side? :)

graszka22:
Nah, it doesn't fix the Android side. I've made a fix myself, since I can reproduce it: https://github.com/pillarwallet/pillarwallet/pull/2015. I think the problem was that `ContainerWithHeader` also has a `ScrollWrapper`. You can merge it to your branch if you want ;)

PR #1956, opened 2020-05-12
[PIL-539] Fix/SW and KW addresses in Receive

alazarevski:
Add address as prop so that the list item is re-rendered when accounts are switched

### DIFF
--- a/src/components/ListItem/ListItemWithImage.js
+++ b/src/components/ListItem/ListItemWithImage.js
@@ -93,6 +93,7 @@ type Props = {
   badge?: string,
   iconBackgroundColor?: string,
   iconBorder?: boolean,
+  address?: string,
 }
 
 type AddonProps = {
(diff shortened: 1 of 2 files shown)


src/components/ListItem/ListItemWithImage.js
```diff
@@ -93,6 +93,7 @@ type Props = {
   badge?: string,
   iconBackgroundColor?: string,
   iconBorder?: boolean,
+  address?: string,
```
jsidorenko, line 96: Can't find any usage of that variable in that file. Why do you need to pass it and where do you use it?

alazarevski: shouldComponentUpdate had logic comparing the props and failing to recognize dofferent accounts

jsidorenko: you're right here.. I'm just afraid someone will remove that property as unused in the future and we will come back to the previous issue

PR #1960, opened 2020-05-13
Feature/referrals reward address [PIL-559]

vyvil:
This PR gets referral reward issuers addresses and changes label of Activity feed item is transaction is received form any of those addresses into 'Referral reward'.

Ir also fixes missing Even modal by allowing to open modal for TRANSACTION_PENDING_EVENT.


src/components/EventDetails/EventDetails.js
```diff
@@ -931,7 +942,9 @@ class EventDetail extends React.Component<Props, State> {
             squarePrimary: true,
           };
 
-          if (isReceived) {
+          if (isReferralRewardTransaction) {
+            eventData.buttons = [];
```
graszka22, line 946:
```suggestion
           buttons = [];
```

vyvil: > We have designs for referrals in activity feed and event detail modals
> https://zpl.io/2jnQ7gQ
> https://zpl.io/bLlPGv7
> but if it's outside of the scope of this PR it's fine as it is 👍

Aahh, wasn't aware of that ;/
Will add designs in other PR then :)

PR #1965, opened 2020-05-13
PIL-452 PIL-572 Svg collectibles

graszka22:
this PR:
- adds support for svg collectibles via `react-native-svg` library
- I needed to change `react-native-image-view` to `react-native-image-zoom-viewer` in order to show svg in full screen preview
- the patch for `react-native-svg` is for the case we have a custom `fontFamily` in svg (the app would crash with `unrecognized font family` error), there is no way as far as I know for setting a default font or overriding it or anything in props.
- in case we have a custom `fontFamily` the text might render incorrectly (for example the position of the text is a bit off etc). But there isn't really anything we can do about it - we can't add every possible font in our app 🤷‍♀️ 
- the icons for collectibles for activity feeds etc are generated somewhere in the backend (so that we have a small `icon` and a full `image` for collectibles). In case of svg collectbles we had  svg `image` and png `icon` so the icons were rendering quite well. But for some collectibles the icons aren't generated correctly (for example there are just empty borders instead of full image). I have no idea how it's generated and whether we can change it, so the second commit adds rendering via svg also for activity feeds, lists etc. (but only for svg images, the png images are rendered normally)
- If you want some collectibles to test just ping me on slack, I can generate any collectible you want :)


src/components/EventDetails/EventDetails.js
```diff
@@ -1128,6 +1137,7 @@ class EventDetail extends React.Component<Props, State> {
     const { theme } = this.props;
     const {
       imageUrl, itemImageSource, profileImage, username, iconName, iconColor, imageBackground, imageBorder,
+      collectibleUrl,
```
poocart, line 1140: This looks weird, maybe we can just separate each as new line?

src/screens/Assets/CollectiblesList.js
```diff
@@ -103,12 +104,14 @@ class CollectiblesList extends React.PureComponent<Props> {
           contentWrapperStyle={{ padding: spacing.medium, alignItems: 'center' }}
           onPress={() => this.handleCardTap(item)}
         >
-          <CachedImage
+          <CollectibleImage
             style={{
               height: 135,
               width: 135,
```
poocart, line 110: This not needed anymore?

src/screens/Collectible/Collectible.js
```diff
@@ -168,25 +140,41 @@ class CollectibleScreen extends React.Component<Props, State> {
...
-      />
+        onModalHide={this.onCloseImageView}
+        fullScreen
+        showHeader
+      >
+        <ImageViewer
+          imageUrls={imageViewImages}
+          renderImage={(props) => {
+            return <CollectibleImage {...props} />;
+          }}
```
poocart, line 171:
```suggestion
          renderImage={(props) => <CollectibleImage {...props} />}
```

PR #1966, opened 2020-05-13
PIL-564 Change numeric text input weight

graszka22:


src/utils/textInput.js
```diff
@@ -53,6 +53,6 @@ export const getLineHeight = (value: string | number, numeric?: boolean) => {
 
 export const getFontFamily = (value: string | number, numeric?: boolean) => {
   if (!(value || value === 0)) return appFont.medium;
```
vyvil, line 55: As we already return medium in upper row, we could also do this:
`if (!(value || value === 0) || numeric) return appFont.medium;`

PR #1970, opened 2020-05-14
[PIL-492] Make amounts pretty (and rounded)

alazarevski:
Pretty rounded amounts for lists, removed duplicated function getFormattedSettleValues


src/components/ListItem/ListItemWithImage.js
```diff
@@ -542,11 +543,13 @@ const Addon = (props: AddonProps) => {
       custom,
       customOnRight,
     } = balance;
+    const decimalPlaces = getDecimalPlaces(token);
```
jsidorenko, line 546: I would rename `getDecimalPlaces` to smth like `getTokenFormattingPrecision` since when I read `getDecimalPlaces(token);` I'm thinking about not formatting, but rather the real amount of decimals those the token has (18 for eth, etc)

PR #1981, opened 2020-05-20
PIL-400: Feature/web recovery portal

poocart:
https://linear.app/pillarproject/issue/PIL-400/implement-recovery-set-up-screens

Includes:
- Flows for Web Recovery portal implementation: setup and recover.
- Smart Wallet account device management.
- Other improvements, fixes.

**Note if testing: make sure you have enough ETH in Smart Wallet account to cover transactions because Web Recovery Portal itself doesn't have feature to cover transaction fee with PLR.**


src/components/EventDetails/EventDetails.js
```diff
@@ -841,6 +843,32 @@ export class EventDetail extends React.Component<Props, State> {
           ],
         };
         break;
+      case SMART_WALLET_ACCOUNT_DEVICE_ADDED:
+        eventData = {
+          name: 'New Smart Wallet account device',
+          actionTitle: 'Added',
+          buttons: [
+            {
+              title: 'View on the blockchain',
```
vyvil, line 852: Just a note to expect changes:
Links to check trx on blockchain will be moved to event time as per #2021

src/components/RecoveryPortalWebView/RecoveryPortalWebView.js
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  return (
+    <ContainerWithHeader
+      headerProps={{
+        centerItems: [{ title: title || 'Recovery Portal' }],
+        customOnBack: handleNavigationBack,
+      }}
+    >
+      <Wrapper style={{ flex: 1 }}>
```
vyvil, line 86: Minor thing:
What if you'd center wrapper's content?
Then there won't be any need to make Spinner's position absolute.
(Now it's top position will be slightly off the center (as you'd need to transform y or add negative top margin to put it the center).

src/components/Typography/Typography.js
```diff
@@ -83,7 +83,7 @@ const getTextStyle = (props) => {
 export const BaseText = styled.Text`
   font-family: ${appFont.regular};
   text-align-vertical: center;
-  color: ${themedColors.text};
+  color: ${({ color }) => color || themedColors.text};
   ${(props) => getTextStyle(props)}
```
vyvil, line 87: Hm, we cover bunch of colors from our color pallet in getTextStyle. One need just to add color name used in theme as a prop. Is that not enough?

src/actions/recoveryPortalActions.js
```diff
@@ -0,0 +1,202 @@
...
+      }
+    }
+  };
+};
+
+export const initRecoveryPortalWalletRecoverAction = () => {
+  return async (dispatch: Dispatch) => {
+    // make sure everything is reset
+    dispatch({ type: RESET_RECOVERY_PORTAL_TEMPORARY_WALLET });
+    await smartWalletService.reset();
```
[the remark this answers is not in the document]
poocart: No, this one is called only before Web Recovery portal is shown and we need temporary wallet for this so we want only to reset what relates to this.

src/components/RecoveryPortalWebView/RecoveryPortalWebView.js
poocart: Makes sense, BUT it would work for the general spinner, but won't work for WebView component spinner as prop, but I'll try to adjust by passing additional value to spinner render method.

poocart: Actually it becomes more complex and requires more code lines (because I have to additionally dismiss centring on wrapper because of WebView component) so I would like to keep it the way it is, hope it makes sense.

src/actions/recoveryPortalActions.js
```diff
@@ -0,0 +1,202 @@
...
+      isImported: true,
+    });
+
+    Toast.show({
+      message: 'Wallet successfully recovered!',
+      type: 'success',
+      title: 'Success',
+    });
+
+    navigateToAppFlow(false);
```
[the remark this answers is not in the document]
poocart: Updated, but for 6th step leaving only creation times as key based is not actually imported.

src/components/EventDetails/EventDetails.js
```diff
@@ -841,6 +843,32 @@ export class EventDetail extends React.Component<Props, State> {
           ],
         };
         break;
+      case SMART_WALLET_ACCOUNT_DEVICE_ADDED:
+        eventData = {
+          name: 'New Smart Wallet account device',
+          actionTitle: 'Added',
```
vyvil, line 849: This action might be pending, so it would be better to handle its title as we handle titles of collectibles' transactions.

src/screens/RecoveryPortal/RecoveryPortalSetupComplete.js
```diff
@@ -0,0 +1,141 @@
...
+
+const Title = styled(MediumText)`
+  ${fontStyles.large};
+  text-align: center;
+  margin-bottom: ${spacing.small}px;
+`;
+
+const skipPrompt = (callback) => Alert.alert(
+  'Are you sure?',
+  'You are going to skip your Private Key backup, but no at all worries – you can still do it later!',
```
vyvil, line 61: "no at all worries" -> "no worries at all" (or "don't worry", or in old english "worry not" :)) )

src/screens/RecoveryPortal/RecoveryPortalWalletRecoveryIntro.js
```diff
@@ -0,0 +1,113 @@
...
+          <Button
+            block
+            title="Proceed"
+            onPress={() => navigation.navigate(RECOVERY_PORTAL_WALLET_RECOVERY)}
+            marginTop={50}
+            marginBottom={spacing.large}
+          />
+          <ButtonText
+            buttonText="Recover Key wallet first"
+            onPress={() => navigation.goBack()}
```
vyvil, line 103: This could lead to Re-import wallet directly, what do you think? :)

src/services/api.js
```diff
@@ -175,15 +184,23 @@ class SDKWrapper {
...
+    if (username) requestPayload = { ...requestPayload, username };
     if (!isEmpty(fcmToken)) requestPayload = { ...requestPayload, fcmToken };
+    if (!isEmpty(recovery)) requestPayload = { ...requestPayload, recovery };
     return Promise.resolve()
-      .then(() => {
-        return this.pillarWalletSdk.wallet.registerAuthServer(requestPayload);
-      })
-      .then(({ data }) => data)
+      .then(() => this.pillarWalletSdk.wallet.registerAuthServer(requestPayload))
+      .then(({ data }) => { console.log('data: ', data); return data; })
```
vyvil, line 203: Do we need this log here?

src/utils/common.js
```diff
@@ -589,3 +595,22 @@ export const formatTransactionFee = (feeInWei: string | number, gasToken: ?GasTo
...
+
+  const words = hexString.toString().split(' ');
+  const firstWord = words[0];
+
+  if (words.length === 1) {
+    if (firstWord.length <= totalTruncatedSum) return firstWord;
+    return `${firstWord.slice(0, startCharsCount)}${separator}${firstWord.slice(-endCharsCount)}`;
+  }
+
+  return hexString;
```
vyvil, line 615: If `hexString` contains two or more words, none of it will be truncated. Is this expected behaviour?

src/actions/connectedDevicesActions.js
```diff
@@ -0,0 +1,187 @@
...
+
+      // check if account was created
+      const smartWalletAccountDevices = get(getState(), 'smartWallet.connectedAccount.devices', []);
+      const newSmartWalletAccountDevice: ?SmartWalletAccountDevice = smartWalletAccountDevices.find(
+        ({ device }) => addressesEqual(device.address, deviceAddress),
+      );
+
+      // check if device was deployed
+      if (!newSmartWalletAccountDevice || !isSmartWalletDeviceDeployed(newSmartWalletAccountDevice)) {
+        dispatch({ type: RESET_ADDING_CONNECTED_DEVICE_ADDRESS });
```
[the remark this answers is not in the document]
poocart: Sorry, some kind of leftover, this check shouldn't include deployment check as the transaction is just sent.

src/screens/ConnectedDevices/ManageConnectedDevices.js
```diff
@@ -0,0 +1,164 @@
...
+          <RemoveAction onPress={() => removeDevice(item)}>
+            <BaseText negative>Remove</BaseText>
+          </RemoveAction>
+        }
+      </ListItemWithImage>
+    );
+  };
+
+  const devicesByLatest = orderBy(devices, ['updatedAt'], ['desc']);
+  const emptyStyle = { justifyContent: 'center', alignItems: 'center' };
```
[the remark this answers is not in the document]
poocart: It doesn't crash, but doesn't provide required output as well, fixed.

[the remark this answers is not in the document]
poocart: Had to add ScrollWrapper as well and then it started crashing so applied props to `contentContainerStyle`.

src/screens/RecoveryPortal/RecoveryPortalSetupIntro.js
```diff
@@ -0,0 +1,116 @@
...
+  <ContainerWithHeader
+    headerProps={{ floating: true }}
+    backgroundColor="#faf3f5"
+  >
+    <ScrollWrapper contentContainerStyle={{ paddingTop: 80 }}>
+      <CustomWrapper>
+        <FeatureIcon source={deviceRecoveryIcon} />
+        <Title>Recovery device</Title>
+        <BodyText>
+          Recovery agents are individuals, services or secondary devices, like,
```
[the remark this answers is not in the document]
poocart: I guess text copies are yet to be updated because some of them were even Lorem Ipsum per design, I want to provide this build that works and then we can go over texts as I believe even the flows might change because it has been over 5 months since the feature was put on hold.

src/screens/RecoveryPortal/RecoveryPortalSetupIntro.js
```diff
@@ -0,0 +1,116 @@
...
+  padding: 0 ${spacing.rhythm}px;
+`;
+
+const FeatureIcon = styled(CachedImage)`
+  height: 124px;
+  width: 124px;
+  margin-bottom: 24px;
+`;
+
+const deviceRecoveryIcon = require('assets/images/logo_recovery_device.png');
```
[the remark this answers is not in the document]
poocart: Fixed this, but I guess this screen will change as well as it was introduced when we didn't have dark theme yet.

src/screens/Menu/RecoverySettings.js
```diff
@@ -55,13 +77,25 @@ class RecoverySettings extends React.Component<Props, State> {
...
         key: 'linkedDevices',
         title: 'Linked devices',
-        subtitle: 'Assign your contacts as recovery agents for restoring the wallet',
-        disabled: true,
-        label: 'soon',
+        subtitle: 'Manage Smart Wallet account devices',
+        onPress: () => navigation.navigate(MANAGE_CONNECTED_DEVICES),
+      },
+      {
+        key: 'linkedDevices',
```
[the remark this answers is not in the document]
poocart: Whoops 👍

src/utils/common.js
poocart: Hex string value cannot have space by design.

src/actions/smartWalletActions.js
```diff
@@ -1300,6 +1376,98 @@ export const importSmartWalletAccountsAction = (privateKey: string, createNewAcc
...
+        type: 'warning',
+        title: 'Unable to deploy device',
+        autoClose: false,
+      });
+      return;
+    }
+
+    const { accounts: { data: accounts } } = getState();
+    const accountId = getActiveAccountId(accounts);
+    const accountAddress = getActiveAccountAddress(accounts);
```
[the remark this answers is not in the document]
poocart: No, good catch though, it was also on other actions I've added and on `setSmartWalletEnsNameAction` action, fixed all.

src/screens/RecoveryPortal/RecoveryPortalSetupComplete.js
poocart: "Worry not you must" – Yoda

poocart: Fixed, but copies are still subject to change after first review 👍

src/screens/RecoveryPortal/RecoveryPortalSetupIntro.js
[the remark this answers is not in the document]
poocart: Please resolve if you agree.

src/screens/RecoveryPortal/RecoveryPortalSetupIntro.js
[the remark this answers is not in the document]
poocart: Please resolve if you agree.

src/actions/smartWalletActions.js
```diff
@@ -1300,6 +1376,98 @@ export const importSmartWalletAccountsAction = (privateKey: string, createNewAcc
...
+    const historyTx = buildHistoryTransaction({
+      from: accountAddress,
+      hash: accountDeviceDeploymentHash,
+      to: accountAddress,
+      value: '0',
+      asset: ETH,
+      tag: SMART_WALLET_ACCOUNT_DEVICE_ADDED,
```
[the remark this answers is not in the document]
poocart: There's more transactions like this, I think we should address this in separate PR.

PR #1982, opened 2020-05-21
feature/referrals-flagging [PIL-602]

vyvil:
This PR adds a flag to control referral reward texts across the app and alters related screens to utilise the flag.

We will need some adjustments when we will be issuing badges as a reward.
And before starting other referral campaigns on incoming rewards screen.


src/actions/referralsActions.js
```diff
@@ -330,12 +331,27 @@ export const fetchReferralRewardAction = () => {
       type: FETCHING_REFERRAL_REWARD_AMOUNT,
     });
 
-    const referralRewards: RewardsByCompany = await api.getReferralRewardValue(walletId, referralToken);
+    const referralCampaignsInfo = await api.getReferralCampaignsInfo(walletId, referralToken);
+    const referralRewards: RewardsByCompany = Object.keys(referralCampaignsInfo)
+      .reduce((rewardsByCampaign, campaignName) => {
+        const isCampaignActive = get(referralCampaignsInfo, `[${campaignName}].isActive`);
```
jsidorenko, line 337: I think there is no need for those brackets (I could be wrong)
```suggestion
        const isCampaignActive = get(referralCampaignsInfo, `${campaignName}.isActive`);
```

jsidorenko: On the other hand, if `campaignName` could have spaces, then my solution won't work

PR #1984, opened 2020-05-21
[PIL-580] Remove Bitcoin from account selection Part 1/2

alazarevski:
- Merges balance in balances of an account (SW/KW) if the Bitcoin asset is enabled and BTC is initialized
- Merges the history of transactions by adjusting the history selector
- Filters out BTC in SendAssets screen (TODO optional: find a way to select a contact's btc address and change this back)
- Remove BTC from Accounts and the Account Switcher 9c7710a
- Init BTC on onboarding and graduate it out of featureFlags efac83d

~TODO: Initialize BTC and limit featureflags~

~TODO: remove BTC Wallet from Accounts and Home Switch, and see if the activeNetwork check is necessary everywhere in the app~

### DIFF
--- a/src/screens/Assets/AssetsList.js
+++ b/src/screens/Assets/AssetsList.js
@@ -318,13 +318,17 @@ class AssetsList extends React.Component<Props, State> {
         {symbol === 'BTC' &&
           <>
             <ListItemWithImage
-              actionLabel="BTC Wallet"
               onPress={() => {
               this.navigateToBTCAsset({ ...props, tokenType: TOKENS });
             }}
               address={props.address}
               label={name}
               avatarUrl={fullIconUrl}
+              balance={{
+                balance: formatAmount(balance),
+                value: formattedBalanceInFiat,
+                token: symbol,
+              }}
               fallbackToGenericToken
             />
             <CheckAuth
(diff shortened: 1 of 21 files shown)


src/screens/Assets/AssetsList.js
```diff
@@ -369,9 +373,9 @@ class AssetsList extends React.Component<Props, State> {
       .map(id => assets[id])
       .map(({ symbol, ...rest }) => ({
         symbol,
-        balance: getBalance(balances, symbol),
         paymentNetworkBalance: getBalance(paymentNetworkBalances, symbol),
         ...rest,
+        balance: getBalance(balances, symbol),
```
alazarevski, line 378: this was moved here because `...rest` for btcAsset unfortunately had a balance: 0 default which overrides the real balance.

alazarevski:
> LGTM, but there is an issue with activity feed now (and event details modals etc). It checks if we're on bitcoin network via `activeBlockchainSelector` and renders bitcoin history differently if we're on bitcoin network. Now we're not switching networks, so it doesn't work. We should probably check if a transaction event is BTC (`event.asset == BTC`).
> Let me know if you want me to do this, I'm low on tasks anyway right now :p

Hi which is the specific difference for the activeNetwork ? I merged down the tx history and the EventDetails was an issue before that, the 'Send Back' and other new modal features were not quite functioning correctly before the PR too

PR #1985, opened 2020-05-21
PIL-593 Feature: Switch fees token

jsidorenko:
In this PR:
- Added UI selector if the user wants to pay tx fees by using PLR token
- Fixed Relayer migration (fixed the way we listen to migration updates and added modal auto-closing)
- Connected all PPN actions (topup, settle, withdraw) to the new selector
- Refactored Send tokens/collectibles/exchange/WC components. The main idea here is that smart wallet api will return you the cost for tx in eth + in gas token + gasToken object itself. Then depending on useGasToken selector we decide which field to take


src/components/SendTokens/EthereumTokens.js
```diff
@@ -148,37 +161,43 @@ class SendETHTokens extends React.Component<Props, State> {
...
 
   componentDidUpdate(prevProps: Props) {
-    const { session, gasToken, gasInfo } = this.props;
+    const {
+      session, useGasToken, gasInfo, isSmartAccount, isGasTokenSupported,
+    } = this.props;
+    const { showRelayerMigrationModal } = this.state;
 
-    if (prevProps.session.isOnline !== session.isOnline && session.isOnline) {
+    if (prevProps.session.isOnline !== session.isOnline && session.isOnline && isSmartAccount) {
```
graszka22, line 188: shouldn't it be `!isSmartAccount`? maybe move this check into `fetchGasInfo` action, so that we don't have to check it every time?

jsidorenko:
@graszka22 those gas prices we get from the BCX: https://api-qa-bcx.pillarproject.io/wallet-client/gasstation

src/components/SendTokens/EthereumTokens.js
jsidorenko: Good catch!
I don't think it's worth now to optimise the `fetchGasInfo` action since we're going to remove KW shortly together with that legacy gas fetching

jsidorenko:
@jdnmr 
> Maybe we should let the users know that paying fees in PLR is only possible for SW? It's quite confusing if you switch to PLR and nothing changes in KW. E.g. we could change the selector label to "Pay Smart Wallet fees with PLR"

Since we're going to remove the KW from the UI I don't think we should anything for KW

> Why do we allow ETH fees for Pillar Tank? It always says that the user doesn't have enough ETH for transaction, so they have to go back to settings and switch every time they want to use the Pillar Tank if they wish to pay with ETH elsewhere.

This is more a product question... The good thing is that new users will start with PLR as gasToken enabled, so no need for them to switch to ETH. And soon we will add the ability to select different tokens for the fee

PR #1993, opened 2020-05-22
PIL-589 Fix header text wrap

graszka22:
- to reproduce this probably you'll need to set a bigger font size in android settings
- I could just set `sideFlex` to `30px` or something like that but if somebody had even bigger font, or longer username he'd experience wrapping again, so with a little of layout tweaking we have no more wrapping  at all


graszka22:
Ok, so I've reverted the previous commit and come to conclusion that you can't have everything just by styles. Absolute positioning will lead to overlap and any flexing will lead to text wrapping or the username not centered. So I've reduced the font scaling so that it doesn't wrap the support link.

PR #1995, opened 2020-05-22
Fix/dapp-fees

jdnmr:
WC request fees calculations were wrongly attempted as if SW was connected, while we only connect KW. So I'm deleting the SW-related code for now as a temporary fix.


src/screens/WalletConnect/WalletConnectCallRequest.js
```diff
@@ -187,17 +197,22 @@ class WalletConnectCallRequestScreen extends React.Component<Props, State> {
...
+    } = this.props;
 
-    let gasLimit;
-    if (!isSmartAccount) {
-      gasLimit = await calculateGasEstimate({ ...this.transactionDetails, from: activeAccountAddress });
-      this.setState({ gasLimit });
-    }
+    // let gasLimit;
+    // if (!isSmartAccount) {
+    const gasLimit = await calculateGasEstimate({ ...this.transactionDetails, from: activeAccountAddress });
```
jsidorenko, line 207: I think we should calculate using the KW address and not `activeAccountAddress`

PR #1979, opened 2020-05-19
[PIL-580] Bitcoin support changes and improvements

alazarevski:
* List BTC and redirect to Bitcoin Asset Screen from SW and back

* Merge BTC asset balance in Account (SW) if enabled

* tidy up constants

* cleanup code

* ~~remove Bitcoin Wallet from Accounts Switch Screen and Toggle Screen~~ (to be agreed upon)

* init a BTC address on onboardingActions.js, ~~remove featureFlags~~ only if imported and btcFeatureFlag enabled

Screen Shot 2020-05-19 at 15 36 00Screen Shot 2020-05-19 at 15 36 11Screen Shot 2020-05-22 at 15 36 11

PR #1999, opened 2020-05-25
Feature/all wallets data in home [PIL-569]

vyvil:


src/components/EventDetails/EventDetails.js
```diff
@@ -356,22 +369,24 @@ class EventDetail extends React.Component<Props, State> {
...
-    } else {
-      formattedFee = parseFloat(formatUnits(btcFee, assetDecimals));
+      const formattedFee = parseFloat(utils.formatEther(fee.toString()));
+      const rate = getRate(rates, ETH, fiatCurrency);
+      const formattedFiatValue = formatFiat(formattedFee * rate, fiatCurrency);
+      return `Fee ${formattedFee} ETH (${formattedFiatValue})`;
+    } else if (btcFee) {
+      const formattedBTCFee = parseFloat(formatUnits(btcFee, assetDecimals));
+      const btcRate = getRate(rates, BTC, fiatCurrency);
+      const formattedFiatValue = formatFiat(formattedBTCFee * btcRate, fiatCurrency);
```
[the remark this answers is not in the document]
vyvil: Will take a look and simplify it, thanks :)

As for duplications, Home feed is supposed to show history of all accounts now so they should be there. One item represents outcoming transaction from one account into another, the other one - incoming transaction from this first one. It is applied to all transactions between all accounts - Key wallet, Smart Wallet, PPN. (Only exception - settlement action, as we show both values (outcoming and incoming) on the same list item).
This should be visible on Home screen no matter which account is active.
If you are seeing this on Smart Wallet only, then there's a bug (as it should look the same when Key wallet, PPN or BTC is active). Let me know what is your case :)

src/components/EventDetails/EventDetails.js
```diff
@@ -120,10 +127,12 @@ import { switchAccountAction } from 'actions/accountsActions';
 import { goToInvitationFlowAction } from 'actions/referralsActions';
 import { updateTransactionStatusAction } from 'actions/historyActions';
 import { lookupAddressAction } from 'actions/ensRegistryActions';
+import { setActiveBlockchainNetworkAction } from 'actions/blockchainNetworkActions';
```
alazarevski, line 130: We won't be using this anymore, please use if asset.symbol === BTC condition to determine BTC specific parts

src/components/EventDetails/EventDetails.js
```diff
@@ -956,28 +961,53 @@ class EventDetail extends React.Component<Props, State> {
...
             } else if (isPending) {
               buttons = [viewOnBlockchainButton, inviteToPillarButton];
             } else {
               buttons = [sendBackToAddress, inviteToPillarButton];
             }
-          } else if (contactFound) {
-            if (isPending) {
-              buttons = [messageButton, viewOnBlockchainButtonSecondary];
+          } else if (isBitcoinTrx) {
+            if (bitcoinFeatureEnabled) {
```
alazarevski, line 1000: A good check would be bitcoinAddresses.length > 0 for this instead of the feature check

src/actions/historyActions.js
```diff
@@ -478,13 +481,25 @@ export const restoreTransactionHistoryAction = () => {
...
+        data: {
+          BITCOIN_ENABLED: bitcoinFeatureEnabled,
+        },
+      },
     } = getState();
 
+    if (forAllAccounts) {
+      await dispatch(restoreTransactionHistoryAction());
+      await dispatch(fetchSmartWalletTransactionsAction());
+      if (bitcoinFeatureEnabled) dispatch(fetchBTCTransactionsHistoryAction());
```
alazarevski, line 499: BTC fetch is handled in the older PR, the inner action checks if there are any addresses generated and handles transactions accordingly https://github.com/pillarwallet/pillarwallet/pull/1979/files#diff-7ea48032f831c00d653dddebf080d034R487

src/actions/collectiblesActions.js
```diff
@@ -181,9 +189,43 @@ export const fetchCollectiblesHistoryAction = () => {
...
+export const fetchAllAccountsCollectiblesAction = () => {
+  return async (dispatch: Dispatch, getState: GetState) => {
+    const {
+      accounts: { data: accounts },
+    } = getState();
+
+    const promises = accounts.map(async account => {
+      await dispatch(fetchCollectiblesAction(account));
+    });
+    await Promise.all(promises).catch(_ => _);
```
jsidorenko, line 202: What do you think about:
```suggestion
    await Promise.allSettled(promises).catch(_ => _);
```

src/actions/collectiblesActions.js
```diff
@@ -181,9 +189,43 @@ export const fetchCollectiblesHistoryAction = () => {
...
+export const fetchAllAccountsCollectiblesHistoryAction = () => {
+  return async (dispatch: Dispatch, getState: GetState) => {
+    const {
+      accounts: { data: accounts },
+    } = getState();
+
+    const promises = accounts.map(async account => {
+      await dispatch(fetchCollectiblesHistoryAction(account));
+    });
+    await Promise.all(promises).catch(_ => _);
```
jsidorenko, line 216:
```suggestion
    await Promise.allSettled(promises).catch(_ => _);
```

src/actions/historyActions.js
```diff
@@ -164,17 +165,18 @@ export const fetchSmartWalletTransactionsAction = () => {
       smartWallet: { lastSyncedTransactionId },
     } = getState();
 
-    const activeAccount = getActiveAccount(accounts);
-    if (!activeAccount || !checkIfSmartWalletAccount(activeAccount)) return;
+    const smartWalletAccount = findFirstSmartAccount(accounts);
```
jsidorenko, line 168: There is a new `isActiveAccountSmartWalletSelector` selector (if you want to check the active account type)

src/selectors/balances.js
```diff
@@ -29,3 +32,28 @@ export const accountBalancesSelector = createSelector(
     return balances[activeAccountId] || {};
   },
 );
+
+export const allBalancesSelector = createSelector(
+  balancesSelector,
+  availableStakeSelector,
+  (balances, ppnBalance) => {
+    const allBalances = Object.keys(balances).reduce((memo, account) => {
+      if (balances[account] && Object.keys(balances[account]).length) {
```
jsidorenko, line 41:
```suggestion
      if (!isEmpty(balances[account])) {
```

src/selectors/balances.js
```diff
@@ -29,3 +32,28 @@ export const accountBalancesSelector = createSelector(
...
+        const accountsBalances = Object.keys(balances[account]).map((symbol) => balances[account][symbol]);
```
jsidorenko, line 42: I think we could use Object.values(balances[account]) here

jsidorenko: P. S. if Flow complains about this, pls check our codebase for 'Object.values(' usage

src/utils/feedData.js
```diff
@@ -66,6 +79,42 @@ export function mapTransactionsHistory(
...
       };
     });
+
+  if (keepHashDuplicatesIfBetweenAccounts) {
+    const accountsAddresses = accounts.map((acc) => getAccountAddress(acc));
+    const ascendingHistory = orderBy(concatedHistory, ['createdAt'], ['asc']);
+
+    const historyWithTrxBetweenAcc = ascendingHistory.reduce((alteredHistory, historyItem) => {
+      const { from: fromAddress, to: toAddress, hash } = historyItem;
+      if (alteredHistory.some((item) => item.hash === hash)) {
```
jsidorenko, line 89: IMO if we extract all those .some into some descriptive variables, the code will become more readable

src/actions/historyActions.js
vyvil: Here we need smart wallet account even if it's not active. Otherwise,  updated fetchTransactionsHistoryAction(true) ([L484](https://github.com/pillarwallet/pillarwallet/pull/1999/files/d6999f0931d69637f410d244b7f734e858d7b338#diff-7ea48032f831c00d653dddebf080d034R484)) would not work as expected.

jsidorenko: Right, this is what I was suspecting :D gut, pls ignore my suggestion above then :)

src/components/EventDetails/EventDetails.js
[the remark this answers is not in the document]
graszka22: > This should be visible on Home screen no matter which account is active.
If you are seeing this on Smart Wallet only, then there's a bug (as it should look the same when Key wallet, PPN or BTC is active). Let me know what is your case :)

I checked the other cases and they're duplicated everywhere, so it's ok :)

src/components/EventDetails/EventDetails.js
vyvil: Is it, though?
If no address exists, no ActivityFeedItem is visible, so this code will not be used at all.
Now, if we would like to turn off this feature and user has addresses stored, no action buttons will be visible here :)

alazarevski: I meant
 ```
else if (isBitcoinTrx) {
              buttons = isPending ? [viewOnBlockchainButton] : [sendBackBtc];
```

alazarevski: got the changes here https://github.com/pillarwallet/pillarwallet/commit/629eefb290674cdbf464cadf2a578f9a35f99b87#diff-ec904b6cf8e8afec5137f06ae6e584f5R979

src/actions/historyActions.js
vyvil: Hm, this is some other case. It fetches history for all accounts.
The one you refer to, fetches history of BTC and of active acc. On Home feed we need it to be fetched for BTC, SW and KW at the same time.
I've also using the same `fetchBTCTransactionsHistoryAction` method to fetch BTC history that is used in the method you refer to. So is there an issue? Am I missing something? :)

src/components/EventDetails/EventDetails.js
vyvil: As of this current version, we still use networks. So I'm keeping this for now. Will adjust when your PR will be in and this branch will be updated. To be sure that it works :)

PR #2006, opened 2020-05-27
[PIL-584] Feature/smart wallet for WalletConnect

jdnmr:
- Use Smart Wallet for any WC connections;
- Delete all current (KW) connections if they exist, and notify user with a Toast;
- Disable connections to certain dapps. For now they include Mooni and LocalCryptos (they don't support SW) and Binance (which never worked before; it doesn't see asset balances, neither on KW nor SW);
- Fix gas settings for WC transactions. Try to use transaction's `gas` parameter before falling back to `gasLimit`.


src/utils/exploreApps.js
```diff
@@ -51,12 +51,12 @@ export const APPS: AppItem[] = [
...
   },
-  {
-    name: 'Mooni',
-    logo: mooniLogo,
-    text: 'A simple solution to cash out cryptocurrencies in fiat to a bank account, without KYC.',
-    url: 'mooni.tech',
-  },
+  // {
+  //   name: 'Mooni',
+  //   logo: mooniLogo,
```
jsidorenko, line 56: What if we still show that app, but put a label as Coming soon?

src/actions/walletConnectActions.js
```diff
@@ -215,8 +217,18 @@ const subscribeToSessionRequestEvent = (connector: Connector) => {
 
         return;
       }
-      dispatch(walletConnectSessionReceived());
 
+      if (!shouldAllowSession(peerMeta.url)) {
+        Toast.show({
+          type: 'warning',
+          title: 'Cannot connect',
+          message: 'We are sorry. We do not support this application.',
```
jsidorenko, line 225: Maybe it's better to say that application you are trying to connect doesn't support Smart accounts?

src/actions/walletConnectActions.js
```diff
@@ -399,11 +426,15 @@ export const approveSessionAction = (peerId: string) => {
     if (peerMeta) {
       dispatch(killWalletConnectSessionByUrl(peerMeta.url, peerId));
     }
-    const { wallet: { data: { address } } } = getState();
-
+    const {
+      accounts: { data: accounts },
+    } = getState();
     try {
+      const smartAcc = findFirstSmartAccount(accounts);
```
poocart, line 433: This should be first Smart Wallet account or just current active Smart Wallet account?

graszka22: Can we have multiple Smart Wallet accounts? 👀

jsidorenko: Technically we can since SW sdk returns us an array of account we have, but we don't allow user via our app to create more than one account

src/actions/walletConnectActions.js
```diff
@@ -510,7 +550,18 @@ const loadLegacySessions = async (): Promise<Session[]> => {
...
+    const {
+      walletConnectSessions: { isImported, sessions },
+      accounts: { data: accounts },
+    } = getState();
+
+    const keyWallet = findKeyBasedAccount(accounts);
+    if (!keyWallet) return;
+    const keyWalletAddress = getAccountAddress(keyWallet);
+
+    if (shouldClearWCSessions(sessions, keyWalletAddress)) {
```
jsidorenko, line 562: Shall we check if the user has a smart account?

graszka22: imho we shouldn't. If user doesn't have a SW and we leave the WC connected with key wallet, then sending transactions etc would break because we assume that user is connected via smart wallet. I've just removed the code for fetching fees for key wallet. 
If user doesn't have a smart account and we remove WC connection then nothing breaks. When user tries to connect without SW he gets an error message and we encourage him to set up SW.

jsidorenko: Agree, we should show then this encouraging message somewhere

PR #2008, opened 2020-05-27
additional encouraging send button in assets screen

poocart:
https://linear.app/pillarproject/issue/PIL-513/encourage-users-to-migrate-to-sw

This button does exactly the same as "Send" and the reason to add this is "to encourage users migrate to Smart Wallet" per Dmitry's explanation.

### DIFF
--- a/src/screens/Asset/Asset.js
+++ b/src/screens/Asset/Asset.js
@@ -36,6 +36,7 @@ import { ScrollWrapper } from 'components/Layout';
 import AssetPattern from 'components/AssetPattern';
 import { BaseText, Paragraph, MediumText } from 'components/Typography';
 import SWActivationCard from 'components/SWActivationCard';
+import Button from 'components/Button';
 
 // actions
 import { fetchAssetsBalancesAction } from 'actions/assetsActions';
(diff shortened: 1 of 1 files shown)


src/screens/Asset/Asset.js
```diff
@@ -409,9 +418,18 @@ class AssetScreen extends React.Component<Props, State> {
               isReceiveDisabled={!isReceiveActive}
               showButtons={isSynthetic ? ['receive'] : undefined}
             />
-            {!isSendActive &&
-              <SWActivationCard />
+            {isKeyBasedActive &&
+              <TransferButtonWrapper>
+                <Button
+                  title="Transfer to Smart Wallet"
+                  onPress={() => this.goToSendTokenFlow(assetData)}
```
vyvil, line 425: Hmmm, why not to go straight to confirmation screen with SW selected as a receiver and value set to all (if we are willing to encourage users to transfer all related asset value?
Or at least to amount selection screen with SW already selected as a receiver (jumping one step further so no extra click would be needed)? :)

vyvil: Just read the last comment in linear (must have missed that before) :))
Still weird we are not doing this click for a user, tho :)))

poocart:
> LGTM but what if we have asset balance = 0 or we don't have the SW?

Then it won't be clickable (same as for "Send"), as Dmitry said – it should do exactly the same, but encourage users to use Smart Wallet more 🤷‍♂️

src/screens/Asset/Asset.js
```diff
@@ -343,6 +351,7 @@ class AssetScreen extends React.Component<Props, State> {
     });
     const relatedTransactions = isSynthetic ? ppnTransactions : mainnetTransactions;
     const isSupportedByExchange = exchangeSupportedAssets.some(({ symbol }) => symbol === token);
+    const isKeyBasedActive = !!activeAccount && activeAccount.type === ACCOUNT_TYPES.KEY_BASED;
```
jsidorenko, line 354: Minor: `const isKeyBasedActive = !isSmartAccount;`
and in createStructuredSelector:
`isSmartAccount: isActiveAccountSmartWalletSelector,`

PR #2009, opened 2020-05-28
Feature/fee toggle [PIL-576]

vyvil:
This PR adds FeeLabelToggle component that allows user to toggle transaction fee values to see transaction fee in ETH / PLR and fiat.


vyvil:
> Do we need to add the same component to confirmation pages? What do you think?

Yeah, would be nice to have it there as well. I'll double check with they guys.
And if so, I could add those in a separate PR. How does it sound to you? :)

jsidorenko:
> > Do we need to add the same component to confirmation pages? What do you think?
> 
> Yeah, would be nice to have it there as well. I'll double check with the guys.
> And if so, I could add those in a separate PR. How does it sound to you? :)

Sounds good!

PR #2013, opened 2020-05-29
[PIL-603] Feature/restore pin flow

jdnmr:
- Reintroduce PIN for app launch, tx/exchange confirmation, changing biometrics settings;
- Store PIN in keychain. This way we don't need to decrypt wallet every time we provide PIN, which should make pin checks much faster;
- "Old" pin flow is left for several rare cases (migrating users to keychain; fallback after failed/rejected biometrics check; changing pin while having biometrics on);
- Fixed the Toast for biometrics settings change;


src/utils/keychain.js
```diff
@@ -75,5 +77,28 @@ export const getPrivateKeyFromKeychainData = (data?: KeyChainData) => {
...
+export const getWalletFromPkByPin = async (pin: string, withMnemonic?: boolean) => {
+  const keychainData: KeyChainData = await getKeychainDataObject();
+  const { pin: pinFromKeychain, privateKey, mnemonic } = keychainData;
+  if (pin && pin === pinFromKeychain && privateKey) {
+    let wallet = constructWalletFromPrivateKey(privateKey);
+    wallet = {
+      ...wallet,
+      privateKey, // we need PK for changing biometrics settings
+    };
+    return withMnemonic ? { ...wallet, mnemonic } : wallet;
```
jsidorenko, line 92:
```suggestion
    const wallet = withMnemonic && mnemonic
      ? constructWalletFromMnemonic(mnemonic)
      : constructWalletFromPrivateKey(privateKey);
    return wallet;
```

src/screens/WalletConnect/WalletConnectPinConfirm.js
```diff
@@ -184,19 +185,23 @@ class WalletConnectPinConfirmScreeen extends React.Component<Props, State> {
...
         headerProps={{ onBack: this.handleBack }}
+        enforcePin={!useBiometrics}
       />
     );
   }
 }
 
 const mapStateToProps = ({
   walletConnect: { requests },
+  appSettings: { data: { useBiometrics = false } },
```
jsidorenko, line 202: Maybe instead of putting the default param here, it's better to use the appSettings migration when we load the data on app start?

src/actions/authActions.js
```diff
@@ -129,14 +131,19 @@ export const loginAction = (
...
         const { wallet: encryptedWallet } = await storage.get('wallet');
         await delay(100);
         const saltedPin = await getSaltedPin(pin, dispatch);
         wallet = await decryptWallet(encryptedWallet, saltedPin, { mnemonic: true });
         // no further code will be executed if pin is wrong
         // migrate older users for keychain access OR fallback for biometrics login
-        await setKeychainDataObject({ privateKey: wallet.privateKey, mnemonic: wallet.mnemonic || '' }, useBiometrics);
+        await setKeychainDataObject({
+          pin, privateKey: wallet.privateKey, mnemonic: wallet.mnemonic || '',
+        }, useBiometrics);
```
jsidorenko, line 146:
```suggestion
        await setKeychainDataObject(
          {
            pin,
            privateKey: wallet.privateKey,
            mnemonic: wallet.mnemonic || '',
          },
          useBiometrics);
```

PR #2016, opened 2020-06-01
Fix/Filter BTC out of asset search

alazarevski:
Filter BTC out of asset search in Smart Wallet or Key Wallet.
This was discussed that it confuses users to see Bitcoin in their assets list if they show it.

### DIFF
--- a/src/actions/assetsActions.js
+++ b/src/actions/assetsActions.js
@@ -551,6 +551,7 @@ export const getSupportedTokens = (supportedAssets: Asset[], accountsAssets: Ass
 
   const updatedAccountAssets = supportedAssets
     .filter(asset => accountAssetsTickers.includes(asset.symbol))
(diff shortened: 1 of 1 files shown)


src/actions/assetsActions.js
```diff
@@ -512,7 +513,7 @@ export const searchAssetsAction = (query: string) => {
 
     const filteredAssets = supportedAssets.filter(({ name, symbol }) => {
       return name.toUpperCase().includes(search) || symbol.toUpperCase().includes(search);
-    });
+    }).filter(({ symbol }) => symbol !== BTC);
```
vyvil, line 516: We are already filtering :)
So maybe just:
```
const filteredAssets = supportedAssets.filter(({ name, symbol }) => {
      return symbol !== BTC && (name.toUpperCase().includes(search) || symbol.toUpperCase().includes(search));
    });
```

m?

PR #2017, opened 2020-06-01
PIL-568 Fix/offers engine label

jdnmr:
Instead of last offer search results, it displays a total number of supported providers basing on `providersMeta`.


src/screens/Services/Services.js
```diff
@@ -49,14 +48,12 @@ const mastercardIcon = require('assets/icons/mastercard.png');
...
-      navigation, theme, offers, baseFiatCurrency,
+      navigation, theme, baseFiatCurrency, providersMeta,
     } = this.props;
     const colors = getThemeColors(theme);
 
-    const providersCount = uniq((offers || []).map(offer => offer.provider)).length;
-
-    const offersBadge = providersCount ? {
-      label: `${providersCount} exchanges`,
+    const offersBadge = Array.isArray(providersMeta) ? {
```
vyvil, line 55: Hm, maybe we would need to check if providersMeta.length > 0? Otherwise we might end up showing "0 exchanges" if something goes wrong. Maybe then it's better not to show anything?

jdnmr: great catch!

PR #2019, opened 2020-06-02
PIL-517 Feature/Sell Max button in Exchange

jdnmr:
Also fixed a minor UX issue with popular swaps, added feeToken balance to ExchangeConfirm if balance is too low to pay the fee.


src/screens/Exchange/Exchange.js
```diff
@@ -443,6 +448,28 @@ class ExchangeScreen extends React.Component<Props, State> {
     }
   }
 
+  handleSellMax = () => {
+    const { value } = this.state;
+    const { balances } = this.props;
+    const selectedAssetSymbol = this.getSelectedFromAssetSymbol();
+    const chosenAssetBalance = formatAmount(getBalance(balances, selectedAssetSymbol));
+    value.fromInput.input = chosenAssetBalance;
```
vyvil, line 456: Hm, maybe it would be better to not mutate value here and instead make a copy of it first?

src/screens/Exchange/Exchange.js
```diff
@@ -831,6 +864,7 @@ class ExchangeScreen extends React.Component<Props, State> {
         >
           {!isSubmitted && <HotSwapsHorizontalList onPress={this.onSwapPress} swaps={swaps} />}
           <FormWrapper bottomPadding={isSubmitted ? 6 : 30}>
+            {this.shouldShowSellMax() && <SellMaxButton onPress={this.handleSellMax} />}
```
vyvil, line 867: What do you think about adding this as a TextInput element?
Would be more reusable and cleaner (no absolute positioning needed).
As for this component itself, we also have ButtonText component that has native Android feel.
Maybe it could be reused here :)

src/screens/Exchange/Exchange.js
```diff
@@ -718,6 +752,7 @@ class ExchangeScreen extends React.Component<Props, State> {
           config: {
             options: { $set: optionsFrom },
             inputAddonText: { $set: valueInFiatToShow },
+            rightLabel: { $set: this.shouldShowSellMax() ? 'Sell max' : '' },
```
graszka22, line 755: services -> buy crypto with fiat -> sell max is visible and GBP is selected

PR #2021, opened 2020-06-02
Feature/transaction notes [PIL-565, PIL-563, PIL-573]

vyvil:
This PR solves 3 linear issues
(since all of them was related to Event Modals, I've added them into one PR):

- Transaction notes [[PIL-565](https://linear.app/pillarproject/issue/PIL-565/implement-transaction-notes)]
- Etherscan reference for transactions [[PIL-563](https://linear.app/pillarproject/issue/PIL-563/etherscan-reference-for-transaction)] (View on blockchain button is removed, event time works as a link to etherscan)
- Show collectibles transaction fees in general transaction history [[PIL-573](https://linear.app/pillarproject/issue/PIL-573/show-collectibles-transaction-fees-in-general-transaction-history)]

And it also updates Referral Reward item in activity feed.


src/actions/collectiblesActions.js
```diff
@@ -229,3 +236,43 @@ export const fetchAllCollectiblesDataAction = (forAllAccounts?: boolean) => {
...
+    const trxInfo = await getTrxInfo(api, hash);
+    if (!trxInfo) return;
+    const {
+      txInfo,
+      txReceipt,
+      nbConfirmations,
+      status,
+    } = trxInfo;
+
+    const accounts = Object.keys(collectiblesHistory);
```
jsidorenko, line 257: I think you can re-use `updateHistoryRecord` helper here

src/components/EventDetails/EventDetails.js
```diff
@@ -325,32 +359,43 @@ export class EventDetail extends React.Component<Props, State> {
...
-    const {
-      event,
-      updateTransactionStatus,
-    } = this.props;
+  syncTxStatus = (txInfo: Transaction | CollectibleTrx) => {
     if (txInfo.status === TX_PENDING_STATUS) {
-      this.timeout = setTimeout(() => updateTransactionStatus(event.hash), 500);
-      this.timer = setInterval(() => updateTransactionStatus(event.hash), 10000);
+      this.timeout = setTimeout(this.updateTransaction, 500);
+      this.timer = setInterval(this.updateTransaction, 10000);
```
jsidorenko, line 384: this will be called every 10 seconds, what if the user will close and open another transaction during that time?

src/actions/collectiblesActions.js
vyvil: Was willing to do so, but collectibles' transactions and other transactions differ in props, hence `updateHistoryRecord` outputs turn out to be different. I could alter typing for this method, but then it goes on an on in other usage places. 

Didn't want to align collectibles and other transactions too much as I would just be adding empty unnecessary data. So I've gone with this approach.

src/components/EventDetails/EventDetails.js
vyvil: We are calling clear method that cleans Timeout and Interval whenever modal is closed:
https://github.com/pillarwallet/pillarwallet/pull/2021/files#diff-ec904b6cf8e8afec5137f06ae6e584f5R344

So if user would open new modal, timeout and timer will be reset (upon close) and then reset into new value (upon open) :)

src/actions/collectiblesActions.js
jsidorenko: Gotcha! While you're using this in one place only - I agree, we could keep that approach and refactor once we need to use it somewhere else

src/components/EventDetails/EventDetails.js
jsidorenko: Yep, you're right!

vyvil:
> All good, but where should be the collectible fees?

Ahh, been working on prod env for this. 
It's fetching transaction info thinking it's transaction made on Ropsten on dev, yet collectibles' transactions are executed on Rinkeby.
I'll update this so we would have it visually the same on prod and dev. Thanks :)

vyvil:
@graszka22,
collectible transaction fee is now visible on dev env as well.
I've also added a spinner if fee is being fetched.

jsidorenko:
Well done 👍

PR #2024, opened 2020-06-03
PIL-582 Feature/upload QR images

jdnmr:
To keep consistent with the camera view in user profile flow, users can upload images with QR code by pressing on a gallery icon displayed in the qr camera view. 

Also reduced WC session timeout from 30s to 10s. Because no information is displayed in the meanwhile, 30s felt like forever and the user had no idea what's going on.


src/components/QRCodeScanner/CameraView.js
```diff
@@ -44,14 +53,83 @@ const HeaderWrapper = styled.SafeAreaView`
...
+    this.setState({ isLoading: true });
+    ImagePicker.openPicker({
+      includeBase64: true,
+    })
+      .then(image => {
+        this.timeout = setTimeout(() => {
+          this.handleError();
+        }, ERROR_TIMEOUT);
+        const buffer = Buffer.from(image.data, 'base64');
+        Jimp.read(buffer, (e, parsedImg) => {
```
graszka22, line 109: This takes ridiculous amount of time on my device (android), maybe because I take large images with my camera by default. I mean it took several minutes already and I'm still waiting :D

jdnmr: Yeah, that's what the timeout is for... guess it doesn't work as it should then :/ will try to reproduce and fix

PR #2027, opened 2020-06-04
Fix/Show full value in EventDetails

alazarevski:
Show full value in EventDetails.js


jsidorenko:
I agree with @vyvil:
'10.0' should be shown as '10';
'10.12345678987654321' it's probably to round to '10.1234' or maybe 6 decimals;
'10.0000000000001' => '10';
but '0.0000000000001' I would suggest to show as is;
meanwhile '0.00001000000003' should be shown as '0.00001'

alazarevski:
this is a regression fix to a task that has already been done, https://linear.app/pillarproject/issue/PIL-492/make-amounts-pretty-and-rounded
please let's consult with Dmitry and refactor in a future PR for the suggestions

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -525,8 +532,9 @@ export class ActivityFeedItem extends React.Component<Props> {
             label: itemLabel,
             subtext,
             avatarUrl,
+            fullItemValue: `${directionSymbol} ${value} ${event.asset}`,
             itemValue: `${directionSymbol} ${formattedValue} ${event.asset}`,
-            valueColor: isReceived && formattedValue !== '0' ? 'positive' : 'text',
+            valueColor: isReceived && value !== '0' ? 'positive' : 'text',
```
poocart, line 537: `value` can be `0.0` as well.

paragramm:
18 decimals isn't the case for real life use, so it's safe to round it up to 6 decimals.

it gets trickier with 0.000000000000000001 ETH (18 decimals). Rounding up to 0.000001 ETH (6 decimals) makes it millions times more, relative to ETH actual amount. However, in absolute fiat values, it still something like $0.000002. I'd argue that the compromise here would be to display 0.000001 ETH (6 decimals) for better readability. Users think and operate in fiat mostly and they are generally loyal to rounding crypto values (unless its inaccuracy doesn't cost them more than .00001 USD)

paragramm:
Furthermore, the rounded 0.000001 ETH could be explained better by showing it as <0.000001 ETH.

On a side note, the exact amounts may be required by users for taxation purposes. This will be addressed in future as exportable transaction reports where all the numbers calculated precisely.

PR #2031, opened 2020-06-05
init smart wallet sdk action without being active

poocart:
The problem was that on auth login we were dispatching smart wallet account init action, but in the action itself there was a check that it should only init smart wallet if smart wallet account is active therefore leading to not initiate the Smart Wallet and then any other calls that calls Smart Wallet SDK fails. Failure easy to replicate: close the app in key based account active and reopen.

### DIFF
--- a/src/actions/accountsActions.js
+++ b/src/actions/accountsActions.js
@@ -287,7 +287,7 @@ export const initOnLoginSmartWalletAccountAction = (
 
     const inactiveAccounts = getInactiveUserAccounts(accounts);
     const activeAccount = getActiveAccount(accounts);
-    const smartWalletAccount = getAccountType(activeAccount) === ACCOUNT_TYPES.SMART_WALLET
+    const smartWalletAccount = activeAccount && getAccountType(activeAccount) === ACCOUNT_TYPES.SMART_WALLET
       ? activeAccount
       : findFirstSmartAccount(inactiveAccounts);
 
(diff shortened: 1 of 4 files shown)


src/actions/accountsActions.js
```diff
@@ -269,23 +275,31 @@ export const switchAccountAction = (accountId: string) => {
...
       appSettings: { data: { blockchainNetwork } },
       accounts: { data: accounts },
     } = getState();
 
-    const activeAccountId = getActiveAccountId(accounts);
-    const activeAccountType = getActiveAccountType(accounts);
-    await dispatch(initSmartWalletSdkAction(privateKey));
+    const inactiveAccounts = getInactiveUserAccounts(accounts);
+    const activeAccount = getActiveAccount(accounts);
+    const smartWalletAccount = activeAccount && getAccountType(activeAccount) === ACCOUNT_TYPES.SMART_WALLET
```
jsidorenko, line 290: What if we simply leave:
```suggestion
    const smartWalletAccount = findFirstSmartAccount(accounts);
```
?

poocart: Makes sense 👍

src/actions/accountsActions.js
```diff
@@ -276,16 +277,18 @@ export const initOnLoginSmartWalletAccountAction = (privateKey: string) => {
...
+    const smartWalletAccount = findFirstSmartAccount(accounts);
+    if (!smartWalletAccount) return;
 
-    if (activeAccountType !== ACCOUNT_TYPES.SMART_WALLET) return;
+    const smartWalletAccountId = getAccountId(smartWalletAccount);
+    await dispatch(initSmartWalletSdkAction(privateKey));
 
-    await dispatch(connectSmartWalletAccountAction(activeAccountId));
+    const activeAccountType = getActiveAccountType(accounts);
+    const setAccountActive = activeAccountType !== ACCOUNT_TYPES.SMART_WALLET;
```
jsidorenko, line 287:
```suggestion
    const setAccountActive = activeAccountType === ACCOUNT_TYPES.SMART_WALLET;
```

PR #2049, opened 2020-06-15
PIL-627 Allow paying the fee with unsettled tokens

jsidorenko:


jsidorenko:
I told Dmitry that we need to show to the user the fee in EventDetails. But this is another story

PR #2053, opened 2020-06-17
PIL-674: Ethers v5 migration

pedrouid:
This PR migrates `ethers` to v5 prior to Connext integration. Missing tasks:
- [x] Replace `scrypt` shim with native bindings
- [x] Replace `wallet.encrypt` method with native bindings

I think both of these native bindings can be done externally to ethers without requiring to fork (especially now that ethers v5 doesn't polyfill getRandomValues internally)

cc @poocart


poocart:
Related Linear ticket – https://linear.app/pillarproject/issue/PIL-674/update-to-ethers-v5.

patches/ethers+5.0.0-beta.190.patch
```diff
@@ -0,0 +1,50 @@
...
+-	                        }
+-	                        resolve(new Uint8Array(key));
+-	                    } else if (progressCallback && progress !== lastProgress) {
+-	                        lastProgress = progress;
+-	                        return progressCallback(progress);
+-	                    }
+-	                });
+-	            });
++	        	return new Promise(function(resolve, reject) {
++								let lastProgress = 0;
```
jsidorenko, line 32: the indentation looks wrong

poocart: agree, but I think patcher recognizes it wrong, didn't put too much effort in trying to fix this, but will check once again

jsidorenko: looks like it was replaced with tabs instead of spaces

poocart: was a tricky one, starts with tab, ends with spacing :) copying over exact indentation worked, thanks 👍

PR #2057, opened 2020-06-19
Feature/value selector card

vyvil:
PR that adds Value selector card component.
It does not use it - just adds a reusable component to unblock two other features.

It can be seen in StoryBook, though. With all of its custom cases.

Component currently allows to get form data after form is being updated, preselect asset by providing assets symbol and adding custom "Max" button label.
Let me know if any custom functionality is needed.


src/components/ValueSelectorCard/ValueSelectorCard.js
```diff
@@ -0,0 +1,338 @@
...
+      value: {
+        formSelector: {
+          selector: {},
+          input: '',
+        },
+      },
+      formOptions: {
+        fields: {
+          formSelector: {
+            keyboardType: 'decimal-pad',
```
alazarevski, line 108: Can we please have an option prop to fill only integers without the decimal option ?

src/components/ValueSelectorCard/ValueSelectorCard.js
```diff
@@ -0,0 +1,338 @@
...
+import { themedColors } from 'utils/themes';
+import { SelectorInputTemplate, selectorStructure, inputFormatter, inputParser } from 'utils/formHelpers';
+import { getFormattedBalanceInFiat } from 'screens/Exchange/utils';
+
+import { accountBalancesSelector } from 'selectors/balances';
+import { accountAssetsSelector } from 'selectors/assets';
+import { defaultFiatCurrency } from 'constants/assetsConstants';
+
+
+type Props = {
```
alazarevski, line 49: Can we please have a preselectedValue option, as well as a prop flag to make assets selectable or fixed for the preselectedAsset only

src/components/ValueSelectorCard/ValueSelectorCard.js
```diff
@@ -0,0 +1,338 @@
...
+            customLabel: { $set: this.renderCustomLabel(formSelector?.selector?.symbol) },
+          },
+        },
+      },
+    });
+
+    stateUpdates.formOptions = newOptions;
+
+    this.setState({ ...stateUpdates });
+    getFormValue(formValue?.formSelector);
```
alazarevski, line 246: There is no callback to this function when 'Spend max' is executed

src/components/ValueSelectorCard/ValueSelectorCard.js
alazarevski: Also we would probably need a hasError callback from the Parent component to disable/enable buttons accordingly

src/components/ValueSelectorCard/ValueSelectorCard.js
alazarevski: Also a callback on asset selected from this component to the parent component

src/components/ValueSelectorCard/ValueSelectorCard.js
alazarevski: And the last needed functionality would be a custom balance label and value to be set as a prop
https://app.zeplin.io/project/5db9b5d14c30ae56faaa3030/screen/5ee37a8ce393b3999399ac8a

poocart:
Agree with Alex's comments and added some improvements under https://github.com/pillarwallet/pillarwallet/pull/2064.

PR #2061, opened 2020-06-24
PIL-631 Implement non-zero balance filter in asset search

graszka22:


src/screens/Assets/AssetSearch.js
```diff
@@ -0,0 +1,408 @@
...
+  doAssetsSearch = (query: string) => {
+    const { searchAssets, resetSearchAssetsResult } = this.props;
+    if (query.length < MIN_QUERY_LENGTH) {
+      resetSearchAssetsResult();
+      return;
+    }
+    searchAssets(query);
+  };
+
+  isInSearchMode = () => this.state.query.length >= MIN_QUERY_LENGTH && !!this.props.assetsSearchState;
```
IAmKio, line 209: Assuming the same type, but prefer:

```suggestion
  isInSearchMode = () => this.state.query.length >== MIN_QUERY_LENGTH && !!this.props.assetsSearchState;
```

... for safety.

graszka22: There is no such operator like `>==` in JS. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators

IAmKio: oh whoops! i saw `==`, this is what happens when i review code at 1am in the morning!

PR #2062, opened 2020-06-26
PIL-545 Fix referral bugs and UI issues

graszka22:
- [x] Reduce OTP code to 4 figures
- [x] Update missing phone/email info block style (already done)
- [x] Update 'Referral nobody found' screen
- [x] Strip parsed phone numbers to simple format without parenthesis or hyphens. (already done)
- [x] Remove overlapping backgrounds on iPhone X and similar
- [ ] ~I have invited a contact more than two hours ago. The contact haven't accepted the invitation. My invitation seems to haven't been expired as I can't re-invite the user after timeout.~ (we can't re-invite users and probably the invitation can't expire)
- [x] Dark mode checkboxes are outdated
- [x] Add success screen after sending out invites https://zpl.io/b60g75q. Close button leads to Home screen. (the success screen is already implemented, I just updated the navigation)


src/constants/referralsConstants.js
```diff
@@ -33,3 +33,4 @@ export const FETCHING_REFERRAL_REWARD_AMOUNT = 'FETCHING_REFERRAL_REWARD_AMOUNT'
 export const SET_REFERRAL_REWARD_ISSUER_ADDRESSES = 'SET_REFERRAL_REWARD_ISSUER_ADDRESSES';
 export const SET_PILLAR_REWARD_CAMPAIGN_STATUS = 'SET_PILLAR_REWARD_CAMPAIGN_STATUS';
 export const SET_1WORLD_CAMPAIGN_STATUS = 'SET_1WORLD_CAMPAIGN_STATUS';
+export const OTP_DIGITS = 4;
```
IAmKio, line 36: I'm assuming this will always be 4...

IAmKio:
@graszka22 could you please update the base branch when you get a moment, i'm keen to get these fixes into the next RC which may go out to internal testing today.

src/constants/referralsConstants.js
graszka22: yeah... that's why it's a constant :D previously it was hardcoded in several places in the code

PR #2065, opened 2020-06-29
PIL-688 Fix action buttons for non-deployed wallets

jsidorenko:


src/components/EventDetails/EventDetails.js
```diff
@@ -718,7 +718,7 @@ export class EventDetail extends React.Component<Props, State> {
         const activateButton = {
           title: 'Activate',
           onPress: this.activateSW,
-          secondary: true,
+          secondary: false,
```
graszka22, line 721: can't you just remove this whole line?

PR #2075, opened 2020-07-02
fix/fastlane-missing-gem

IAmKio:
- Attempting to mitigate the dependency issue where fastlane is looking for rake to install a gem.


.circleci/config.yml
```diff
@@ -70,6 +70,7 @@ aliases:
     run:
       name: Create Vendor/bundle for gem dependencies
       command: |
+        gem "rake"
```
antonioned, line 73: I don't think we need this at all @IAmKio

PR #2077, opened 2020-07-03
PIL-693 Add option to disable PIN on login

graszka22:
Additionally I disabled check if we have a backup of 12 words but **only in dev mode**. I often find myself disabling it temporarily.

### DIFF
--- a/src/screens/PinCodeUnlock/PinCodeUnlock.js
+++ b/src/screens/PinCodeUnlock/PinCodeUnlock.js
@@ -44,7 +44,7 @@ import {
 } from 'utils/keychain';
 
 import type { InitSmartWalletProps } from 'models/SmartWalletAccount';
-import omit from 'lodash.omit';
+
 
 const ACTIVE_APP_STATE = 'active';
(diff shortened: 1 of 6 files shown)


src/screens/PinCodeUnlock/PinCodeUnlock.js
```diff
@@ -88,7 +88,7 @@ class PinCodeUnlock extends React.Component<Props, State> {
     this.errorMessage = navigation.getParam('errorMessage', '');
     this.onLoginSuccess = navigation.getParam('onLoginSuccess', null);
 
-    if (!this.props.useBiometrics || navigation.getParam('forcePin')) {
+    if ((!this.props.useBiometrics || navigation.getParam('forcePin')) && !navigation.getParam('omitPin')) {
```
jsidorenko, line 91: I think forcePin has a higher priority on omitPin

jsidorenko: Sorry, your approach is correct

PR #2080, opened 2020-07-03
PIL-694 Fix no allowance tx record from Ethplorer

graszka22:


src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -344,7 +344,11 @@ export class ActivityFeedItem extends React.Component<Props> {
     } = this.props;
 
     const isReceived = this.isReceived(event);
-    const value = formatUnits(event.value, assetDecimals);
+    let value = formatUnits(event.value, assetDecimals);
+    if (event.transactionType === 'approve') {
+      // if it's approval transaction, the value is some rubbish, so we hide it from the user
+      value = '0';
```
poocart, line 350: Interesting, it should come as `0` already because it's data that does the thing in the transaction 🤔

jsidorenko: Not in ethplorer case :(

PR #2095, opened 2020-07-09
Delete connections code

jdnmr:
Delete connections, invitations, chats, transaction notes and everything related


src/actions/notificationsActions.js
```diff
@@ -268,40 +211,6 @@ export const subscribeToPushNotificationsAction = () => {
...
-        if (notification.message === MESSAGE_DISCONNECTED) {
-          dispatch(deleteChatAction(notification.title));
-        }
-
-        dispatch(fetchInviteNotificationsAction());
-      }
-      if (notification.type !== SIGNAL) {
-        dispatch({ type: ADD_NOTIFICATION, payload: notification });
-        dispatch({ type: SET_UNREAD_NOTIFICATIONS_STATUS, payload: true });
-      }
```
poocart, line 304: I think this should be kept.

jdnmr: Amazing catch, thanks!

PR #2104, opened 2020-07-14
[PIL-695] Feature/new receive modal

alazarevski:
Adds a new Receive Modal according to the new designs.
Removes Wallet switching in receive (the active address is the one in context)


src/screens/Asset/ReceiveModal.js
```diff
@@ -18,31 +18,52 @@
...
+import type { RootReducerState } from 'reducers/rootReducer';
+import type { Theme } from 'models/Theme';
+import type { User } from 'models/User';
+
+// constants
+import { ACCOUNT_TYPES } from 'constants/accountsConstants';
 
 const ContentWrapper = styled(SafeAreaView)`
-  padding: 0 ${spacing.layoutSides}px 60px;
+  padding: 0 ${spacing.layoutSides}px 50px;
```
vyvil, line 52: Seems like wrapper should have bigger paddings on the sides: https://app.zeplin.io/project/5db9b5d14c30ae56faaa3030/screen/5efdff1fae9030a1cd837873

src/screens/Asset/ReceiveModal.js
```diff
@@ -126,34 +180,61 @@ export default class ReceiveModal extends React.Component<Props, *> {
...
+          <ImageWrapper style={{ position: 'absolute', marginTop: -24 }}>
+            {!!profileImage && <ProfileImage
+              uri={`${profileImage}?t=${lastUpdateTime}`}
+              userName={username}
+              diameter={48}
+              borderWidth={0}
+              noShadow
+            />}
+            {!profileImage &&
+            <ProfileImagePlaceholder>
```
vyvil, line 193: Hm, `ProfileImage` handles placeholders by default if image is not preset.
See it's `DefaultPicture` component. If we need some changes in placeholders UI, maybe they are needed across the app? If not, maybe we can extend this component to accept placeholder's styling?

src/screens/Home/ActionButtons.js
```diff
@@ -159,19 +167,33 @@ class ActionButtons extends React.Component<Props, State> {
...
-          value: formattedBalance,
-          ...additionalInfo,
-          onPress: () => this.setState({ receiveAddress: address }),
-          label: `To ${additionalInfo.title}`,
-        }),
-        );
+        return [
+          {
+            key: 'buy',
+            label: 'Buy with card or Apple Pay',
```
vyvil, line 173: Hm, maybe we should show "Apple Pay" to iOS users only? :)

src/screens/Home/ActionButtons.js
```diff
@@ -279,11 +301,20 @@ class ActionButtons extends React.Component<Props, State> {
             />
           </ActionButtonsWrapper>
         </Sizer>
+        {visibleActionModal === RECEIVE &&
```
vyvil, line 304: Why double visibility setting is needed here? 
Wouldn't `isVisible={visibleActionModal === RECEIVE}` be enough?

src/screens/Home/ActionButtons.js
```diff
@@ -279,11 +301,20 @@ class ActionButtons extends React.Component<Props, State> {
...
+          <ActionOptionsModal
+            onModalClose={this.closeActionModal}
+            isVisible={!!visibleActionModal}
+            items={modalActions}
+          />
+        }
+        {visibleActionModal !== RECEIVE &&
```
vyvil, line 311: Same here :)

src/screens/Home/ActionButtons.js
```diff
@@ -159,19 +167,33 @@ class ActionButtons extends React.Component<Props, State> {
...
+            label: 'Exchange',
+            iconName: 'flip',
+            onPress: () => this.navigateToAction(activeWallet, EXCHANGE),
+          },
+          {
+            key: 'invite',
+            label: 'Invite and earn free tokens',
+            iconName: 'present',
+            hide: !rewardActive,
+            onPress: () => goToInvitationFlow(),
```
vyvil, line 194: Nit:
```suggestion
            onPress: goToInvitationFlow,
```

src/screens/Home/ActionButtons.js
alazarevski: well the Or clause is probably grammatically separating the possibilities, but yea might be an option if product team wants that (anyway will these strings be extracted for transaction ?)

src/components/ActionModal/ActionOptionsModal.js
```diff
@@ -0,0 +1,109 @@
...
+    return (
+      <SlideModal
+        isVisible={isVisible}
+        noClose
+        background={colors.card}
+        hideHeader
+        onModalHide={onModalClose}
+      >
+        <SafeAreaView>
+          <MainContainer>
```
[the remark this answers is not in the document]
alazarevski: i updated this, but should 'add funds to Pillar' be changed to something like 'Add funds to <your> Pillar Wallet' ?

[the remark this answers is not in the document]
vyvil: Hm, that's the question for @paragramm, I suppose :)

src/screens/Home/ActionButtons.js
vyvil: Yeah, they will be, eventually :).
Still, this will be going out before we merge in all translations. 
And on Android, in my opinion, it seems a bit weird to see 'Apple Pay' option :)

src/components/ActionModal/ActionOptionsModal.js
```diff
@@ -19,84 +19,76 @@
...
-  onPress: Function,
-  value: string,
-  isDisabled?: boolean,
-  iconName?: string,
-};
-
-
 const MainContainer = styled.View`
-  padding: 20px ${spacing.layoutSides}px 30px;
+  padding: 22px ${spacing.layoutSides}px 50px;
```
vyvil, line 48: Seems like horizontal paddings are still too large. 
It's 30px in design.
SlideModal has a 16px padding by default. So either here horizontal padding should be set to 0, and ListItem horizontal paddings should be set to 14px or wise versa :)

As for bottom padding - designs are showing "safe space" padding (as they are designed on iPhone). In action we do not need such a long padding - it is being handled by SafeArea. (You can check how we are handling this on EventDetails modal :) )

src/components/ActionModal/ActionOptionsModal.js
```diff
@@ -0,0 +1,101 @@
...
+
+    return (
+      <SlideModal
+        isVisible={isVisible}
+        noClose
+        hideHeader
+        onModalHide={onModalClose}
+      >
+        <SafeAreaView>
+          <BaseText style={{ paddingTop: 24, fontWeight: '500' }} center big>{title}</BaseText>
```
vyvil, line 91: Since `title` is optional (based on types), what do you think about conditional rendering here? :)

Also, `fontWeight` as a style isn't working here. We are using `MediumText` component for 500 weight texts.

src/screens/Asset/Asset.js
```diff
@@ -292,6 +297,61 @@ class AssetScreen extends React.Component<Props, State> {
...
+    const fiatCurrency = baseFiatCurrency || defaultFiatCurrency;
+    const { token } = assetData;
+    const balance = getBalance(balances, token);
+    const isSupportedByExchange = exchangeSupportedAssets.some(({ symbol }) => symbol === token);
+    return [
+      {
+        key: 'buy',
+        label: `Buy with a card${Platform.OS === 'ios' ? ' or Apple Pay' : ''}`,
+        iconName: 'wallet',
+        onPress: () => this.goToExchangeFlow(fiatCurrency, token),
```
vyvil, line 319: Maybe it should also be check if the asset is supported by the exchange (as you did for 'Exchange' option)?

alazarevski: I don't know if these fiat-crypto assets are in the same group since they rely on external fiat providers

PR #2107, opened 2020-07-15
PIL-700 Add Ramp.Network / Wyre to Services screen

wjkg:


src/screens/Exchange/Exchange.js
```diff
@@ -343,57 +338,34 @@ class ExchangeScreen extends React.Component<Props, State> {
...
-    const fromAsset = fiatCurrencies.find(currency => currency.symbol === fromAssetCode)
-      || getAssetData(assetsData, exchangeSupportedAssets, fromAssetCode);
-    const selectedAssetOptions = isFiatCurrency(fromAssetCode)
-      ? this.generateFiatOptions().find(({ symbol }) => symbol === fromAssetCode)
-      : this.generateAssetsOptions({ [fromAssetCode]: fromAsset })[0];
-    const initialFormState = {
-      ...this.state.value,
-      fromInput: {
+
+    if (!isFiatCurrency(fromAssetCode)) {
```
jdnmr, line 359: Again - do we need this check, if there is no fiat support anymore?

src/screens/Exchange/Exchange.js
```diff
@@ -672,11 +635,9 @@ class ExchangeScreen extends React.Component<Props, State> {
       && smartWalletStatus.status !== SMART_WALLET_UPGRADE_STATUSES.ACCOUNT_CREATED
       && !deploymentData.error;
 
-    const isSelectedFiat = !isEmpty(selectedFromOption) &&
-      fiatCurrencies.some(({ symbol }) => symbol === selectedFromOption.symbol);
+    const isSelectedFiat = !isEmpty(selectedFromOption) && isFiatCurrency(selectedFromOption.symbol);
```
jdnmr, line 638: This will always return false, right? So maybe we don't need this check at all?

src/screens/Exchange/ExchangeOffers.js
```diff
@@ -550,10 +502,7 @@ class ExchangeOffers extends React.Component<Props, State> {
       || !allowanceSet
       || (isShapeShift && !shapeshiftAccessToken);
 
-    const isFiat = isFiatProvider(offerProvider);
-
-    const disableFiatExchange = isFiat && (minOrMaxNeeded || !!offerRestricted);
-    const disableOffer = disableNonFiatExchange || disableFiatExchange;
+    const disableOffer = disableNonFiatExchange;
```
jdnmr, line 505: We can remove the `disableOffer` const now, just use `disableNonFiatExchange ` below

src/screens/Services/Services.js
```diff
@@ -47,20 +56,19 @@ import { isActiveAccountSmartWalletSelector, isSmartWalletActivatedSelector } fr
...
 
 type Props = {
   theme: Theme,
   providersMeta: ProvidersMeta,
   navigation: NavigationScreenProp<*>,
-  baseFiatCurrency: ?string,
   getMetaData: () => void,
   isActiveAccountSmartWallet: boolean,
   isSmartWalletActivated: boolean,
+  user: Object,
```
jdnmr, line 68: We have a `User` type under `models/User`

src/screens/Services/Services.js
```diff
@@ -142,6 +132,46 @@ class ServicesScreen extends React.Component<Props> {
     ];
   };
 
+  getBuyCryptoServices = () => {
+    const {
+      user: { email = '' },
+      accounts,
+    } = this.props;
+
+    const address = getActiveAccountAddress(accounts);
```
jdnmr, line 141: Just a thought: if we're about to drop KW soon, maybe we should enforce all fiat-bought tokens to settle on SW? Any thoughts on this @IAmKio  ?

IAmKio: I agree, Smart Wallet only going forward - if they haven't activated their smart wallet yet then they should be prompted to do that first.

src/screens/Services/Services.js
```diff
@@ -142,6 +133,46 @@ class ServicesScreen extends React.Component<Props> {
     ];
   };
 
+  getBuyCryptoServices = () => {
+    const {
+      user: { email = '' },
```
IAmKio, line 138: Is it better to put `null` instead of an empty string? Whilst `''` is falsey, some tools or services might result in this being _something_, instead of nothing...

src/utils/inAppBrowser.js
```diff
@@ -0,0 +1,46 @@
...
+export const openInAppBrowser = async (url: string, options?: InAppBrowserOptions) => {
+  try {
+    if (await InAppBrowser.isAvailable()) {
+      await InAppBrowser.open(url, options);
+    } else {
+      reportOrWarn('InAppBrowser.isAvailable() returned false', null, 'warning');
+    }
+  } catch (error) {
+    reportOrWarn('InAppBrowser.error', error, 'warning');
+  }
```
IAmKio, line 45: We have an `async` function here but we're invoking the use of a `try` / `catch` - i'm just wondering is there a scenario here where `await InAppBrowser.isAvailable()`  and `await InAppBrowser.open(url, options)` `.catch` methods would not suffice? Do we have an uncatchable error here?

src/actions/exchangeActions.js
```diff
@@ -205,91 +202,59 @@ export const searchOffersAction = (fromAssetCode: string, toAssetCode: string, f
...
-          type: 'warning',
-          message: 'Could not find asset',
-        });
-        return;
-      }
+    if (!fromAsset || !toAsset) {
+      Toast.show({
+        title: 'Exchange service failed',
+        type: 'warning',
+        message: 'Could not find asset',
```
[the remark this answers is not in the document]
wjkg: It's not apparent from the diff, but the only modification in this larger chunk that was done by me is indentation. I can change it, but I don't think I should do random refactoring -- AFAIK it's unrelated and there's work planned for exchanges anyway?

src/actions/exchangeActions.js
```diff
@@ -205,91 +202,59 @@ export const searchOffersAction = (fromAssetCode: string, toAssetCode: string, f
...
+      const message = responseErrorMessage || 'Unable to connect';
+      if (message.toString().toLowerCase().startsWith('access token')) {
+        /**
+         * access token is expired or malformed,
+         * let's hit with user info endpoint to update access tokens
+         * or redirect to pin screen (logic being sdk init)
+         * after it's complete (access token's updated) let's dispatch same action again
+         * TODO: change SDK user info endpoint to simple SDK token refresh method when it is reachable within SDK
+         */
+        await api.userInfo(userWalletId).catch(() => null);
```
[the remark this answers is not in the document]
wjkg: (same as https://github.com/pillarwallet/pillarwallet/pull/2107/files/4abbd8fc909521f014e8453c97de7420693b6279#r455722157)

src/actions/exchangeActions.js
```diff
@@ -205,91 +202,59 @@ export const searchOffersAction = (fromAssetCode: string, toAssetCode: string, f
...
+      } else {
+        Toast.show({
+          title: 'Exchange service failed',
+          type: 'warning',
+          message,
+        });
```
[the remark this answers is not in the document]
wjkg: (same as https://github.com/pillarwallet/pillarwallet/pull/2107/files/4abbd8fc909521f014e8453c97de7420693b6279#r455722157)

src/actions/exchangeActions.js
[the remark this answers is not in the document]
IAmKio: Whilst i understand that this is not related to the work that you're doing and that you only changed the indentation, it has highlighted an improvement that the company are attempting to make across the app which is improving all error messages / user facing messages. I would imagine that any lines of code that are changed, the developer would take ownership of the changed lines of code.

I'll leave this one up to you to decide.

PR #2108, opened 2020-07-16
Feature/translations

vyvil:
This PR initialises translation mechanism and applies translated strings to onboarding and auth flows.

For now locally stored jsons are used for storing translation strings.

Later on it will be fetched. Most probably from s3.
Currently I'm using 'auth' namespace for translations so we could fetch those first and then fetch other translations after translation is initialised.

If everything is okay, it's safe to merge it in as even if language detection plugin detects other than 'en' language, app will still be using 'en' as it is the only "supported language" so far :).

P.S.
When adding new translations, it's important to stick to common string identifier naming convention.
As translations are still in works it might change, but for now I'm sticking to those principles:

1. DESCRIPTIVENESS: Identifiers should be as descriptive as possible in a sense that they describe the purpose of the string rather than a meaning of it (as it might change over time). 
(Unless it's some funky, trademarked, reusable phrase)
For example, if in designs we see title "Make crypto easy" in "Biometrics prompt screen", it's better to use `'title.biometricsPromptOnLogin'` as an identifier rather than `'title.makeCryptoEasy'` as later on we would might change it to something like 'Wanna just touch the phone and get logged in?' or even 'I don't usually hate Pins, but when I do...'.
And in the case of "funkiness" button's "Magic" translations I've referenced as `button.magic`.
The takeaway should be this, that translation values should be treated like _CMS content_ as eventually, translations will be fetched from a server.

2. NAMESPACES: i18next allows to load separate translation files based on name spaces. That's cool, but we can have namespaces inside those jsons as well, to group translations by type (that not only provide a structure, but also allow to have some "formatting" rules over the strings. For example - we could decide that all strings, that are put in title group (e.g. `title.string`) should start with a capital letter and end without a dot. While paragraph group strings should start in capital letter AND end with a dot / ? / ! (some sort of punctuation).
2.1. Namespaces can also be used for "features" to show strings' uniquenesses and context so we would know that it should solely be used in a contexts of the feature (so even if some things change, text would only change for that feature).

I've also added some postprocessors. One is dedicated for managing question marks and exclamation points (if we would need to add one to reusable string). Since those punctuations are treated differently in Spanish and in Arabic, for example. So it can't be added just like that: `${t('label.forgot')?}`;

But it could also be solved by adding a separate translation sting that could have a tag. For example:
```
"label": {
  "forgot": "forgot"
  "forgot_question": "forgot?"
}
```
As well as suffixes and prefixes (..., for example). What is suffix in 'en' (Loading...) is a prefix in Arabic (...جار التحميل)

Feel free to suggest any changes or rules :)
(Will put them in readme when it will be settled)


App.js
```diff
@@ -217,23 +225,20 @@ class App extends React.Component<Props, *> {
...
                   const themeToChangeTo = current === LIGHT_THEME ? DARK_THEME : LIGHT_THEME;
                   setAppTheme(themeToChangeTo);
                 }}
-              >
-                <Text style={{ color: colors.text }}>{`THEME: ${current}`}</Text>
-              </TouchableOpacity>}
+              />}
+              {!!SHOW_LANG_TOGGLE && <Button
+                title={`Change lang (current: ${i18n.language})`}
+                onPress={() => changeLanguage(i18n.language === 'fr' ? 'en' : 'fr')}
```
graszka22, line 240: is it some kind of leftover from testing?

__mocks__/react-native-localize.js
```diff
@@ -0,0 +1,47 @@
...
+const getNumberFormatSettings = () => ({
+  decimalSeparator: '.',
+  groupingSeparator: ',',
+});
+
+const getCalendar = () => 'gregorian';
+const getCountry = () => 'GB';
+const getCurrencies = () => ['GBP', 'EUR'];
+const getTemperatureUnit = () => 'celsius';
+const getTimeZone = () => 'Europe/London';
```
graszka22, line 32: so all users will get dates in London time zone and not their own? 🤔

vitorpy: Well, that's the time standard!

.env
```diff
@@ -40,3 +40,4 @@ DAI_ADDRESS=0x4F96Fe3b7A6Cf9725f59d353F723c1bDb64CA6Aa
 USDC_ADDRESS=0x75B0622Cec14130172EaE9Cf166B92E5C112FaFF
 POOLTOGETHER_GRAPH_ID=QmSmtdcYD9Zuk7W9zuk73rRCNema26vMdF7muyfGwLzqfT
 THE_GRAPH_API_URL=https://api.thegraph.com/subgraphs/id
+SHOW_LANG_TOGGLE=
```
IAmKio, line 43: I'm assuming this is a boolean value, should we put at least _something_ here?

App.js
IAmKio: Looks like it could be...

environments/.env.production
```diff
@@ -40,3 +40,4 @@ DAI_ADDRESS=0x6B175474E89094C44Da98b954EedeAC495271d0F
 USDC_ADDRESS=0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48
 POOLTOGETHER_GRAPH_ID=QmSmtdcYD9Zuk7W9zuk73rRCNema26vMdF7muyfGwLzqfT
 THE_GRAPH_API_URL=https://api.thegraph.com/subgraphs/id
+SHOW_LANG_TOGGLE=
```
IAmKio, line 43: Shall we assume this to be `true`?

environments/.env.staging
```diff
@@ -40,3 +40,4 @@ DAI_ADDRESS=0x4F96Fe3b7A6Cf9725f59d353F723c1bDb64CA6Aa
 USDC_ADDRESS=0x75B0622Cec14130172EaE9Cf166B92E5C112FaFF
 POOLTOGETHER_GRAPH_ID=QmSmtdcYD9Zuk7W9zuk73rRCNema26vMdF7muyfGwLzqfT
 THE_GRAPH_API_URL=https://api.thegraph.com/subgraphs/id
+SHOW_LANG_TOGGLE=
```
IAmKio, line 43: Same as above :)

src/components/ProfileForm/profileFormDefs.js
```diff
@@ -11,49 +13,49 @@ const maxLength = 100;
...
   return isValidPhoneWithoutCountryCode(input);
 });
 
 EmailStructDef.getValidationErrorMessage = (email): string => {
   if (email && !isValidEmail(email)) {
-    return 'Please, check your email address. It may contain only latin letters (a-z), numbers (0-9) and dot (.)';
+    return t('auth:invalidEmailAddress');
   } else if (email && email.length > maxLength) {
-    return `Email should not be longer than ${maxLength} symbols`;
+    return t('auth:invalidEMailAddress_tooLong', { requiredLength: maxLength });
```
IAmKio, line 51: The mixture of camelCase and snake case bothers me a bit here 😆

src/translations/changeLanguage.js
```diff
@@ -0,0 +1,43 @@
...
+    .then(() => {
+      Toast.show({
+        message: 'Language has been changed',
+        type: 'info',
+        title: '',
+      });
+    })
+    .catch(() => {
+      Toast.show({
+        message: 'Could not change language',
```
IAmKio, line 36:
```suggestion
        message: 'Sorry, there was an issue changing language. Please try again later.',
```

src/translations/deviceLanguageDetector.js
```diff
@@ -0,0 +1,43 @@
...
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import * as RNLocalize from 'react-native-localize';
+import { SUPPORTED_LANGUAGES, DEFAULT_LANGUAGE } from './config';
+
+
```
IAmKio, line 24: I know it's really minor but there's two lines here instead of one - we'll catch this in future with stronger linting rules.

src/translations/postProcessors.js
```diff
@@ -0,0 +1,65 @@
...
+  type: 'postProcessor',
+  name: PUNCTUATION_POSTPROCESSOR,
+  process: (value: string, key: string, options: TranslationOptions) => {
+    if (options.exclamation || options.questionMark) {
+      const exclamation = options.exclamation ? '!' : '';
+      const questionMark = options.questionMark ? '?' : '';
+      if (i18n.language === 'es') {
+        const invertedExclamation = options.exclamation ? '¡' : '';
+        const invertedQuestionMark = options.questionMark ? '¿' : '';
+        return `${invertedExclamation}${invertedQuestionMark}${value}${exclamation}${questionMark}`;
```
IAmKio, line 36: ¡Love it! ❤️

App.js
vyvil: Not really, it's left there intentionally. This works just like a "theme" toggle - tool for easier development and it's visibility is managed using env variables. 
It renders language switcher button at the bottom of the screen, so it would be easier to check if translations are working in each single screen by simply toggling between two languages.
'en' - 'fr' pair is picked because French is a second language for us to go :)

__mocks__/react-native-localize.js
vyvil: As much as I agree with Vitor on this one, this is merely a mock for tests (as react-native-localize is a native module), so it won't override user settings :)

.env
vyvil: ENV variables are treated like strings, so if we set it to `false` here, we'll get it as `'false'` which ends up to be truthy value.
I'm leaving it empty so it would be `undefined`. 
(And I'm leaving it empty on dev as well, as it is not always needed - just when we would be working on translations)

environments/.env.production
vyvil: Nope, it should be false by default, as SHOW_LANG_TOGGLE manages development tool to toggle languages quickly on each screen :)

src/components/ProfileForm/profileFormDefs.js
vyvil: Aah, that's a mistype :)) will fix that shortly. Sorry for any disturbance that this might caused 😊

vyvil:
> > Later on it will be fetched. Most probably from s3.
> 
> I don't get it why we should fetch those files from for example s3. Translation files are small, comparable to the smallest images and we have no problem with keeping the images locally. Fetching a couple of files at app startup doesn't seem like a good idea in terms of performance. And there will be some versioning issues that will need to be solved. The only gain I can see is that the user uses only one language and doesn't need other languages. But as I said before, translations are just text files so they're fairly small and we won't support dozens of languages, right?

Well, there's a business need to be able to manage content externally (to have a content management system of some sort).
That's why eventually we'll be fetching translations from a server.

And true, versioning is an obstacle that we would need to overcome in order to move to external translation files. (For cases when texts are being changed drastically).
Yet it could be done using i18next's backend plugin, that supports versioning using query params:
https://github.com/i18next/i18next-http-backend#backend-options (`queryStringParams`)

__


> Another issue are identifiers with underscores - wouldn't it be simpler to make just a nested object? An example:
> 
> ```js
> "invalidUsername_cantStartWithNumber": "Username can not start with a number",
>  "invalidUsername_cantStartEndWithDash": "Username can not start or end with a dash",
>  "invalidUsername_useAlphanumericSymbolsOnly": "Only use alpha-numeric characters or dashes",
> ```
> 
> isn't this better:
> 
> ```js
> "invalidUsername": {
>    "cantStartWithNumber": "Username can not start with a number",
>     "cantStartEndWithDash": "Username can not start or end with a dash",
>     "useAlphanumericSymbolsOnly": "Only use alpha-numeric characters or dashes",
> }
> ```

Yeah, I agree - nesting is better. I've added underscore so I could use default `error.invalidUsername` value, but I could add it as `error.invalidUsername.default`. Will adjust it. Thanks! :)

src/translations/deviceLanguageDetector.js
vyvil: We are usually leaving two empty lines after imports - to separate content from imports.
It's kind of what we have agreed on (more like, no one was against it :) ) couple of months ago: https://pillarwallet.slack.com/archives/CEEV48867/p1583529140199300

__mocks__/react-native-localize.js
graszka22: oh my bad sorry 😅

vyvil: No worries! Always better to double check, in my opinion :)

android/app/build.gradle
```diff
@@ -239,6 +239,7 @@ android {
 }
 
 dependencies {
+    implementation project(':react-native-localize')
```
poocart, line 242: Not needed?

android/app/src/main/java/com/pillarproject/wallet/MainApplication.java
```diff
@@ -4,6 +4,7 @@
 import androidx.multidex.MultiDexApplication;
 import com.facebook.react.PackageList;
 import com.facebook.react.ReactApplication;
+import com.reactcommunity.rnlocalize.RNLocalizePackage;
```
poocart, line 7: Not needed?

src/components/RetryApiRegistration/RetryApiRegistration.js
```diff
@@ -60,7 +65,7 @@ class RetryApiRegistration extends React.Component<Props> {
     return (
       <Container center>
         {walletState !== REGISTRATION_FAILED && (
-          <Loader messages={['Registering on backend']} />
+          <Loader messages={[t('auth.loadingMessage.registering')]} />
```
poocart, line 68: Why one call starts with "auth:" and this "auth."? Maybe all can be just "auth." if both works?

src/models/Translations.js
```diff
@@ -0,0 +1,42 @@
...
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+import { Trans } from 'react-i18next';
+import { BaseText } from 'components/Typography';
+
+export type I18n = {
+  language: string,
+}
```
poocart, line 25:
```suggestion
};
```

src/navigation/rootNavigation.js
```diff
@@ -71,6 +62,10 @@ import {
 
 import AppFlow from './appNavigation';
 
+type Props = {
+  language: string,
+}
```
poocart, line 67:
```suggestion
};
```

src/translations/translationLoader.js
```diff
@@ -0,0 +1,51 @@
...
+*/
+
+import type { TranslationResource } from 'models/Translations';
+
+
+// en
+const COMMON_EN = require('./locales/en/common.json');
+const AUTH_EN = require('./locales/en/auth.json');
+
+
```
poocart, line 28:
```suggestion
```

src/utils/strings.js
```diff
@@ -0,0 +1,23 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+export const capitalize = (str: string) => {
+  return str.charAt(0).toUpperCase() + str.slice(1);
```
poocart, line 22: Would add "str" type check if it's string just in case, leaving it up to you :D

App.js
IAmKio: Apologies - my misunderstanding!

.env
IAmKio: Ahh forgot about the ENV strings!

environments/.env.production
IAmKio: Thanks!

src/components/ProfileForm/profileFormDefs.js
IAmKio: It's not me it's me OCD 😂

src/translations/changeLanguage.js
IAmKio: Thank you!

src/translations/deviceLanguageDetector.js
IAmKio: No problem :) thanks for letting me know!

android/app/build.gradle
vyvil: True!

PR #2110, opened 2020-07-20
Feature/launcher splash screen update [PIL-745]

vyvil:
This PR updates launcher icons, splash screen and Pillar logos in relation to Pillar's rebranding.

It also adjusts Welcome screen (and since this is the only screen that is being updated according to the new branding, I'm just slightly adjusting UI elements and not applying color changes to the components themselves).


ios/pillarwallet/Info.plist
```diff
@@ -81,7 +81,7 @@
 		<string>remote-notification</string>
 	</array>
 	<key>UILaunchStoryboardName</key>
-	<string>LaunchScreen</string>
+	<string>Launch Screen Pixel</string>
```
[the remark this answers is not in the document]
graszka22: @vyvil it was really a cache issue,  I've restarted the simulator and I can see the new splash screen 👍

src/screens/Splash/Splash.js
```diff
@@ -0,0 +1,113 @@
...
+    const { splashPatternTop, splashPatternBottom } = images(theme);
+    const imgResizePercentage = BIGGEST_PATTERN_HEIGHT > MAX_PATTERN_HEIGHT ? getResizePercentage() : 1;
+
+    return (
+      <BackgroundWrapper>
+        <PatternImg
+          source={splashPatternTop}
+          resizeMode="cover"
+          height={BIGGEST_PATTERN_HEIGHT * imgResizePercentage}
+          width={216 * imgResizePercentage}
```
graszka22, line 88: pls don't put numbers like that in the middle of the code, I have a hard time later wondering "what does that number mean"

graszka22:
and it's just a wild idea but can we use the animation time to start loading something in the background? forcing the user to wait for 2 seconds just to show him the animation seems like a bad idea so at least let's use that time productively :D 
for example when mounting the Home screen we do a bunch of requests:
```js
if (isSmartWalletActive) {
      fetchPoolStats(true);
    }
    fetchTransactionsHistory();
    fetchBadges();
    fetchBadgeAwardHistory();
    fetchReferralRewardsIssuerAddresses();
    fetchDepositedAssets();
```
and the app is sluggish because of that during the first seconds
but I realize that might be difficult to do or out of scope of this PR

src/screens/Splash/Splash.js
vyvil: Hm, this is an actual image width multiplied by explained value - `imgResizePercentage`.
You mean, you'd like to have a const, for example, `adjustedImageWidth` instead of this?
Or I could add a note explaining why we need `imgResizePercentage` here.
Which one you'd prefer?

vyvil:
> android and ios are a bit different
> on iOS the logo goes up and down and to the next screen
> on Android the logo goes up, down, up and then to the next screen
> 

I do not notice "jumping" of the logo in the splash screens - they are positioned in the center of the screen.
While the logo on Welcome screen is (and is supposed to be) slightly off.

(from a PIL-745 description: 
Onboarding start
Show upon fresh app install/launch. Notice the logo is slightly offset vertically for -20px to the top.

Splash
Show upon regular app launch. The logo is strictly vertically centered.)

> also between the launch screen and animated splash screen there is a bit of difference, the squares jump a little
> on Android additionally the screen flickers
> I think that the idea was to make this transition smooth as if it was one splash screen
> 

Yeah, squares must be jumping because of this resizing (to avoid overlapping the logo). I see what I can do about it.
Thanks! :)

src/screens/Splash/Splash.js
graszka22: I'd like a const saying that 216 is the image width. Just 216 in the middle of the code doesn't tell me anything. Then I can deduce "aha, so we scale the image with by some percentage, ok!"

vyvil:
> and it's just a wild idea but can we use the animation time to start loading something in the background? forcing the user to wait for 2 seconds just to show him the animation seems like a bad idea so at least let's use that time productively :D
> for example when mounting the Home screen we do a bunch of requests:
> 
> ```js
> if (isSmartWalletActive) {
>       fetchPoolStats(true);
>     }
>     fetchTransactionsHistory();
>     fetchBadges();
>     fetchBadgeAwardHistory();
>     fetchReferralRewardsIssuerAddresses();
>     fetchDepositedAssets();
> ```
> 
> and the app is sluggish because of that during the first seconds
> but I realize that might be difficult to do or out of scope of this PR

Great idea! 
If I'm not mistaken, it's in the plans to optimise the way we fetch initial data :)
I think we could address this then :)

graszka22:
>> android and ios are a bit different
on iOS the logo goes up and down and to the next screen
on Android the logo goes up, down, up and then to the next screen

>I do not notice "jumping" of the logo in the splash screens - they are positioned in the center of the screen.

Yes the logo is centered, I've meant the animation of going down and up. So in other words:
on iOS the logo appears, disappears and it goes to the next screen
on Android the logo appears, disappears, appears again and then it goes to the next screen
does the animation on Android take longer if it can show 3 steps of animation?

vyvil:
> Yes the logo is centered, I've meant the animation of going down and up. So in other words:
> on iOS the logo appears, disappears and it goes to the next screen
> on Android the logo appears, disappears, appears again and then it goes to the next screen
> does the animation on Android take longer if it can show 3 steps of animation?

This might be because the device takes longer to execute the method / fetch data.
I'm seeing only one iteration of animation on my Android emulator at the moment :)

PR #2118, opened 2020-07-23
Feature/animated splash screen [PIL-754]

vyvil:
This PR adds animated splash screen


graszka22:
👀 

Screenshot 2020-07-23 at 15 27 54

vyvil:
> 👀
> 
> Screenshot 2020-07-23 at 15 27 54

Thought that I've replaced it ;/ Should be good now :)

PR #2127, opened 2020-07-29
Feature/translations assets

vyvil:
PR to translate Assets screen and related components
(content based translation indentifiers are used here)


IAmKio:
Love it

PR #2132, opened 2020-08-03
PIL-699 Add Sablier integration

graszka22:
How to test: it works out of the box on ropsten network with *any* ERC-20 token :3 so testing it is a breeze

Things left to do:
- [ ] badges
- [x] test it on mainnet


src/screens/Sablier/Withdraw.js
```diff
@@ -0,0 +1,222 @@
...
+        preselectedAsset={assetData.symbol}
+        customOptions={assetOptions}
+        balances={streamedAssetBalance}
+        baseFiatCurrency={baseFiatCurrency}
+        rates={rates}
+        maxLabel="Max"
+        getFormValue={onValueChanged}
+        isLoading={isFetchingMaxWithdraw}
+      />
+      <BaseText regular secondary center>You will receive {assetData.symbol} in your wallet.</BaseText>
```
IAmKio, line 188: Just a note - we're moving away from the language of referring to the app as a "wallet". Nothing to action here, but i just wanted to let you know.

src/screens/Sablier/WithdrawReview.js
```diff
@@ -0,0 +1,184 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import * as React from 'react';
```
IAmKio, line 21: Could just copy and past from the others here.

src/screens/Services/Services.js
```diff
@@ -164,6 +169,14 @@ class ServicesScreen extends React.Component<Props, State> {
...
       });
     }
+    services.push({
+      key: 'sablier',
+      title: 'Sablier money streaming',
+      body: 'Stream money to people and organizations in real-time with just one deposit',
+      disabled: SWServiceDisabled,
+      label: SWServiceLabel,
+      action: () => navigation.navigate(SABLIER_STREAMS),
+    });
```
IAmKio, line 179: Could you please wrap this in `firebaseConfig.getBoolean('feature_services_sablier')` please? This will allow us to toggle this feature remotely.

src/services/sablier.js
```diff
@@ -0,0 +1,323 @@
...
+    startTimestamp,
+    endTimestamp,
+  ]);
+};
+
+export const fetchStreamBalance = (streamId: string, address: string): Promise<EthersBigNumber> => {
+  const provider = getEthereumProvider(NETWORK_PROVIDER);
+  const contract = new Contract(SABLIER_CONTRACT_ADDRESS, SABLIER_ABI, provider);
+  return contract.balanceOf(streamId, address).catch((e) => {
+    reportLog('Error getting sablier stream balance', {
```
IAmKio, line 66: Thanking you 👍

src/services/sablier.js
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  return {
+    gasToken,
+    txFeeInWei,
+    transactionPayload,
+  };
+};
+
+export const getApproveFeeAndTransaction = async (assetData: Asset, useGasToken: boolean): Promise<Object> => {
```
IAmKio, line 250: We're `async`ing here but no `await` - is this not supposed to be here, maybe?

src/services/sablier.js
```diff
@@ -0,0 +1,323 @@
...
+        symbol,
+      },
+    },
+    tag: SABLIER_ALLOWANCE,
+  };
+
+  return getTxFeeAndTransactionPayload(transactionPayload, useGasToken);
+};
+
+export const getCancellationFeeAndTransaction = async (stream: Stream, useGasToken: boolean): Promise<Object> => {
```
IAmKio, line 275: We have another `async` but no `await`ing...

src/services/sablier.js
```diff
@@ -0,0 +1,323 @@
...
+      amount: stream.deposit,
+      contactAddress: stream.recipient,
+      streamId: stream.id,
+    },
+  };
+
+  return getTxFeeAndTransactionPayload(transactionPayload, useGasToken);
+};
+
+export const getCreateStreamFeeAndTransaction = async (
```
IAmKio, line 296: As above - the same `async` but no `await`ing - could be due to copy/pasta

src/services/sablier.js
```diff
@@ -0,0 +1,323 @@
...
+  sender: string,
+  receiver: string,
+  amount: EthersBigNumber,
+  tokenAddress: string,
+  startTimestamp: number,
+  endTimestamp: number,
+  useGasToken: boolean,
+): Promise<Object> => {
```
IAmKio, line 304: Oh i've just seen this - but does this need to be a promise? Maybe error control? Please let me know if i've misunderstood something here.

src/actions/historyActions.js
```diff
@@ -87,11 +88,13 @@ import { syncVirtualAccountTransactionsAction } from './smartWalletActions';
 import { checkEnableExchangeAllowanceTransactionsAction } from './exchangeActions';
 import { checkPoolTogetherApprovalTransactionAction } from './poolTogetherActions';
 import { extractEnsInfoFromTransactionsAction } from './ensRegistryActions';
+import { checkSablierApprovalTransactionAction } from './sablierActions';
 
-const afterHistoryUpdatedAction = () => {
+export const afterHistoryUpdatedAction = () => {
   return async (dispatch: Dispatch) => {
```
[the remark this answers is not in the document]
graszka22: Hmm I think it's more of a convention in our code than it's actually needed. It's used in most places where we have asynchronous actions (actions that return function) even though we have no awaiting. In the redux-thunk docs they don't use it.

src/actions/historyActions.js
```diff
@@ -143,7 +146,8 @@ export const fetchSmartWalletTransactionsAction = () => {
       relayerExtensionDevice?.address,
     );
     const aaveHistory = await mapTransactionsHistoryWithAave(accountAddress, smartWalletTransactionHistory);
-    const history = await mapTransactionsPoolTogether(accountAddress, aaveHistory);
+    const poolTogetherHistory = await mapTransactionsPoolTogether(accountAddress, aaveHistory);
+    const history = await mapTransactionsHistoryWithSablier(accountAddress, poolTogetherHistory);
```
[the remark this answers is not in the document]
graszka22: The history that comes from the backend are just plain transactions. We don't know whether those transactions came from sablier, pool together, aave or are just plain transactions. So we have those "map" functions to add this information from different sources to the transactions.

src/screens/Sablier/IncomingStream.js
```diff
@@ -0,0 +1,199 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import * as React from 'react';
```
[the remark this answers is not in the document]
graszka22: Is there any difference in performance between `import * as React from 'react'` and `import React from 'react'`? I think in both cases we're just importing React. And React must be imported if we use JSX syntax, if not we'd get an error. What do you mean by optimisation here?

src/screens/Sablier/NewStream.js
```diff
@@ -0,0 +1,510 @@
...
+const ContentWrapper = styled.View`
+  padding: 0 20px;
+`;
+
+const PickerWrapper = styled.View`
+  padding: 18px 20px;
+`;
+
+const START_TIME = 'START_TIME';
+const END_TIME = 'END_TIME';
```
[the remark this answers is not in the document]
graszka22: They're constants used just locally for this file for modals so there is no need to move them into the constants file. There is no way we'd have to import it anywhere.

src/screens/Sablier/Withdraw.js
```diff
@@ -0,0 +1,222 @@
...
+
+  useEffect(() => {
+    // the exact max amount to withdraw must be fetched from the blockchain
+    // as we can only estimate due to the fact that the contract uses block.timestamp
+    fetchStreamBalance(streamId, recipient)
+      .then((streamBalance) => {
+        setMaxWithdraw(streamBalance);
+        setIsFetchingMaxWithdraw(false);
+      })
+      .catch(() => {});
```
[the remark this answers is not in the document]
graszka22: We catch and report the error in the `fetchStreamBalance` function. The `catch` here is just to silence eslint. And I can't make this function async or return the promise because we're in `useEffect` hook.  But I'll silence the linter in proper way by `// eslint-disable-next-line` so it's not confusing

src/services/sablier.js
graszka22: I've removed `async`s but the promise is correct: `getTxFeeAndTransactionPayload` is `async` so it returns the promise and so these functions also return promises

IAmKio: Got it - thank you.

src/actions/historyActions.js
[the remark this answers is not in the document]
IAmKio: Thanks for that explanation. In the near future, we will eventually get this data pre-processed and decorated from the backend APIs.

src/screens/Sablier/NewStream.js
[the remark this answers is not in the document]
IAmKio: 👍

src/screens/Sablier/IncomingStream.js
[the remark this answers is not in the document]
IAmKio: Actually I'm not sure what else is imported where when you use `*` - but I was thinking that it would be more than just `React` into the module. My thinking here was to decrease the memory footprint by only including the modules needed from `react` as opposed to _everything_..

PR #2145, opened 2020-08-05
Exchange cleanup

jdnmr:
- Keep exchange providers' metadata locally;
- Delete exchange service (as the above was the last use for it);
- Delete ShapeShift support, obsolete constants, redux fields etc.


src/screens/Exchange/Exchange.js
```diff
@@ -597,7 +589,7 @@ class ExchangeScreen extends React.Component<Props, State> {
 
     const formStructure = generateFormStructure(balances);
     const rightItems = [{ label: 'Support', onPress: () => Intercom.displayMessenger(), key: 'getHelp' }];
-    if ((!isEmpty(exchangeAllowances) || !isEmpty(connectedProviders))
+    if ((!isEmpty(exchangeAllowances))
```
IAmKio, line 592: Is there an extra left bracket there?

jdnmr:
> There is a little UX bug:
> 
> * sign out from the app
> * log in again
> * go to exchange
> * a following screen appears, user can't select anything
> 
> Screenshot 2020-08-11 at 13 21 53
> 
> * after a while exchanges appear
>   maybe it's a matter of just adding a loading spinner or something like that?
> 
> Also I can't get 1inch exchanges to appear on ropsten. Is it ok?
> 
> Otherwise LGTM 👍

The UX bug is caused by Uniswap taking forever to fetch initial assets - I will add a loading spinner.
As of 1inch offers on Ropsten, very few exist (ETH => SAI is an example). In the previous implementation, we displayed Homestead offers, but it was a hack and the exchanges themselves didn't work. Now we see a very limited set of offers, but they work - at least for Uniswap.
I found that old "feature" rather useless since you couldn't do any actual exchanges on Ropsten anyway, so I didn't implement it.

src/components/TextInput/TextInput.js
```diff
@@ -355,8 +355,19 @@ class TextInput extends React.Component<Props, State> {
       value: selectedValue,
     } = selector;
 
+    const shouldDisplaySpinner = this.getSelectorOptionsCount(selectorOptions) <= 1;
```
graszka22, line 358: why `<= 1`? sometimes we have just one option and it's fine, see aave withdraw or sablier withdraw

src/utils/exchange.js
```diff
@@ -19,29 +19,31 @@
...
+    const image = staticImages[providerIconName] || '';
+    return image;
   }
   return '';
 };
 
-export const getCryptoProviderName = (providersMeta: ProvidersMeta, provider: string) => {
-  const providerInfo = providersMeta.find(({ shim }) => shim === provider) || {};
+export const getCryptoProviderName = (provider: string) => {
+  const providerInfo = PROVIDERS_META.find(({ shim }) => shim === provider) || {};
```
poocart, line 46: This repeats a lot, I wonder if we can add this as an util where we can only provide "provider" to lookup?

PR #2148, opened 2020-08-10
PIL-751 Altalix Integration

wjkg:


src/actions/fiatToCryptoActions.js
```diff
@@ -0,0 +1,37 @@
...
+import { SET_ALTALIX_INFO } from 'constants/fiatToCryptoConstants';
+
+import type { Dispatch, GetState } from 'reducers/rootReducer';
+import type SDKWrapper from 'services/api';
+
+export const loadAltalixInfoAction = () => {
+  return async (dispatch: Dispatch, getState: GetState, api: SDKWrapper) => {
+    const { fiatToCrypto: { altalix }, user: { data: { walletId } } } = getState();
+    if (altalix === null) {
+      const isAvailable = await api.fetchAltalixAvailability(walletId);
```
IAmKio, line 30: Could we add a `.catch` here and report the error please?

src/screens/Services/Services.js
```diff
@@ -227,39 +230,69 @@ class ServicesScreen extends React.Component<Props, State> {
...
+    return getAccountAddress(activeAccount);
   }
 
+  tryOpenCryptoPurchaseUrl = (url: string | null) => {
+    if (url) {
+      openInAppBrowser(url);
+    } else {
+      Toast.show({
+        type: 'warning',
+        message: 'Unable to prepare transaction. Please try again later.',
```
IAmKio, line 291:
```suggestion
        message: 'Sorry, we had a problem trying to open this service. Please try again later.',
```

src/screens/Services/Services.js
```diff
@@ -227,39 +230,69 @@ class ServicesScreen extends React.Component<Props, State> {
...
 
-    const address: string = getAccountAddress(activeAccount);
-    const url = getUrlWithAddress(address);
-    openInAppBrowser(url);
+    return getAccountAddress(activeAccount);
   }
 
+  tryOpenCryptoPurchaseUrl = (url: string | null) => {
+    if (url) {
+      openInAppBrowser(url);
```
IAmKio, line 287: Does `openInAppBrowser` return an error or something that we can report when `openInAppBrowser` doesn't work?

src/services/api.js
```diff
@@ -703,6 +706,42 @@ class SDKWrapper {
...
+        method: 'POST',
+        httpsAgent: new https.Agent({ rejectUnathorized: false }),
+      },
+      data,
+    };
+
+    return this.pillarWalletSdk.configuration.executeRequest(requestOptions)
+      .then(response => response.data.url)
+      .catch(error => {
+        reportLog('generateAltalixTransactionUrl: SDK request error', error.response.data, Sentry.Severity.Error);
```
IAmKio, line 723: 👍

src/services/api.js
```diff
@@ -703,6 +706,42 @@ class SDKWrapper {
...
+        method: 'GET',
+        httpsAgent: new https.Agent({ rejectUnathorized: false }),
+      },
+      params: { walletId },
+    };
+
+    return this.pillarWalletSdk.configuration.executeRequest(requestOptions)
+      .then(response => ALTALIX_AVAILABLE_COUNTRIES.includes(response.data.country))
+      .catch(error => {
+        reportLog('fetchAltalixAvailability: SDK request error', error.response.data, Sentry.Severity.Error);
```
IAmKio, line 741: 👍

src/utils/fiatToCrypto.js
```diff
@@ -0,0 +1,53 @@
...
+  return `${SENDWYRE_WIDGET_URL}?${querystring.stringify({
+    accountId: SENDWYRE_ACCOUNT_ID,
+    dest: `ethereum:${address}`,
+    redirectUrl: SENDWYRE_RETURN_URL,
+  })}`;
+}
+
+export function altalixWidgetUrl(params: AltalixTrxParams, api: SDKWrapper) {
+  return api.generateAltalixTransactionUrl(params);
+}
```
IAmKio, line 53: Could this be better written as:

```suggestion
export function altalixWidgetUrl(params: AltalixTrxParams, api: SDKWrapper) => api.generateAltalixTransactionUrl(params);
```

src/actions/__tests__/fiatToCryptoActions.test.js
```diff
@@ -0,0 +1,53 @@
...
+import ReduxAsyncQueue from 'redux-async-queue';
+
+import { SET_ALTALIX_INFO } from 'constants/fiatToCryptoConstants';
+import { loadAltalixInfoAction } from 'actions/fiatToCryptoActions';
+
+function mockStore({ state, pillarSdk }) {
+  return configureMockStore([thunk.withExtraArgument(pillarSdk), ReduxAsyncQueue])(state);
+}
+
+describe('Fiat to crypto providers actions', () => {
```
graszka22, line 32: whoa tests! 👀   respect 👍

src/screens/Services/Services.js
wjkg: `openInAppBrowser` takes care of reporting to Sentry, but I imagine we'd like to show the same error message as with failed URL construction -- I'll make it return a rejected promise on top of that.

src/utils/fiatToCrypto.js
```diff
@@ -0,0 +1,53 @@
...
+export function wyreWidgetUrl(address: string) {
+  return `${SENDWYRE_WIDGET_URL}?${querystring.stringify({
+    accountId: SENDWYRE_ACCOUNT_ID,
+    dest: `ethereum:${address}`,
+    redirectUrl: SENDWYRE_RETURN_URL,
+  })}`;
+}
+
+export function altalixWidgetUrl(params: AltalixTrxParams, api: SDKWrapper) {
+  return api.generateAltalixTransactionUrl(params);
```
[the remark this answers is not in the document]
wjkg: It's not a state change, but an address to be used only once -- it would add artificial complexity to a simple 1. construct url 2. open browser; action. Unless that's a solved problem somewhere else in the app?

src/utils/fiatToCrypto.js
```diff
@@ -48,6 +48,4 @@ export function wyreWidgetUrl(address: string) {
   })}`;
 }
 
-export function altalixWidgetUrl(params: AltalixTrxParams, api: SDKWrapper) {
-  return api.generateAltalixTransactionUrl(params);
-}
+export function altalixWidgetUrl(params: AltalixTrxParams, api: SDKWrapper) => api.generateAltalixTransactionUrl(params);
```
graszka22, line 51: syntax error :D

src/utils/fiatToCrypto.js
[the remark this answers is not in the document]
graszka22: I haven't seen anything similar that's why it caught my eye.  The most similar case I can think of would be in lending actions, for example `calculateLendingDepositTransactionEstimateAction`. We use reducer for storing the fact that we're calculating transaction fee and the transaction fee even though we use it only once on one screen and it doesn't need to be in redux store really. And as far as I understand (it's a bit hackish) you're in fact creating an action that returns (?) the api and that's weird to me.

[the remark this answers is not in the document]
wjkg: I looked at that action -- the resulting reducer fragment is still passed to a component, whereas here it's given to a function (indirectly to `InAppBrowser.open`) when a button is pressed. You're correct, it's a hack, and I'd like to have a better solution, but it's needed because the api wrapper is accessible only from within actions while having a method that returns something that doesn't belong to the state. What I wanted was to avoid breaking up the code between a click handler and a `componentDidUpdate`. (By _only once_ I meant not that it occurs in a single screen/component, but that it's used immediately after construction and should be discarded after.)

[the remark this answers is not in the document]
graszka22: Ok, let's leave it 👍

PR #2149, opened 2020-08-10
Fix Max's 2nd device UDID

antonioned:
Remove Max Zaripov's 2nd device so the ios pipeline is not blocked


antonioned:
Fixed the UDID

PR #2153, opened 2020-08-12
reset existing key based balances on balance fetch

poocart:
Resets existing balances for Key based account.


src/actions/assetsActions.js
```diff
@@ -376,6 +377,24 @@ export const fetchAssetsBalancesAction = () => {
   };
 };
 
+export const resetAccountBalancesAction = (accountId: string) => {
+  return async (dispatch: Dispatch, getState: GetState) => {
```
IAmKio, line 381: Do we need the `async` here?

poocart: No actually, good catch, removed 👍

PR #2157, opened 2020-08-12
Release Candidate 2.16.0

IAmKio:
**Release Candidate 2.16.0**

- New! The Exchange is now entirely in-app
- New! Altalix is now available from the Services screen
- New! Sablier is now available from the Services screen
- New! Emoji in-app toasts 😄 
- New! Brand new splash screen
- Improved: Now easier to close toasts
- Fixed: Key wallet balances are now removed from the total balance on the Home screen


graszka22:
Isn't the new splash screen missing in the release notes? :p

IAmKio:
@graszka22 updated - sorry, i'm so sure i added that because i was looking at it in the milestone!

PR #2158, opened 2020-08-13
PIL-775: change the graph api implementations to call by subgraph names

poocart:
https://linear.app/pillarproject/issue/PIL-775/investigate-and-fix-aave-service-failed-errors

Includes a change to call The Graph API using subgraph names instead of IDs. Reason of that is because IDs tend to change once subgraph is updated and this causes data loss (old subgraph IDs no longer synced) and request drops (400 error).


src/constants/exchangeConstants.js
```diff
@@ -39,4 +39,5 @@ export const SLOW = 'min';
 export const NORMAL = 'avg';
 export const FAST = 'max';
 
-export const UNISWAP_GRAPH_ID = 'QmXKwSEMirgWVn41nRzkT3hpUBw29cp619Gx58XW6mPhZP';
+// mainnet only
+export const UNISWAP_SUBGRAPH_NAME = 'uniswap/uniswap-v2';
```
jdnmr, line 43: Why is it "mainnet only"? When using ID, it worked just fine on ropsten.

poocart: Per our discussion in DM – each subgraph represents single network, however, in this case you're fetching only symbols from this subgraph so it's ok as long as we're not fetching something more network related like contract addresses, etc.

PR #2162, opened 2020-08-13
PIL-767 Wyre wallet order reservations

wjkg:
Instead of opening the Wyre widget directly from the app, the user is shown a short form with currencies and amount, which are then sent to Wyre's WOR endpoint in order to generate the URL.


src/screens/SendwyreInput/SendwyreInput.js
```diff
@@ -0,0 +1,330 @@
...
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import * as React from 'react';
+import { ScrollView, Keyboard } from 'react-native';
+import type { NavigationScreenProp } from 'react-navigation';
+import styled, { withTheme } from 'styled-components/native';
+import { connect } from 'react-redux';
+import { SDK_PROVIDER } from 'react-native-dotenv';
+import t from 'tcomb-form-native';
```
jdnmr, line 27: `tcomb-form-native` hasn't been maintained for over 2 years now. While we use it here and there, I think we should avoid introducing it in new components. The way `formOptions` is structured is confusing, and I also don't like the idea of keeping all of the form data (including constant values) in `this.state`. Do you think you could re-write this to avoid using the lib? You could just import `TextInput` and `SelectorOptions` components directly, and keep just the several changing values in state instead of the whole `formOptions` thing.
But if you feel this is too much work, you can just leave it as it is.

src/reducers/fiatToCryptoReducer.js
```diff
@@ -18,34 +18,89 @@
...
-  isAvailable: boolean,
-};
+import type { SendwyreRates } from 'models/FiatToCryptoProviders';
 
-export type FiatToCryptoReducerState = {
-  altalix: null | AltalixInfo,
-};
+export type FiatToCryptoReducerState = {|
+  isAltalixAvailable: null | boolean,
+  sendwyreExchangeRates: null | SendwyreRates,
```
jdnmr, line 34:
```suggestion
  isAltalixAvailable: ?boolean,
  sendwyreExchangeRates: ?SendwyreRates,
```

wjkg: I'd prefer to have only one value that means "no data" to compare against - if I understand correctly, `?type` would also accept `undefined`.

jdnmr: Yes, `undefined` would also be accepted. In that case, leave it as it is

src/reducers/fiatToCryptoReducer.js
```diff
@@ -18,34 +18,89 @@
...
+  sendwyreCountrySupport: $Values<typeof SENDWYRE_SUPPORT>,
+|};
 
 type AltalixInfoSetAction = {
   type: typeof SET_ALTALIX_INFO,
-  payload: AltalixInfo,
+  payload: { isAvailable: boolean },
```
poocart, line 40: I wonder if we really need `isAvailable` or we can just leave `boolean` value for payload?

src/screens/SendwyreInput/SendwyreInput.js
```diff
@@ -0,0 +1,330 @@
...
+import { supportedAssetsSelector } from 'selectors/selectors';
+
+// models, types
+import type { Assets, Asset } from 'models/Asset';
+import type { Dispatch, RootReducerState } from 'reducers/rootReducer';
+import type { Theme } from 'models/Theme';
+import type { FormSelector } from 'models/TextInput';
+import type { Option } from 'models/Selector';
+import type { SendwyreTrxValues } from 'models/FiatToCryptoProviders';
+
```
poocart, line 59:
```suggestion


```

src/services/api.js
```diff
@@ -726,22 +741,72 @@ class SDKWrapper {
...
+    return this.makeDirectSdkRequest({
+      path: '/partners/wyre/generate-order-reservation',
+      method: 'POST',
+      data: { walletId },
+    })
+      .then(() => true)
+      .catch(error => {
+        const { response: { status, data } } = error;
+
+        if (status === 400) return true;
```
poocart, line 785: Bad request still means that country is supported? If it's an edge case – please describe in comment 🙏

wjkg: Yes, because the location check happens first, validation second. Do you think the above comment (https://github.com/pillarwallet/pillarwallet/pull/2162/files/0e0ee87036521162aec7fca06e50929cec56a233#diff-4c562ce127f5f3fc6eee5530c0c52d30R768-R774) is unclear or missing something? I'd welcome a suggestion on how to explain it better.

src/reducers/fiatToCryptoReducer.js
wjkg: It got left over from where I flattened state to avoid making the reducer more complex. It's fixed now.

src/screens/SendwyreInput/SendwyreInput.js
```diff
@@ -0,0 +1,288 @@
...
+      this.setState({ destCurrency: this.props.defaultDest });
+    }
+  }
+
+  getCurrencyPair = (state: State): [string, string] => {
+    return [
+      get(state, 'sourceCurrency.symbol'),
+      get(state, 'destCurrency.symbol'),
+    ];
+  }
```
jdnmr, line 155: 1) Why pass `state` as an argument if you can just access it inside the function? 
2) nit: instead of using `get`, you can access nested values via optional chaining, it's less prone to typos and it's fresh and fancy :) 
3) Both `get` and optional chaining might technically return `undefined`, which should be reflected in the return type.
How about something like:
```
getCurrencyPair = (): [?string, ?string] => {
  const { sourceCurrency, destCurrency } = this.state;
  return [ sourceCurrency?.symbol, destCurrency?.symbol ];
}
```

src/screens/SendwyreInput/SendwyreInput.js
```diff
@@ -0,0 +1,288 @@
...
+    this.setState({ isHandlingSubmit: false });
+  }
+
+  tryFillDefaultCurrencies = (prevState: State) => {
+    if (prevState.sourceCurrency === null && this.props.defaultSource !== null) {
+      this.setState({ sourceCurrency: this.props.defaultSource });
+    }
+
+    if (prevState.destCurrency === null && this.props.defaultDest !== null) {
+      this.setState({ destCurrency: this.props.defaultDest });
```
jdnmr, line 146: Can we destructure `defaultSource` and `defaultDest` to make this shorter and simpler?

PR #2163, opened 2020-08-13
Fix toasts shadow in dark mode

graszka22:
before:
image

after:
Screenshot 2020-08-13 at 16 36 34


src/components/ShadowedCard/ShadowedCard.js
```diff
@@ -145,27 +151,32 @@ class ShadowedCard extends React.Component<Props, State> {
...
-          useSVGShadow
-          wrapperStyle={{
-            position: 'absolute',
-            top: -(SHADOW_LENGTH / 2),
-            left: -(SHADOW_LENGTH / 2),
-            opacity: disabled ? 0.4 : 0.8,
-          }}
-          shadowRadius={(borderRadius || 6) - 2}
-        />}
+        {!noShadow && !!(cardHeight && cardWidth) && readyToRenderShadow && (forceShadow || !isDarkTheme) &&
```
poocart, line 162: Just thinking about difference of `forceShadow` and `noShadow` – both can "force" either use or not to use shadow 🤔

PR #2170, opened 2020-08-17
Release Candidate 2.16.2

IAmKio:
**Release Candidate 2.16.2**

- Optimised: token queries to limit to just what is required
- Fixed security issue with the app launcher
- Fixed Sablier issues that prevented the app from starting
- Fixed upcoming requirement with Wyre to continue operating without interruptions


IAmKio:
Just waiting for the iOS task to be fixed and we'll get this in.

PR #2175, opened 2020-08-18
fix/PIL-789-ramp-query-parameter-change

IAmKio:
- Modified fiatToCrypto.js utility to switch on PLR token purchases
- Added test around `fiatToCrypto.js` / `rampWidgetUrl` function


src/utils/fiatToCrypto.js
```diff
@@ -29,9 +29,10 @@ import type SDKWrapper from 'services/api';
 
 export function rampWidgetUrl(address: string, email?: string) {
   const params = {
+    swapAsset: 'PLR', // This turns on the ability to purchase PLR
     hostApiKey: RAMPNETWORK_API_KEY,
     userAddress: address,
-    ...(email ? {} : { userEmailAddress: email }),
+    ...(email ? { userEmailAddress: email } : {}),
```
poocart, line 35: 😱

IAmKio: I never would have noticed if i didn't write a test haha!

PR #2180, opened 2020-08-19
Translate/wallet connect

vyvil:
PR that applies translations to wallet connect screens.
And it also adds missing translation in Log out alert.


IAmKio (approved):
Thank you!

PR #2183, opened 2020-08-19
Upgrade/node version ios vm

antonioned:
Fix the node version issue we currently have with the new babel module dependencies.
This is a "hack" until we can upgrade to a newer xcode VM in circleci, which will be 11.5.0 which has node version 14.2.0


IAmKio (approved):
God bless `brew`!

PR #2187, opened 2020-08-20
updated modal hidden behaviour for ios

poocart:
Includes:
- Fix for as token send flow modals including updated contact details modal to show each if others are hidden as `react-native-modal` does not allow to have multiple modals open and it is needed to wait until visible is completely hidden before presenting another.
- Small fix for 🤷‍♀️  emoji as per `react-native-emoji` library it had different code and was failing to render as actual emoji.


IAmKio (approved):
Thanks for all your hard work on this, it's really appreciated!

PR #2189, opened 2020-08-21
Revert/some env changes

alazarevski:
removes environments/... and adds a single .js file for environment variables
removes react-native-dotenv as a dependency
Keeps .env with the only root level build needed variables (ios needs sentry and build type and number)


poocart (approved):
yolo

PR #2193, opened 2020-08-24
Fix/transaction fee estimation

jdnmr:
For failed tx fee estimations, default to 0 to prevent app crashes. This is done in the app in various places, but not all scenarios were covered. A particular usecase for this was trying to exchange a very small amount of a token (e.g. 0,0000000001 ETH).
IMO this should be treated as a temporary fix. "Confirmation" screens should handle error scenarios and display some kind of "could not calculate transaction fee" message instead, preventing sending the tx, but this would take some amount of work. Maybe we could line it up as part of unifying send flows @IAmKio


src/utils/assets.js
```diff
@@ -189,8 +189,12 @@ export const isEnoughBalanceForTransactionFee = (
 
   // subtract from balance if transaction asset matches fee asset, not suitable for collectibles
   if (transactionAmount && feeSymbol === transactionSymbol) {
-    const amountInWei = new BigNumber(utils.parseUnits(transactionAmount.toString(), transactionDecimals));
-    balanceInWei = balanceInWei.minus(amountInWei);
+    try {
+      const amountInWei = new BigNumber(utils.parseUnits(transactionAmount.toString(), transactionDecimals));
+      balanceInWei = balanceInWei.minus(amountInWei);
+    } catch (e) {
+      //
```
poocart, line 196: Wdyt if we add Sentry report here?

poocart (approved):

IAmKio (approved):
Thank you.

PR #2197, opened 2020-08-26
Translations/missing (components)

vyvil:
PR to apply translations to several components that were previously missed.


src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -592,23 +599,23 @@ export class ActivityFeedItem extends React.Component<Props> {
           data = {
             ...data,
             subtext: t('label.outgoingSablierStream'),
-            itemValue: `- ${formattedAmount} ${symbol}`,
-            fullItemValue: `- ${formattedAmount} ${symbol}`,
+            itemValue: getValueWithSymbol(formattedAmount, symbol, false, false),
```
graszka22, line 602: Just a thought: wouldn't it be nicer to introduce functions like `getPositiveValueWithSymbol(amount, symbol)`? It'd be more readable than `getValueWithSymbol(formattedAmount, symbol, false, false)`, this way I have to wonder "what does this false, false mean?".

src/components/Selector/Selector.js
```diff
@@ -181,7 +181,7 @@ const Selector = ({
   const hasValue = !isEmpty(selectedOption);
   const hasOptions = !!options?.length;
   const disabled = !hasOptions && !allowEnteringCustomAddress;
-  const placeholderText = !disabled ? `${placeholder}...` : 'no options to select';
+  const placeholderText = !disabled ? `${placeholder}...` : t('label.noOptionsToSelect');
```
graszka22, line 184: shouldn't the dots be somewhere in translation?

graszka22 (commented):
LGTM 👍  two minor comments

src/components/ActivityFeed/ActivityFeedItem.js
wjkg: I was going to mention the same thing, except with the alternative method of using an object to pass the options - so the parameter types would change from `(value: string, token: string, isPositive: boolean, doNotFormat: boolean)` to `(value: string, token: string, options: { isPositive: boolean, doNotFormat: boolean })`. This way one could still use conditions (like `formattedValue === '0'` in EventDetails).

vyvil: Yeah, good idea! Will update it shortly :) Thanks!

src/components/Selector/Selector.js
vyvil: Oh, yeah. I'll add them as suffix. Thanks! :)

src/components/ActivityFeed/ActivityFeedItem.js
vyvil: Renamed it too, to make it more readable :)

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -268,7 +268,10 @@ export class ActivityFeedItem extends React.Component<Props> {
     const { amount, symbol, decimals }: AaveExtra = event.extra;
     if (!amount || !symbol) return '';
     const value = formatUnits(amount, decimals);
-    return getValueWithSymbol(formatAmount(value, getDecimalPlaces(symbol)), symbol, isPositive, !value);
+    return getFormattedValue(formatAmount(value, getDecimalPlaces(symbol)), symbol, {
+      isPositive,
+      doNotFormat: !value,
```
wjkg, line 273:
```suggestion
      noSymbol: !value,
```

Also, is it possible to catch this with type checking?

src/utils/strings.js
```diff
@@ -20,14 +20,19 @@
...
 
-export const getValueWithSymbol =
-  (value: string, token: string, isPositive: boolean, doNotFormat: boolean) => {
-    if (doNotFormat) return t('tokenValue', { value, token });
-    if (isPositive) return t('positiveTokenValue', { value, token });
-    return t('negativeTokenValue', { value, token });
-  };
+export const getFormattedValue = (value: string, token: string, options: FormattedValueOptions = {}) => {
+  const { isPositive, noSymbol } = options;
+  if (noSymbol || isPositive === 'undefined') return t('tokenValue', { value, token });
```
wjkg, line 35: A missing `typeof`?

src/components/ActivityFeed/ActivityFeedItem.js
vyvil: Uff ;/
Yeah. I'll add exact object type for this in a minute :)

src/utils/strings.js
vyvil: Shoot, yup.
Fixing it and calling it a day :D
Thanks!

graszka22 (approved):

PR #2198, opened 2020-08-26
Fix Sablier ENS name resolving 

graszka22:
To reproduce:
- try to create a new stream
- enter ENS name instead of an address
- continue creating new stream
- crash after pressing "Next"

Also improved approving transactions:
- try to create a new stream
- an allowance modal should appear, allow that
- wait for it to approve
- after approving, app says you still need to approve


IAmKio (approved):
Thank you!

alazarevski (approved):

PR #2204, opened 2020-08-27
wabi token symbol fix in rates fetching

poocart:
Recently had an issue where we changed token from WABI to WaBi and this issue seems to have another impact on token price rates therefore returned result from crypto compare needs to be remapped to WaBi.

Added a general fix as this seems to happen with every other token.


IAmKio (approved):
Great work - thank you!

alazarevski (approved):

PR #2207, opened 2020-08-28
PIL-821: max send fixes

poocart:
https://linear.app/pillarproject/issue/PIL-821/max-button-calculation-failures

Includes:
1. Max send button was taking previous `txFeeInfo` which was always lower for max balance therefore wrong max amount was set and new estimate was bigger which led to not enough fees message.
2. Max send button was visible even if recipient was not set which is wrong because no right estimate can be made without recipient.
3. Max send button was clickable even if another estimate is in progress which caused trouble in calculations as well and we had it before value selector component was introduced so it's good to bring it back.
4. Fees calculation method needed to be wrapped into `useCallback` to avoid race condition calls on debounce.


IAmKio (approved):
LGTM! 🎉

alazarevski (approved):

graszka22 (commented):
something is wrong :/ why do I get not enough eth for tx fee?

graszka22 (approved):
can't reproduce anymore 🤷‍♀️

PR #2213, opened 2020-09-02
Translations/optimise

vyvil:
PR to prevent using non-translated strings across the app.
It also adds couple of missing translations.


src/actions/authActions.js
```diff
@@ -109,15 +109,15 @@ export const updateFcmTokenAction = (walletId: string) => {
   return async (dispatch: Dispatch, getState: GetState, api: SDKWrapper) => {
     const fcmToken = await firebaseMessaging.getToken().catch(e => {
       // We was unable to fetch the FCM token.
-      reportLog(`Unable to fetch Firebase FCM token: ${e.message}`, e);
+      reportLog(`Unable to fetch Firebase FCM token: ${e.message}`, e); // eslint-disable-line i18next/no-literal-string, max-len
```
wjkg, line 112: It seems the most recent version of eslint-plugin-i18next fixes the ignore options for template literals in `no-literal-string`. Is there a reason not to use it? Also, if it's the _eslint-disable_ comment that causes `max-len` to trigger, I'd try `eslint-disable-next-line`.

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferStatus.js
```diff
@@ -164,7 +164,7 @@ const KeyBasedAssetTransferStatus = ({
         <ScrollWrapper contentContainerStyle={{ paddingVertical: spacing.large }}>
           <Wrapper flex={1} center regularPadding>
             <Animation source={animationSuccess} />
-            <Title center>Key based assets migration complete</Title>
+            <Title center>{t('transactions.keyWalletAssetsMigrationDone')}</Title>
```
wjkg, line 167: It's
```suggestion
            <Title center>{t('transactions.title.keyWalletAssetsMigrationDone')}</Title>
```
in the json file.

.eslintrc
```diff
@@ -51,6 +52,243 @@
     "function-paren-newline": [0],
     "no-param-reassign": ["off"],
     "react/default-props-match-prop-types": ["off"],
-    "react/no-did-mount-set-state": ["off"]
-  }
+    "react/no-did-mount-set-state": ["off"],
+    "i18next/no-literal-string": ["error", {
+      "ignore": [
```
wjkg, line 57: Just an idea, but maybe we could use eslint's _extends_ to keep the ignore options in a separate file.

wjkg (commented):
If you run the linter with `--report-unused-disable-directives` there are some inline exceptions for `no-literal-string` that show up as unnecessary.

src/components/CircleButton/CircleButton.js
```diff
@@ -67,7 +67,7 @@ function getLabelColor(theme: Theme, isDisabled: boolean) {
   return colors.control;
 }
 
-const CircleButtonIconWrapperColors = ['#ffffff', '#f2f4f9'];
+const CircleButtonIconWrapperColors = ['#ffffff', '#f2f4f9']; // eslint-disable-line i18next/no-literal-string
```
wjkg, line 70: The plugin options accept regexes, so maybe all color codes could be ignored - I think the format is specific enough that it wouldn't match any UI text by accident.

src/services/api.js
```diff
@@ -781,14 +787,16 @@ class SDKWrapper {
         const { response: { status, data } } = error;
 
         if (status === 400) return true;
-        if (status === 403 && data.message === 'Location not supported') return false;
+        if (status === 403 && data.message === t('error.locationNotSupported')) return false;
```
wjkg, line 790: It's an error message returned from the backend, so it shouldn't be checked against a translated string.

wjkg (approved):

src/actions/authActions.js
vyvil: Updated :) Somehow I've missed this update.
It solves many literal string validation issues.
Yet it seems like it now ignores nested string values in objects that have properties named in all caps.
That's not that common in our project, but it's good to have this in mind. 
I'll put it into translations readme

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferStatus.js
vyvil: Fixed. Thanks for spotting it! :)

.eslintrc
vyvil: Correct me if I'm wrong, but only shared modules can be used to extend eslint config, right?
Since those options are quite specific not sure if it's that beneficial to export it as a separate shared module.
And it would be faster for us to extend those whenever project changes and requires new exceptions.

src/components/CircleButton/CircleButton.js
vyvil: Good call! Added. :)

src/services/api.js
vyvil: ! True, updated it. Thanks for spotting it!

.eslintrc
wjkg: I didn't test it, but this: https://eslint.org/docs/user-guide/configuring#using-a-configuration-file reads as if it could be any file. I may be wrong here.

vyvil: Neat! It works 👍 
Pushing an update :)

PR #2218, opened 2020-09-03
Fix sablier keyboard

graszka22:
The keyboard covers the next button and on iOS you can't hide it


IAmKio (approved):
The best kind of fix

PR #2224, opened 2020-09-08
Add Table component

graszka22:
An example of this component: https://zpl.io/anDZN3v

### DIFF
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -27,7 +27,7 @@ import { Spacing } from 'components/Layout';
 
 type TableAmountProps = {
   amount: string,
-  fiatAmount: string,
+  fiatAmount?: string,
 };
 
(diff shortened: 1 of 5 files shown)


src/components/Table/Table.js
```diff
@@ -0,0 +1,94 @@
...
+
+const AmountContainer = styled.View`
+  flex-direction: row;
+`;
+
+export const TableAmount = ({ amount, fiatAmount }: TableAmountProps) => {
+  return (
+    <AmountContainer>
+      <BaseText regular>{amount}</BaseText>
+      {fiatAmount && (
```
wjkg, line 69: It's clear the component is not supposed to be used with `fiatAmount` set to an empty string, but if for some reason that happens, this will cause an error.

wjkg (approved):

PR #2227, opened 2020-09-09
Improve/ramp asset type select

IAmKio:
An issue has appeared where by the Ramp widget is now locked to PLR. This is an issue on the Ramp side whereby the PLR token cannot be shown in the same list as ETH etc. It's either PLR or "other tokens".

This PR adds a simple Alert asking if you if you want to buy PLR or another token, until the Ramp widget can fix this issue on their side.


src/screens/Services/Services.js
```diff
@@ -213,7 +213,22 @@ class ServicesScreen extends React.Component<Props, State> {
...
           if (address === null) return;
-          this.tryOpenCryptoPurchaseUrl(rampWidgetUrl(address, email));
+
+          Alert.alert(
+            t('servicesContent.ramp.assetDecisionAlert.title'),
+            t('servicesContent.ramp.assetDecisionAlert.description'),
+            [
+              {
+                text: t('servicesContent.ramp.assetDecisionAlert.actionNonPlr'),
+                onPress: () => this.tryOpenCryptoPurchaseUrl(rampWidgetUrl(address, email, false)),
```
jdnmr, line 223: nit: you can just ommit the `false` because `plrMode` in `rampWidgetUrl ` is optional

src/utils/fiatToCrypto.js
```diff
@@ -24,14 +24,15 @@ import { getEnv } from 'configs/envConfig';
...
+export function rampWidgetUrl(address: string, email?: string, plrMode?: boolean) {
   const params = {
-    swapAsset: 'PLR', // This turns on the ability to purchase PLR
+    swapAsset: plrMode ? 'PLR' : null, // This turns on the ability to purchase PLR
     hostApiKey: getEnv().RAMPNETWORK_API_KEY,
     userAddress: address,
     ...(email ? { userEmailAddress: email } : {}),
   };
 
+  console.log('Launching:', `${getEnv().RAMPNETWORK_WIDGET_URL}?${querystring.stringify(params)}`);
```
jdnmr, line 35: Do we want this? If we do, we don't normally use `console.log` directly, we have a `printLog` function for this

jdnmr (approved):
LG!

src/screens/Services/Services.js
IAmKio: Great spot! This has been corrected.

src/utils/fiatToCrypto.js
IAmKio: Removed 👍

PR #2228, opened 2020-09-09
Fix sablier datemodal issues

graszka22:
Fixes:
- try to set the ending date, but press the "next" button in modal very quickly - the date is not selected
- if you don't set the date but tap +1 hour for example, it defaults to December 31


IAmKio (approved):
Thanks for the quick turnaround!

PR #2231, opened 2020-09-10
PIL-849 PIL-850 Uniswap data points

graszka22:


IAmKio (commented):
Could we get some tests around this please?

graszka22:
@IAmKio the tests are with some dirty mocks, I'm not sure if it's possible to test these functions without them 😅

src/constants/liquidityPoolsConstants.js
```diff
@@ -0,0 +1,24 @@
...
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+export const UPDATE_POOL_DATA = 'UPDATE_POOL_DATA';
+export const SET_FETCHING_POOL_DATA = 'SET_FETCHING_POOL_DATA';
+
+export const UNISWAP_PLR_ETH_POOL_ADDRESS = '0xae2d4004241254aed3f93873604d39883c8259f0';
```
poocart, line 24: Maybe we can deploy and add one for Kovan?

src/services/sablier.js
```diff
@@ -283,21 +223,8 @@ export const getCreateStreamFeeAndTransaction = async (
...
-      from: sender,
-      to: asset.address,
-      data: approveTransactionData,
-      amount: 0,
-      symbol: ETH,
-      sequentialSmartWalletTransactions: [sablierCreateStreamTransaction],
-    };
-  }
+  sablierCreateStreamTransaction = await addAllowanceTransaction(
+    sablierCreateStreamTransaction, sender, getEnv().SABLIER_CONTRACT_ADDRESS, asset, amount);
```
poocart, line 227: The transaction will fail if the required amount is already allowed.

src/services/uniswap.js
```diff
@@ -344,3 +346,96 @@ export const fetchUniswapSupportedTokens = async (supportedAssetCodes: string[])
...
+  let addLiquidityTransaction = {
+    from: sender,
+    to: ADDRESSES.router,
+    data: addLiquidityTransactionData,
+    amount: utils.formatUnits(ethAmount, 18),
+    symbol: ETH,
+  };
+
+  addLiquidityTransaction = await addAllowanceTransaction(
+    addLiquidityTransaction, sender, ADDRESSES.router, tokenAsset, tokenAmount);
```
poocart, line 406: The transaction will fail if amount is already allowed.

src/constants/liquidityPoolsConstants.js
graszka22: it's not used anywhere, the uniswap router will figure out the pool address himself on both kovan and mainnet, removing 👍

src/services/sablier.js
graszka22: The function `addAllowanceTransaction` checks for allowance and adds allowance transaction only when it's needed
And btw why would it fail? Afaik you can ask for allowance multiple times for whatever amount you want.

poocart: The estimation will fail and then transaction.

PR #2237, opened 2020-09-14
PIL-113 Add option to delete profile picture

graszka22:


src/services/api.js
```diff
@@ -379,6 +379,16 @@ class SDKWrapper {
...
   }
 
+  deleteUserAvatar(walletId: string): Promise<boolean> {
+    return Promise.resolve()
+      .then(() => this.pillarWalletSdk.user.deleteProfileImage({ walletId }))
+      .then(response => response.status === 204)
+      .catch((error) => {
+        reportLog('Failed to delete user avatar', { error }, Sentry.Severity.Error);
+        return false;
+      });
```
IAmKio, line 389: Just a comment, if the return type of this function is already a Promise, do we need return Promise.resolve()? For example, could the function be written like this:

```suggestion
  deleteUserAvatar(walletId: string): Promise<boolean> {
    await this.pillarWalletSdk.user.deleteProfileImage({ walletId }))
      .then(response => response.status === 204)
      .catch((error) => {
        reportLog('Failed to delete user avatar', { error }, Sentry.Severity.Error);
        return false;
      };
```

IAmKio (commented):
Just a small comment regarding a Promise'd function, other than that - looks good.

src/services/api.js
graszka22: It doesn't really matter but it's a convention in this file. Each api function starts with Promise.resolve(). Maybe code aesthetic?

IAmKio: No problem - not going to cry about this one 😂

IAmKio (approved):

PR #2242, opened 2020-09-17
do not estimate deployment fee if modal is invisible

poocart:
Per title.


IAmKio (approved):
https://linear.app/pillarproject/issue/PIL-855/ghost-call-to-estimate-deployment-cost

PR #2243, opened 2020-09-17
PIL-856: fix bignumber exponential values for gas token display values

poocart:
https://linear.app/pillarproject/issue/PIL-856/white-screend-when-paying-with-plr

Note: separated value parser code from `formatTransactionFee` as `getFormattedTransactionFeeValue` since it's hard to test values with translations added on top.

### DIFF
--- a/src/utils/__tests__/common.test.js
+++ b/src/utils/__tests__/common.test.js
@@ -272,7 +272,7 @@ describe('Common utils', () => {
 
   describe('getFormattedTransactionFeeValue', () => {
     it('should parse from BigNumber', () => {
-      const txWeeInWei = new BigNumber(1234500000000000000);
+      const txFeeInWei = new BigNumber(1234500000000000000);
       const formattedEth = getFormattedTransactionFeeValue(txWeeInWei);
       const formattedGasToken = getFormattedTransactionFeeValue(txWeeInWei, gasToken);
       expect(formattedEth).toBe('1.2345');
--- a/src/utils/__tests__/common.test.js
+++ b/src/utils/__tests__/common.test.js
@@ -33,7 +33,14 @@ import {
   formatFiat,
   extractJwtPayload,
   parseTokenAmount,
+  getFormattedTransactionFeeValue,
 } from '../common';
+
+const gasToken = {
+  address: '0x0',
+  decimals: 18,
+  symbol: 'PLR',
+};
 
 describe('Common utils', () => {
   describe('delay', () => {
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/common.test.js
```diff
@@ -262,4 +269,36 @@ describe('Common utils', () => {
       expect(parseTokenAmount('100', 0)).toBe(expectedValue);
     });
   });
+
+  describe('getFormattedTransactionFeeValue', () => {
+    it('should parse from BigNumber', () => {
```
IAmKio, line 274: What happens if `getFormattedTransactionFeeValue` encounters an invalid value altogether? How would it behave?

IAmKio (approved):
Thanks for sorting this out at such short notice! 🚀

src/utils/__tests__/common.test.js
```diff
@@ -262,4 +269,36 @@ describe('Common utils', () => {
...
+      const txWeeInWei = new BigNumber(1234500000000000000);
```
jsidorenko, line 275:
```suggestion
      const txFeeInWei = new BigNumber(1234500000000000000);
```

src/utils/common.js
```diff
@@ -591,18 +591,26 @@ export const getDeviceWidth = () => {
   return Dimensions.get('window').width;
 };
 
-export const formatTransactionFee = (feeInWei: string | number, gasToken: ?GasToken): string => {
-  if (!feeInWei) return '';
+export const getFormattedTransactionFeeValue = (feeInWei: string | number | BigNumber, gasToken: ?GasToken): string => {
+  // fixes exponential values with BigNumber.toPrecision()
+  const parsedFeeInWei = typeof feeInWei === 'object' // BigNumber
```
jsidorenko, line 596: I think there is `Bignumber.IsBignumber()` method available

src/utils/common.js
```diff
@@ -591,18 +591,26 @@ export const getDeviceWidth = () => {
...
+    ? feeInWei.toPrecision()
```
jsidorenko, line 597: How about putting a note (todo), that we should replace this to `feeInWei.toFixed()` after we upgrade to the newest Bignumber version? As this will be a breaking change

src/utils/common.js
```diff
@@ -591,18 +591,26 @@ export const getDeviceWidth = () => {
...
+    : feeInWei.toString();
 
   if (gasToken && !isEmpty(gasToken)) {
-    const { symbol, decimals } = gasToken;
-    return t('tokenValue', {
-      value: formatAmount(utils.formatUnits(feeInWei.toString(), decimals), 2),
-      token: symbol,
-    });
+    return formatAmount(utils.formatUnits(parsedFeeInWei, gasToken.decimals), 2);
```
jsidorenko, line 601: So here we're returning the rounded value, but later we use the real value for validation.. could that cause a problem when you have a bit less than a fee displayed?

src/utils/__tests__/common.test.js
poocart: It would throw unhandled exception.

src/utils/common.js
poocart: Not related to this PR, but we can definitely change the rounding to default, however, I'm not sure I remember why we adde rounding here in first place.

src/utils/common.js
poocart: 👍

src/utils/common.js
poocart: will check and amend if works, thanks 👍

src/utils/common.js
jsidorenko: Yep, that could be checked after that PR gets merged

poocart: 👍

jsidorenko (approved):

PR #2251, opened 2020-09-28
PIL-867: sync sw acc deployment state on acc connect

poocart:
https://linear.app/pillarproject/issue/PIL-867/account-deployment-state-issue

Smart Wallet account deployment state is synced on deployment transaction confirmed event and when Smart Wallet account is set as active (previously we had Key based as second account).

Currently account set as active action is no longer dispatched after account is set as active (as we only have one Smart Wallet account) therefore the only Smart Wallet account state sync that remains is on transaction event, which is not triggered (TODO: find out why) in scenario of interrupting app online state (close/open app, toggle network) during Smart Wallet deployment process.

Added Smart Wallet account deployment state sync to action when account is being connected which is triggered on each app successful auth.

### DIFF
--- a/src/actions/smartWalletActions.js
+++ b/src/actions/smartWalletActions.js
@@ -284,7 +284,7 @@ export const connectSmartWalletAccountAction = (
 ) => {
   return async (dispatch: Dispatch, getState: GetState) => {
     if (!smartWalletService || !smartWalletService.sdkInitialized) return;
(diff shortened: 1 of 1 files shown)


jsidorenko (approved):

jsidorenko:
👍

IAmKio (approved):

PR #2253, opened 2020-09-29
Add tooltip component

graszka22:
Yeah, I know we have already a Walkthrough component, but it's a bit different and harder to use. To use this one you just have to wrap your component with a Tooltip component and that's all - it will be automatically measured etc.


src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,200 @@
...
+  body: string,
+  positionOnBottom?: boolean,
+  isVisible: boolean,
+  wrapperStyle?: Object,
+  children?: React.Node,
+};
+
+const TooltipParagraph = styled(Paragraph)`
+  color: ${themedColors.control};
+  ${fontStyles.regular}px;
```
wjkg, line 54: Looks like those styles are from the walkthrough tooltip, but Isn't `fontStyles.regular` an object?
```suggestion
  ${fontStyles.regular};
```

wjkg (approved):
LGTM, but in case your intention was to include the tooltip itself in the snapshot tests, it doesn't show up.

PR #2254, opened 2020-09-29
Add new ValueInput component

graszka22:
This component is supposed to replace the Value Selector Card and inputs in the exchange flow with one, easy to use component with updated design (because it will be heavily used for example in liquidity pools). For now I've replaced the Exchange, Send flow, and send synthetics flow (btw is something broken with synthetics, because I can only select PLR?). The next PR will replace the rest of Value Selector Cards (they should be easier) and delete it completely.


src/components/TextInput/TextInput.js
```diff
@@ -575,26 +558,38 @@ class TextInput extends React.Component<Props, State> {
...
+                          this.multilineInputField = input;
+                        }}
+                        onChange={this.handleChange}
+                        onBlur={this.handleBlur}
+                        onEndEditing={this.handleBlur}
+                        onFocus={this.handleFocus}
+                        onSubmitEditing={this.handleSubmit}
+                        value={textInputValue}
+                        autoCorrect={autoCorrect}
+                        style={[defaultInputStyle, customStyle, additionalStyle, !editable && { color: colors.accent }]}
```
vyvil, line 582: Input is not taking the whole `ItemHolder` width, so even regular (not numeric) inputs have text aligned to the right ;/

src/components/TextInput/TextInput.js
```diff
@@ -575,26 +558,38 @@ class TextInput extends React.Component<Props, State> {
...
+                        onLayout={onLayout}
+                        placeholderTextColor={colors.accent}
```
vyvil, line 584: Based on designs it's `colors.text` :)

src/components/ValueInput/ValueInputHeader.js
```diff
@@ -0,0 +1,112 @@
...
+const Wrapper = styled.View`
+  width: 100%;
+  flex-direction: row;
+  align-items: center;
+  justify-content: space-between;
+  margin-bottom: 15px;
+  height: 24px;
+`;
+
+const SideWrapper = styled.TouchableOpacity`
```
vyvil, line 50: This could use some paddings on sides (based on position (right / left)).
As if max value is super long, it touches with asset value. Or wise versa.

Or maybe we need to wrap / shorten token name to fit everything into one line?

src/configs/buildConfig.js
```diff
@@ -28,7 +28,7 @@ export const buildEnvironment = {
   BUILD_NUMBER: '_build_number_',
   BUILD_TYPE: buildType,
   OPEN_SEA_API_KEY: '_open_sea_api_key_',
-  INFURA_PROJECT_ID: '_infura_project_id_',
+  INFURA_PROJECT_ID: '9c4d7497182349fc90ef646a3e0b8c05',
```
vyvil, line 31: Do you really want to commit this? :)

src/containers/SendContainer/SendContainer.js
```diff
@@ -101,12 +109,15 @@ const SendContainer = (props: Props) => {
         allowEnteringCustomAddress
         {...customSelectorProps}
       />
-      <ValueSelectorCard
-        selectorModalTitle={t('transactions.title.valueSelectorModal')}
-        maxLabel={t('button.sendMax')}
-        wrapperStyle={{ paddingTop: spacing.medium }}
-        {...customValueSelectorProps}
-      />
+
```
vyvil, line 112:
```suggestion
```

src/containers/SendContainer/SendContainer.js
```diff
@@ -101,12 +109,15 @@ const SendContainer = (props: Props) => {
...
+      <InputWrapper>
+        {!!isLoading && <Spinner />}
+        {!isLoading && (
+        <ValueInput
+          {...customValueSelectorProps}
+        />
+        )}
```
vyvil, line 119:
```suggestion
        {!!isLoading ? <Spinner /> : <ValueInput {...customValueSelectorProps} />}
```

src/screens/Exchange/Exchange.js
```diff
@@ -211,9 +193,9 @@ class ExchangeScreen extends React.Component<Props, State> {
       ((
         fromAsset !== prevFromAsset ||
         toAsset !== prevToAsset ||
-        fromAmount !== prevFromAmount ||
-        (prevErrorMessage && !errorMessage)) &&
-      validateInput(fromAmount, fromAsset, toAsset, errorMessage))
+        fromAmount !== prevFromAmount) &&
+      isFormValid &&
+      validateInput(fromAmount, fromAsset, toAsset))
```
vyvil, line 198:
```suggestion
        isFormValid &&
        validateInput(fromAmount, fromAsset, toAsset))
```

src/components/SendTokens/EthereumTokens.js
```diff
@@ -449,15 +456,15 @@ const SendEthereumTokens = ({
...
+        onValueChange: setAmount,
+        assetData,
+        onAssetDataChange: setAssetData,
+        showCollectibles: true,
         txFeeInfo,
-        preselectedAsset: token,
-        preselectedCollectible,
-        showAllAssetTypes: true,
-        gettingFee,
         hideMaxSend: gettingFee || !selectedContact, // we cannot calculate max if no receiver is set
```
vyvil, line 465: I guess, this (`// we cannot calculate max if no receiver is set`) is no longer valid, isn't it?

src/screens/Exchange/Exchange.js
```diff
@@ -242,106 +224,61 @@ class ExchangeScreen extends React.Component<Props, State> {
...
-    let toAmount;
-    let toAmountInFiat;
+    let toAmount = '0';
     if (offers?.length && fromAmount) {
-      toAmount = getBestAmountToBuy(offers, fromAmount);
-      if (toAmount) {
-        toAmountInFiat = formatAmount(getBalanceInFiat(baseFiatCurrency, toAmount, rates, toAsset.symbol || ''), 2);
-        value = displayFiatToAmount ? toAmountInFiat : formatAmount(toAmount || '', 6);
-      }
+      toAmount = getBestAmountToBuy(offers, fromAmount) || '0';
```
vyvil, line 259: We used to shorten this value. 
Now if value has multiple decimal points, it's hard to comprehend it. Especially when input is disabled and there's no way to "scroll" the value.

src/screens/Exchange/Exchange.js
```diff
@@ -242,106 +224,61 @@ class ExchangeScreen extends React.Component<Props, State> {
...
+        assetData={toAsset}
+        onAssetDataChange={(assetData) => this.setState({ toAsset: assetData })}
+        selectorOptionsTitle={t('label.buy')}
+        customAssets={toOptions}
+        horizontalOptions={horizontalOptions}
         leftSideSymbol="+"
-        onLeftSideTextPress={() => this.setState({ displayFiatToAmount: !displayFiatToAmount })}
-        rightPlaceholder={displayFiatToAmount ? fiatCurrency : toAsset.symbol || ''}
+        onBlur={this.blurFromInput}
+        hideMaxSend
```
vyvil, line 273: Based on designs, it shouldn't be hidden. 
Although, I'd argue that it isn't needed here, too.
At least not clickable :)

vyvil (commented):
Had some notes :)

Also: what do you think about renaming "SendTokens/EthereumTokens.js" to reflect the content of this file more? :)

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,200 @@
...
+const Tooltip = (props: Props) => {
+  const {
+    body,
+    children,
+    positionOnBottom = true,
+    isVisible,
+    wrapperStyle,
+  } = props;
+
+  const [wrapperLayout, setWrapperLayout] = React.useState<Layout>({
```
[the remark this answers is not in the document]
graszka22: https://flow.org/en/docs/react/children/ 
> Note: You need to use import * as React from 'react' here instead of import React from 'react' to get access to the React.Node type. We explain why that is in the React Type Reference.

src/components/ValueInput/ValueInputHeader.js
graszka22: I'm not sure I follow - where should I add paddings?

src/screens/Exchange/Exchange.js
graszka22: https://zpl.io/awKZWGd

src/components/Tooltip/Tooltip.js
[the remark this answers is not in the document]
vyvil: Aaaah, I see.
Didn't check that React.Node was needed :)

src/components/ValueInput/ValueInputHeader.js
vyvil: On the horizontal sides.
Try selecting asset with a long name and / or the one that has significant balance (that takes up lots of scree space)

src/utils/formHelpers.js
```diff
@@ -201,7 +200,7 @@ export function SelectorInputTemplate(locals: Object) {
   return (
     <TextInput
       style={{ width: '100%' }}
-      hasError={!!locals.error}
+      hasError={false}
```
vyvil, line 203: I guess it could just be removed?

src/components/ValueInput/ValueInputHeader.js
graszka22: Added some padding between the asset and amount. Also asset name is ellipsized if necessary.

Screenshot 2020-10-08 at 15 48 28

vyvil: Perfect :)

PR #2255, opened 2020-09-29
Theming/new colors non interchangeable

vyvil:
This PR is adding new theme colors and a new method to get color by active theme since in new design system not all colors are interchangeable.
(I'll remove previously used ones after the switch).

Also in Zeplin all color names have suffix "Light" or "Dark".
I'm not including it in color naming to allow interchange them where possible.

(note that it is being merged into `theme` branch.
It is so that we could release all color changes at once but changes's PR could be chunked (so, no 🐋)


src/utils/themes.js
```diff
@@ -284,3 +288,20 @@ export function getThemeType(currentTheme: Theme = defaultTheme) {
 export function getThemeName(currentTheme: Theme = defaultTheme) {
   return currentTheme.current.replace('Theme', '');
 }
+
+export const getColorByTheme = (props: ColorsByThemeProps) => {
```
graszka22, line 292: Just making sure I understand: this function will be used instead of `themedColors` in styled components like this?
```js
const Root = styled.View`
  background-color: ${getColorByTheme({ lightKey: 'basic070', darkKey: 'basic070' })}
`
```
looks like a lot of typing :(

vyvil: Yeah, but just on cases where colors are not interchangeable.

Usually, this will be enought:
```
const Root = styled.View`
  background-color: ${({ theme }) => theme.colors.basic070};
`
```

And as we might have more themes in the future, I've added props as an object.

graszka22: that's better :D

graszka22 (approved):

PR #2257, opened 2020-09-30
Theming/surface colors

vyvil:
Merge to **`theme`** branch:
Change surface color to basic070.


graszka22 (approved):
The dark theme is darker than before, but I guess it's the new theme, right? 
Screenshot 2020-09-30 at 15 51 07

vyvil:
@graszka22, yes that's correct :)

PR #2259, opened 2020-09-30
Update selector component

graszka22:


src/components/ArrowIcon/ArrowIcon.js
```diff
@@ -0,0 +1,44 @@
...
+import { withTheme } from 'styled-components/native';
+import { images } from 'utils/images';
+import type { Theme } from 'models/Theme';
+
+
+type Props = {
+  theme: Theme
+};
+
+const ArrowIcon = ({ theme }: Props) => {
```
vyvil, line 32: Hm, why not to use our Icon component for that?
(it would require to add this `directIcon` to be added to our icon font first)

src/components/Selector/Selector.js
```diff
@@ -146,43 +141,43 @@ const Selector = ({
...
-        itemImageSource={imageSource}
-        padding="0 14px"
-      />
+      <SelectedOption>
+        <ProfileImage
+          uri={updatedUserImageUrl}
+          userName={name}
+          diameter={16}
+          noShadow
+          borderWidth={0}
```
vyvil, line 168: I guess, since diameter is small, initials should be smaller.
`initialsSize={10}`
What do you think? :)

src/components/Selector/Selector.js
```diff
@@ -192,17 +187,10 @@ const Selector = ({
...
-          {!hasValue && (
-            <MediumText style={{ paddingHorizontal: spacing.layoutSides }} big>
-              {placeholderText}
-            </MediumText>
-          )}
-        </SelectedOption>
-      </Wrapper>
+      <SelectorPill onPress={() => setIsOptionsVisible(true)} disabled={disabled}>
+        {hasValue && renderOption(selectedOption)}
+        {!hasValue && <BaseText link medium>{placeholderText}</BaseText>}
```
vyvil, line 192:
```suggestion
        {hasValue ? renderOption(selectedOption) : <BaseText link medium>{placeholderText}</BaseText>}
```

src/containers/SendContainer/SendContainer.js
```diff
@@ -90,23 +96,25 @@ const SendContainer = (props: Props) => {
...
-      />
+      <Wrapper>
+        <ValueSelectorCard
+          selectorModalTitle={t('transactions.title.valueSelectorModal')}
+          maxLabel={t('button.sendMax')}
+          wrapperStyle={{ paddingTop: spacing.medium }}
+          {...customValueSelectorProps}
+        />
+        <ArrowIcon />
+        <Spacing h={16} />
```
vyvil, line 107: Since this tends to repeat, why not include it into `Selector` (based on prop? Say, topIconProps? And then if it's present, render an Icon with such props?)

Also, `h={20}` :)

vyvil (commented):
Some notes:

src/containers/SendContainer/SendContainer.js
graszka22: I don't think it's worth it. Sometimes the arrow is above the selector, sometimes it's under. And those are just two lines of code. I don't feel like the arrow is a part of the selector, but rather it's a part of the screen and I like to keep components as simple as possible.

vyvil: Alright 👍

vyvil (approved):
👍

PR #2265, opened 2020-10-06
Add Table component

graszka22:
https://zpl.io/2ZeJJAq


wjkg (approved):
Looks OK. One minor optional thing -- in most files, the licence at the top has line breaks between paragraphs, so if you want to keep them consistent, it could be replaced with the more common version.

PR #2269, opened 2020-10-07
Add ReviewSummary component

graszka22:
Component like this https://zpl.io/2ZeJJAq (token icon + some text + nicely formatted amount) repeats many times in the designs so I've made it as a separate component


src/components/ReviewSummary/ReviewSummary.js
```diff
@@ -0,0 +1,104 @@
...
+
+const StyledCollectibleImage = styled(CollectibleImage)`
+  width: 192px;
+  height: 192px;
+`;
+
+const ReviewSummary = ({
+  asset, collectible, text, amount, rates, baseFiatCurrency,
+}: Props) => {
+  if (asset && amount) {
```
wjkg, line 63: Wouldn't it be simpler to have two separate components? The only shared part is the container style, and the variant for collectibles doesn't even have to be wrapped with `connect`. Then all relevant props could be marked as non-optional.

Unless having it this way makes more sense with the intended use -- then I'd only fix this condition so `amount` being equal to 0 does not cause the summary to disappear, just in case.

wjkg (approved):

PR #2270, opened 2020-10-07
offline onboarding fixes

poocart:
Part of onboarding refactor, this includes:
- Extra fixes for connectivity check.
- Address error on offline retry.
- Multiple offline toast fix.


src/actions/onboardingActions.js
```diff
@@ -385,6 +384,7 @@ export const checkUsernameAvailabilityAction = (username: string) => {
 
     // if user is offline then proceed with local registration
     if (!getState()?.session?.data?.isOnline) {
+      Toast.closeAll(); // close current ones to not block with multiple toasts
```
graszka22, line 387: 🤔  I'm not getting multiple modals anymore 👍  but they're jumping every time I enter a letter and it's still covering input a bit. Can we show this toast just once? If that's hard, let's leave as it is.

Screenshot 2020-10-07 at 16 46 22

src/actions/historyActions.js
```diff
@@ -122,6 +123,8 @@ export const fetchSmartWalletTransactionsAction = () => {
       }
     }
 
+    if (!isOnline) return;
```
graszka22, line 126: Are those checks necessary? I've counted 4 of them in this PR :p Should we remember about offline check every time we make an api call?

graszka22 (commented):
I have two suggestions and two bugs:
1.
- create account while offline
- be home screen
- close the app
- turn on the internet
- start the app, enter pin, registration failed

2. Sign out while being offline. A crash with "Error: failed" message should appear

src/actions/historyActions.js
poocart: Per out last discussion this is needed, however, we can think how we can improve the check throughout the app, however, it's not matching this ticket scope.

src/actions/onboardingActions.js
poocart: Thanks, solved

IAmKio (approved):
Great work mate

graszka22 (approved):

PR #2274, opened 2020-10-09
fixed wallet backup flow

poocart:
Fixes 2 issues:
1. Minor –  we added new screen called Wallet Settings (ex Recovery Settings) and this screen had backup button which should not ask for PIN because Wallet Settings already asked for one, just minor UI issue that caused double PIN prompt.
2. Major – after onboarding refactor lots of things moved and Backup flow was missing refactored implementation.

### DIFF
--- a/src/actions/walletActions.js
+++ b/src/actions/walletActions.js
@@ -19,7 +19,6 @@
 */
 import { ethers } from 'ethers';
 import { NavigationActions } from 'react-navigation';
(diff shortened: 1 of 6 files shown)


IAmKio (approved):
Thanks for the fix!

vyvil (approved):
👍 

(also, we have backup toast, that if clicked takes user to settings.
Maybe we could direct user straight to Backup screen?)

PR #2276, opened 2020-10-09
Remove ValueSelectorCard

graszka22:


vyvil (approved):

src/screens/Lending/EnterDepositAmount.js
```diff
@@ -113,7 +114,7 @@ const EnterDepositAmount = ({
 }: Props) => {
   const preselectedAssetSymbol: string = navigation.getParam('symbol');
   const [selectedAssetSymbol, setSelectedAssetSymbol] = useState(preselectedAssetSymbol);
-  const [depositAmount, setDepositAmount] = useState(null);
+  const [depositAmount, setDepositAmount] = useState('');
```
poocart, line 117: Just wondering if we can keep it as `null`? Recently we're trying to get rid of positive fallback values (empty array, empty object, empty string).

src/screens/Sablier/Withdraw.js
```diff
@@ -78,31 +75,41 @@ const FooterWrapper = styled.View`
...
-    baseFiatCurrency,
-    rates,
     withdrawTransactionEstimate,
     isCalculatingWithdrawTransactionEstimate,
     useGasToken,
     balances,
   } = props;
 
-  const [withdrawAmountInWei, setWithdrawAmount] = useState(0);
+  const [withdrawAmount, setWithdrawAmount] = useState('');
```
poocart, line 94: Per earlier comment – can we keep it as `0`?

graszka22: Not really - if I add `'0'` here (string as it's a text field everywhere) then it would add leading zero. So if user inputs "1" he'll get "01". Not cool.
The initial state `''` means that user just hasn't typed anything yet = the text input is empty so it's a valid use case imho. And empty value will be detected in validation anyway. Setting it to null or 0 will add unnecessary complexity.

PR #2277, opened 2020-10-09
Add docker auth

antonioned:
Add docker auth


vnovacev (approved):

IAmKio (approved):
All good mate - as discussed i will merge this when we remove the hold label 👍

PR #2282, opened 2020-10-12
Update CollectiblesList component

graszka22:
This component is updated for this screen: https://zpl.io/b6LQPjg. And by the way this one https://zpl.io/VOePwQX (it's the same component).


src/components/CollectiblesList/CollectiblesList.js
```diff
@@ -0,0 +1,179 @@
...
+`;
+
+const Name = styled(BaseText)`
+  ${fontStyles.small};
+  letter-spacing: ${fontTrackings.small};
+  width: 100%; 
+  text-align: center;
+`;
+
+const Card = styled.TouchableOpacity`
```
vyvil, line 77: So no more shadows on cards?
(I remember there was an issue with Zeplin, and those were not included in designs, yet required. So just checking if this is not such case)

src/components/CollectiblesList/CollectiblesList.js
```diff
@@ -0,0 +1,179 @@
...
+const EmptyStateWrapper = styled.View`
+  align-items: center;
+  justify-content: center;
+  padding: 20px;
+  flex-grow: 1;
+`;
+
+const Name = styled(BaseText)`
+  ${fontStyles.small};
+  letter-spacing: ${fontTrackings.small};
```
vyvil, line 72: Guessing, this is a legacy code - in current designs there are no letter spacing :)

src/components/CollectiblesList/CollectiblesList.js
```diff
@@ -0,0 +1,179 @@
...
+
+const CollectiblesList = ({
+  collectibles, onCollectiblePress, isSearching, theme, activeAccountAddress,
+}: Props) => {
+  const [isReceiveModalVisible, setIsReceiveModalVisible] = useState<boolean>(false);
+
+  const renderCollectible = ({ item }: CollectibleItem) => {
+    const { name, image } = item;
+
+    const icon = image;
```
vyvil, line 103:
```suggestion
    const { name, image: icon } = item;
```

src/screens/Assets/WalletView.js
```diff
@@ -297,10 +302,10 @@ class WalletView extends React.Component<Props, State> {
           {activeTab === COLLECTIBLES && (
             <CollectiblesList
               collectibles={this.getFilteredCollectibles()}
-              searchQuery={query}
-              navigation={navigation}
-              activeAccountAddress={activeAccountAddress}
-            />)}
+              onCollectiblePress={this.handleCollectiblePress}
+              isSearching={!!query}
```
vyvil, line 306: Now it isn't changing much, but I guess, technically it should be:
```suggestion
              isSearching={query.length >= MIN_QUERY_LENGTH}
```
(as only then collectibles are being filtered)

vyvil (approved):
☀️
Just some notes:

src/components/CollectiblesList/CollectiblesList.js
graszka22: we can remove them https://zpl.io/b6LQPjg

PR #2290, opened 2020-10-14
PIL-871: added coingecko service

poocart:
https://linear.app/pillarproject/issue/PIL-871/change-price-oracle-to-coingecko

Includes change for price oracle to CoinGecko with fallback to CryptoCompare that can be controlled by feature flag (default is legacy off).

### DIFF
--- a/src/services/assets.js
+++ b/src/services/assets.js
@@ -19,7 +19,7 @@
 */
 import ethers, { Contract, utils, BigNumber as EthersBigNumber } from 'ethers';
 import cryptocompare from 'cryptocompare';
-import { getEnv } from 'configs/envConfig';
+import isEmpty from 'lodash.isempty';
 
 // constants
 import { ETH, HOT, HOLO, supportedFiatCurrencies } from 'constants/assetsConstants';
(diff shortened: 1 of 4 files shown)


src/services/assets.js
```diff
@@ -343,28 +355,59 @@ export function getExchangeRates(assets: string[]): Promise<?Object> {
...
+    ? await getLegacyExchangeRates(assetSymbols)
+    : await getCoinGeckoTokenPrices(assetSymbols);
+
+  // by any mean if CoinGecko failed let's try legacy way
+  if (!useLegacyCryptoCompare && isEmpty(rates)) {
+    rates = await getLegacyExchangeRates(assetSymbols);
+  }
+
+  if (!rates) {
+    reportErrorLog('getExchangeRates failed: no rates data', { rates });
```
vyvil, line 393: Maybe here it would be good to also have a method used to get the rates (legacy/non-legacy) as extra data?

vyvil (approved):

src/services/assets.js
poocart: 👍

graszka22 (approved):

PR #2292, opened 2020-10-15
Fix/fastlane register devices

antonioned:
This PR needs to be merged after fastlane guys merge their PR with the fix:

https://github.com/fastlane/fastlane/pull/17433


antonioned:
@poocart since the fastlane PR has been merged, this is also now safe to be merged as well.

poocart (approved):
🇲🇰

PR #2296, opened 2020-10-16
Fix/username error tooltip [PIL-886]

vyvil:
PR to fix error tooltip (of error on new username) being cut.

I've extended scrollview of ContainerWithHeader content height.
But it isn't enough as toast is absolute, hence it does not extend layout height and it isn't scrollable enough to allow user to see the toast when keyboard is visible.

For this I've added min-height for a wrapper.
It could also be solved in more complex way -by listening for toast layout changes and getting its height and then applying it as a bottom padding to the wrapper. If you'd prefer this approach - I'm happy to add it.

### DIFF
--- a/src/components/Layout/ContainerWithHeader.js
+++ b/src/components/Layout/ContainerWithHeader.js
@@ -141,7 +141,11 @@ class ContainerWithHeader extends React.Component<Props, State> {
       }
 
       return (
-        <ScrollView style={{ flex: 1 }} keyboardShouldPersistTaps={keyboardShouldPersistTaps}>
+        <ScrollView
+          style={{ flex: 1 }}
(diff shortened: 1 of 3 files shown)


graszka22 (approved):
I cannot reproduce the bug even on my smaller Android but it looks ok. Another idea is to try to use zIndex but for me it's fine as it is.

poocart (approved):

vyvil:
> I cannot reproduce the bug even on my smaller Android but it looks ok. Another idea is to try to use zIndex but for me it's fine as it is.

I though of it, but guess it wouldn't work - tooltip is Scrollview content. And what crops it when keyboard is open is ScrollView itself. So zIndex has no influence here.

PR #2297, opened 2020-10-16
PIL-812 Feature/firebase storage

jdnmr:
This aims to replace local storage with Firebase Realtime Database.

In the current shape, it replaces all data managed by direct calls to AsyncStorage (via `saveDbAction`), but does NOT yet replace `redux-persist`.

Tests will be added.


IAmKio:
### Closing PR - superseded by future plans

Unfortunately this PR is being closed due to the future plans to move the app to an API driven presentation layer. What would have been an optimisation for the app in the current state is now scheduled to be included at the API layer instead as a datastore. This makes implementing Cloud Firestore at this level redundant.

PR #2298, opened 2020-10-17
fix for coingecko endpoint prices

poocart:
Fixes CoinGecko token price parsing for multiple matching token symbols. Initial approach used fetch by symbols, this approach replaces prices endpoint to fetch by token contract addresses.

### DIFF
--- a/src/services/assets.js
+++ b/src/services/assets.js
@@ -44,7 +44,7 @@ import ERC721_CONTRACT_ABI_TRANSFER_FROM from 'abi/erc721_transferFrom.json';
 import BALANCE_CHECKER_CONTRACT_ABI from 'abi/balanceChecker.json';
 
 // services
-import { getCoinGeckoTokenPrices } from 'services/coinGecko';
+import { getCoinGeckoEtherPrice, getCoinGeckoTokenPrices } from 'services/coinGecko';
 import { firebaseRemoteConfig } from 'services/firebase';
 
 // types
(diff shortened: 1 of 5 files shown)


graszka22 (commented):
It seems to me it doesn't work for ETH. It shows just 0$ all the time even though it shows I have a small amount of eth. And as far as I know ETH doesn't have a real contract, right? So maybe that's it?

poocart:
Interesting, last night I checked it seemed to be good, but indeed it's not. After import was able to get 0 price as well. Updated this, thanks 👍

src/services/assets.js
```diff
@@ -382,7 +384,13 @@ export async function getExchangeRates(assetSymbols: string[]): Promise<?Object>
...
+    : await getCoinGeckoTokenPrices(assets);
+
+  // ether price fetched separately as it doesn't fit into CoinGecko token price endpoint
+  const etherPrice = await getCoinGeckoEtherPrice();
+  if (etherPrice) {
+    rates = { ...rates, [ETH]: etherPrice };
+  }
 
   // by any mean if CoinGecko failed let's try legacy way
   if (!useLegacyCryptoCompare && isEmpty(rates)) {
```
graszka22, line 396: if rates is empty before fetching eth, this check won't detect it

graszka22 (approved):
now it works for me 👍 one suggestion below

PR #2300, opened 2020-10-19
initiate SDK before pin success callback

poocart:
SDK must be initiated before onLoginSuccess in case OAuth tokens fail/expire since it's used to update front-end with new tokens.


IAmKio (approved):
Thank you 🙏

PR #2306, opened 2020-10-26
PIL-844: gas estimate maintenance

poocart:
https://linear.app/pillarproject/issue/PIL-844/maintenance-gas-price-estimation

This includes unified estimate with single action that does it all for
- Exchange.
- Collectibles.
- Tokens and Ether transactions.

Open PRs to this:
- Sablier, AAVE, Pool Together: https://github.com/pillarwallet/pillarwallet/pull/2309.
-  Wallet Connect: https://github.com/pillarwallet/pillarwallet/pull/2310.

WIP, will be added as separate PRs and still works without this root PR:
- PPN.


src/components/SendAsset/SendAsset.js
```diff
@@ -142,86 +155,37 @@ const SendAsset = ({
...
-      Toast.show({
-        message: t('toast.transactionFeeEstimationFailed'),
-        emoji: 'woman-shrugging',
-        supportLink: true,
-      });
-      setGettingFee(false);
-      setTxFeeInfo(null);
-      return null;
+    if ((!isCollectible && (!isValidAmount || value === 0)) || !assetData || !selectedContact) {
+      return;
```
IAmKio, line 164: Does this result in when the user taps on the related UI element - nothing happens? Is it worth just saying something to the user for this condition?

src/components/SendAsset/SendAsset.js
```diff
@@ -279,7 +243,7 @@ const SendAsset = ({
   };
 
   const handleFormSubmit = async () => {
-    if (submitPressed || !txFeeInfo || !selectedContact || !assetData) return;
+    if (submitPressed || !feeInfo || !selectedContact || !assetData) return;
```
IAmKio, line 246: I think this also results in the user tapping but no feedback / message in this condition?

src/locales/en/common.json
```diff
@@ -488,7 +488,8 @@
       "invalidSession": "Invalid Wallet Connect session",
       "alreadyPending": "A connection is already waiting"
     },
-    "reverted": "reverted"
+    "reverted": "reverted",
```
IAmKio, line 491: I know you didn't change this line but - just wondering how we might humanise this term "reverted"? I have horrible memories of this term when building out PPN!

src/locales/en/common.json
```diff
@@ -488,7 +488,8 @@
...
+    "unableToEstimateTransaction": "Unable to estimate transaction"
```
IAmKio, line 492:
```suggestion
    "unableToEstimateTransaction": "Unable to estimate transaction. Please try again."
```

src/screens/Exchange/ExchangeConfirm.js
```diff
@@ -134,146 +112,78 @@ const Row = styled.View`
...
-  getKeyWalletTxFee = (txSpeed?: string, gasLimit?: number): TransactionFeeInfo => {
-    const { gasInfo } = this.props;
-    txSpeed = txSpeed || SPEED_TYPES.NORMAL;
-    gasLimit = gasLimit || this.state.gasLimit || 0;
+  useEffect(() => { fetchTransactionEstimate(); }, []);
 
-    const gasPrice = gasInfo.gasPrice[txSpeed] || 0;
-    const gasPriceWei = utils.parseUnits(gasPrice.toString(), 'gwei');
+  const onConfirmTransactionPress = () => {
+    if (!feeInfo) return;
```
IAmKio, line 181: How does this appear to the user? Is it like a no-op?

src/screens/SendCollectible/SendCollectibleConfirm.js
```diff
@@ -97,285 +81,196 @@ const WarningMessage = styled(Paragraph)`
...
+   * we're fetching Rinkeby ETH if current network is Kovan because
+   * our used collectibles in testnets are sent only using Rinkeby
+   * so if we're not on Rinkeby itself we can only check Rinkeby balance
+   * using this additional call
+   */
+  const [rinkebyEth, setRinkebyEth] = useState(0);
+  const fetchRinkebyEth = () => fetchRinkebyETHBalance(keyBasedWalletAddress)
+    .then(setRinkebyEth)
+    .catch((error) => {
+      reportErrorLog('SendCollectibleConfirm screen fetchRinkebyEth failed', { error, keyBasedWalletAddress });
```
IAmKio, line 127: ❤️

src/services/smartWallet.js
```diff
@@ -524,7 +525,35 @@ class SmartWallet {
...
+           */
+          if (error.message.includes(errorReplaceString)) {
+            try {
+              const messageJsonPart1 = JSON.parse(error.message.replace(errorReplaceString, '').trim().slice(1, -1));
+              const messageJsonPart2 = JSON.parse(messageJsonPart1?.value?.body);
+              const estimateError = messageJsonPart2?.error;
+              if (estimateError?.message) errorMessage = estimateError.message;
+              if (estimateError?.data) errorMessage = `${estimateError.data}: ${errorMessage}`;
+            } catch {
+              // unable to decrypt json
```
IAmKio, line 548: Could we report this with the `e` from `catch` please? This will at least notify us if something changes on the Archanova side, or we receive an unexpected message from the service.

IAmKio (changes requested):
Fantastic optimisations and refactor! Just a few requested changes and everything else, i would be happy with an explanation or something! Thanks again!

src/locales/en/common.json
poocart: This translation here is hooked with Smart Wallet deployment errors, it's generic error that may be returned from Archanova, but I think we applied it as translatable and it's returned when Smart Wallet account deployment estimate method returns error. However, it is also used as string constant to check the error so it actually can either be split as translation and as a constant separately and translation changed to something more readable or we can leave as it is.

src/screens/Exchange/ExchangeConfirm.js
poocart: Actually I've covered this scenario in other screens as Toast message, will apply here as well, thanks!

src/components/SendAsset/SendAsset.js
poocart: Added a Toast and error log 👍

src/components/SendAsset/SendAsset.js
poocart: No, this is fine here, it just doesn't update the fee on input change if there are not enough data provided. Hope this makes sense.

IAmKio: Ahh got it - thanks!

src/locales/en/common.json
IAmKio: I personally would split it - the justification is that these translation files can, at any time, be changed remotely. A public translator may innocently come along and translate the term "reverted" - which would result the string constant check breaking.

IAmKio (approved):

PR #2309, opened 2020-10-29
PIL-844: services estimates refactor

poocart:
Part of https://linear.app/pillarproject/issue/PIL-844/maintenance-gas-price-estimation.

Includes estimates refactor for services:
- Sablier.
- AAVE.
- Pool Together.


src/locales/en/common.json
```diff
@@ -1353,7 +1353,12 @@
     "cannotFindTransactionHash": "Cannot find transaction hash. Please try again.",
     "cannotGetBlockchainExplorerLink": "Cannot get blockchain explorer link. Please try again.",
     "cannotEnableAsset": "Cannot enable asset. Please try again.",
-    "cannotSendCollectible": "Cannot send collectible. Please try again."
+    "cannotSendCollectible": "Cannot send collectible. Please try again.",
+    "cannotCreateStream": "Cannot create stream. Please try again.",
+    "cannotCancelStream": "Cannot cancel stream. Please try again.",
+    "cannotAllowAsset": "Cannot allow asset. Please try again.",
+    "cannotPurchaseTicket": "Cannot purchase ticket. Please try again.",
+    "cannotWithdraw": "Cannot withdraw. Please try again."
```
IAmKio, line 1361: Thanks for being polite ❤️

src/screens/Lending/DepositTransactionConfirm.js
```diff
@@ -80,7 +76,7 @@ const DepositTransactionConfirm = ({
       accountAddress,
       depositAmount,
       depositAsset,
-      txFeeInfo?.fee || new BigNumber(0),
+      feeInfo?.fee || new BigNumber(0),
```
IAmKio, line 79: Is this why we have the 0 gas estimation? Just wondering... is it worth displaying something other than 0? I'm just thinking that this is one of the flash points that are reported - 0 gas fee estimate - would a message be better here or is it possible that actually - something might cost 0 gas?

src/screens/Lending/WithdrawTransactionConfirm.js
```diff
@@ -80,10 +76,10 @@ const WithdrawTransactionConfirm = ({
       accountAddress,
       withdrawAmount,
       depositedAsset,
-      txFeeInfo?.fee || new BigNumber(0),
+      feeInfo?.fee || new BigNumber(0),
```
IAmKio, line 79: Same question here as before - as this might repeat, i won't point it out again.

src/screens/PoolTogether/PoolTogetherPurchase.js
```diff
@@ -151,77 +156,46 @@ class PoolTogetherPurchase extends React.Component<Props, State> {
...
-            txFeeInWei,
-            gasToken,
-          };
-          this.setState({ purchasePayload });
-        }
+      this.setState({ purchasePayload: null }, () => {
+        const purchasePayload = getPurchaseTicketTransaction(numberOfTickets, poolToken);
+        const { to, data, amount } = purchasePayload;
+        estimateTransaction(to, amount, data);
+        this.setState({ purchasePayload });
```
IAmKio, line 198: Is it worth moving:

```suggestion
        const purchasePayload = getPurchaseTicketTransaction(numberOfTickets, poolToken);
        const { to, data, amount } = purchasePayload;
        estimateTransaction(to, amount, data);
```
To a function, to avoid calling `this.setState` inside another `this.setState`?

src/screens/PoolTogether/PoolTogetherWithdraw.js
```diff
@@ -152,80 +156,46 @@ class PoolTogetherWithdraw extends React.Component<Props, State> {
...
-            txFeeInWei,
-            gasToken,
-          };
-          this.setState({ withdrawPayload });
-        }
+      this.setState({ withdrawPayload: null }, () => {
+        const withdrawPayload = getWithdrawTicketTransaction(ticketsCount, poolToken);
+        const { to, data, amount } = withdrawPayload;
+        estimateTransaction(to, amount, data);
+        this.setState({ withdrawPayload });
```
IAmKio, line 198: Just the same thing as above

src/screens/PoolTogether/PoolTogetherWithdraw.js
```diff
@@ -372,92 +359,88 @@ class PoolTogetherWithdraw extends React.Component<Props, State> {
...
+              </ContentRow>
+            )}
+            <ContentRow>
+              <Button
+                title={t('button.next')}
+                onPress={() => {
+                  if (submitDisabled) return null;
+
+                  if (!hasAllowance && !isApprovalExecuting) {
+                    this.setState({ isAllowModalVisible: true });
```
IAmKio, line 396: Just wondering if it's good practice to put a setState insider a render? I understand that we have two flag evaluations so it's really unlikely it will get stuck in a re-render loop, but it's happened to me before 😂

IAmKio (commented):
**Great work on the estimation refactor!**

Just a few small comments but please feel free to explain the decisions etc as there may be a reason which i haven't noticed. Thanks!

src/screens/PoolTogether/PoolTogetherWithdraw.js
wjkg: Just a comment, this state change looks like it's supposed to open a modal, so unless this branch will get merged into develop before the modals refactor (#2256, awaiting re-review atm), it'll have to look different anyway.

src/actions/sablierActions.js
```diff
@@ -76,14 +82,16 @@ export const calculateSablierWithdrawTransactionEstimateAction = (
...
-        return null;
-      });
-
-    dispatch({ type: SET_SABLIER_WITHDRAW_TRANSACTION_ESTIMATE, payload: estimate });
+    dispatch(estimateTransactionAction(recipient, value, data));
   };
 };
 
+export const calculateSablierCancelTransactionEstimateAction = (stream: Stream) => {
+  return (dispatch: Dispatch) => {
```
graszka22, line 90: can we make a function that given a transaction payload it will set loading and dispatch the estimate action? 
`dispatch({ type: SET_ESTIMATING_TRANSACTION, payload: true });`
and 
`dispatch(estimateTransactionAction(to, 0, data));`
and creating a new action every time we want an estimate would be avoided
instead we'd have just one function call

graszka22 (commented):
Hey, found a few bugs ^^ 

- try to deposit some eth via aave on kovan
- you'll get an error - that's ok, eth doesn't work with aave on kovan afaik
- go back, select another token 
- the error is still shown even though user didn't do anything
- the same applies for example to pool together 

Another one:
- try to deposit some token on aave
- enter more than you have
- the tooltip "Not enough token" will be shown
- but the tx estimation is still calculated without error  👀  you can go ahead and send it and it will fail

src/locales/en/common.json
poocart: Oh I miss Canada haha!

src/screens/Lending/DepositTransactionConfirm.js
poocart: Thanks, it was some fallback value leftover, completely removed 👍

src/screens/Lending/WithdrawTransactionConfirm.js
poocart: Removed per answer above 👍

src/screens/PoolTogether/PoolTogetherPurchase.js
poocart: Previously `this.setState` callback had a promise inside, but now it looks that we can get rid of reset (set as `null`) and just go ahead with `this.setState` after new value is calculated 👍

src/screens/PoolTogether/PoolTogetherWithdraw.js
poocart: 👍

src/actions/lendingActions.js
```diff
@@ -150,47 +147,42 @@ export const calculateLendingDepositTransactionEstimateAction = (
...
-      sequentialTransactions: estimateTransactions.slice(1), // exclude first, take rest if exist
-    };
-
-    const estimate = await smartWalletService
-      .estimateAccountTransaction(estimateTransaction)
-      .catch(() => null);
-
-    dispatch({ type: SET_LENDING_DEPOSIT_TRANSACTION_ESTIMATE, payload: estimate });
+    const sequentialTransactions = aaveDepositNeededTransactions
+      .slice(1) // exclude first, take rest if exist
```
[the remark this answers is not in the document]
poocart: I guess the only concern here can be is that if `aaveDepositNeededTransactions` can be undefined which will cause `.slice(1)` to fail, however, it's always an array so `.slice(1)` will always work. On single size array it will exclude first and become empty which means `sequentialTransactions` is empty, otherwise it will return anything that goes after 1 index. Does it makes sense?

src/screens/PoolTogether/PoolTogetherWithdraw.js
poocart: @IAmKio it's onPress so it's not on actual render, just like any other onPress method.

poocart: In this scenario this opens Asset allowance modal on next button press in case Asset requires allowance.

src/actions/sablierActions.js
poocart: So it's actually doing it inside `estimateTransactionAction`, however, different integrations have different estimate scenarios with different amount of promises or calculations to perform before hitting `estimateTransactionAction` therefore we want `dispatch({ type: SET_ESTIMATING_TRANSACTION, payload: true });` to be fired earlier.

What we can do is to change `dispatch({ type: SET_ESTIMATING_TRANSACTION, payload: true });` into `dispatch(setTransactionIsEstimatingAction());`, wdyt?

graszka22: > different integrations have different estimate scenarios

ok, wasn't aware of that

> What we can do is to change dispatch({ type: SET_ESTIMATING_TRANSACTION, payload: true }); into dispatch(setTransactionIsEstimatingAction());, wdyt?

I think it doesn't matter much, that's just a cosmetic change

src/actions/lendingActions.js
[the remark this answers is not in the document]
IAmKio: As long as we can be sure that it's always an array - we're all good!

src/screens/PoolTogether/PoolTogetherWithdraw.js
IAmKio: > @IAmKio it's `onPress` so it's not on actual render, just like any other `onPress` method.

Sorry i was referring to the actual call to `setState` which would facilitate a render call - but i must have not noticed the onPress - which renders my point invalid. I'm just going to leave it there :)

graszka22 (approved):

src/navigation/appNavigation.js
```diff
@@ -657,6 +661,8 @@ const poolTogetherFlow = createStackNavigator({
   [POOLTOGETHER_PURCHASE_CONFIRM]: PoolTogetherPurchaseConfirmScreen,
   [POOLTOGETHER_WITHDRAW]: PoolTogetherWithdrawScreen,
   [POOLTOGETHER_WITHDRAW_CONFIRM]: PoolTogetherWithdrawConfirmScreen,
+  [SEND_TOKEN_PIN_CONFIRM]: SendTokenPinConfirmScreen,
+  [SEND_TOKEN_TRANSACTION]: SendTokenTransactionScreen,
```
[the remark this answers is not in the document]
poocart: Thanks, pushed changes 👍

graszka22 (approved):

PR #2311, opened 2020-11-02
PIL-860 Synthetix integration

jdnmr:
- Integrate Synthetix exchanges;
- Remove obsolete Key Wallet code from ExchangeOffers;
- Refactor and unify allowance transactions logic. 

A Synthetix offer will appear on Exchange screen if the user asks for exchanging one Synth asset to another (e.g. sETH - sUSD). In such case, no other providers' offers will be displayed. Please note that SNX is not a Synth asset. 

Kovan note: you can obtain sUSD directly from Synthetix [like this](https://docs.synthetix.io/integrations/testnets/) (use `exchangeEtherForSynths` method). The only Synth assets currently supported on our side for Kovan are sUSD, sETH and sBTC.


src/screens/Exchange/ExchangeConfirm.js
```diff
@@ -177,21 +169,8 @@ class ExchangeConfirmScreen extends React.Component<Props, State> {
...
-    if (!isSmartAccount) {
-      gasLimit = await calculateGasEstimate({ ...this.transactionPayload, from: activeAccountAddress });
-      this.setState({ gasLimit });
-    }
-
-    const txFeeInfo = isSmartAccount
-      ? await this.getSmartWalletTxFee()
-      : this.getKeyWalletTxFee(txSpeed, gasLimit);
-
+    const txFeeInfo = await this.getSmartWalletTxFee();
```
IAmKio, line 173: Could we add a .catch clause here and report to Sentry please? We tend to have issues around this functionality (not necessarily the app's fault but it's just for visibility)

src/services/1inch.js
```diff
@@ -133,40 +131,11 @@ export const create1inchOrder = async (
...
-    nonce: txCount.toString(),
-    to: fromAssetAddress,
-    gasLimit: '0',
-    gasPrice: '0',
-    chainId: '1',
-    value: '0',
-    data: encodedContractFunction,
+export const create1inchAllowanceTx =
+  async (fromAssetAddress: string, clientAddress: string): Promise<AllowanceTransaction | null> => {
+    const allowanceTx = await createAllowanceTx(fromAssetAddress, clientAddress, EXCHANGE_ADDRESS);
```
IAmKio, line 136: Could this error? Just thinking if we can catch the error - even if we return the error, at least we can report then return the error.

src/services/synthetix.js
```diff
@@ -0,0 +1,124 @@
...
+import ERC20_CONTRACT_ABI from 'abi/erc20.json';
+import RATES_ABI from 'abi/synthetixRates.json';
+import EXCHANGE_ABI from 'abi/synthetixExchange.json';
+
+const ethProvider = getEthereumProvider(getEnv().NETWORK_PROVIDER);
+
+const exchangeAddress = '0xC011a73ee8576Fb46F5E1c5751cA3B9Fe0af2a6F';
+const ratesAddress = isProdEnv
+  ? '0xda80E6024bC82C9fe9e4e6760a9769CF0D231E80'
+  : '0xA36f5A656c48EB0b43b63293E690DA746162d40B';
```
IAmKio, line 52: Just wondering if these are values that we might need to store in Remote Config as opposed to hardcoding them?

src/services/synthetix.js
```diff
@@ -0,0 +1,124 @@
...
+      EXCHANGE_ABI,
+      'exchange',
+      [
+        utils.formatBytes32String(fromAsset.symbol),
+        parseTokenBigNumberAmount(amount, fromAsset.decimals),
+        utils.formatBytes32String(toAsset.symbol),
+      ],
+    );
+
+    if (!data) throw new Error();
```
IAmKio, line 68: Could we an an error message within the `new Error` please? I appreciate that we could trace the call stack but for tools like Sentry, it's more beneficial to have a message here.

src/services/synthetix.js
```diff
@@ -0,0 +1,124 @@
...
+
+    return {
+      orderId: '-',
+      sendToAddress: exchangeAddress,
+      transactionObj: txObject,
+    };
+  } catch (e) {
+    reportOrWarn(`Synthetix exchange failed for pair ${fromAsset.symbol}-${toAsset.symbol}`, e, 'warning');
+    return null;
+  }
```
IAmKio, line 86: Just curious to know why we're using a try / catch here? I'm suspecting that `encodeContractMethod` doesn't play nice and just throws raw `Error`s?

src/services/synthetix.js
```diff
@@ -0,0 +1,124 @@
...
+};
+
+const getSynthetixAllowance = async (clientAddress: string, fromTokenAddress: string): Promise<boolean> => {
+  const assetContract = new Contract(fromTokenAddress, ERC20_CONTRACT_ABI, ethProvider);
+  const allowance = await assetContract.allowance(clientAddress, exchangeAddress);
```
IAmKio, line 91: Could this fail? What happens if it does?

src/services/synthetix.js
```diff
@@ -0,0 +1,124 @@
...
+      utils.formatBytes32String(toAsset.symbol),
+    );
+    const toAmount = utils.formatUnits(toValue.toString(), toAsset.decimals);
+    const allowanceSet = await getSynthetixAllowance(clientAddress, fromAsset.address);
+    const amountBN = new BigNumber(amount.toString());
+    const toAmountBN = new BigNumber(toAmount.toString());
+    const askRate = toAmountBN.dividedBy(amountBN).toNumber().toFixed(5);
+
+    return parseOffer(fromAsset, toAsset, allowanceSet, askRate, PROVIDER_SYNTHETIX);
+  } catch (e) {
```
IAmKio, line 113: Same question as above re: try/catch

src/utils/exchange.js
```diff
@@ -95,3 +105,46 @@ export const parseOffer = (
...
+
+const ethProvider = () => getEthereumProvider(getEnv().NETWORK_PROVIDER);
+
+export const createAllowanceTx = async (
+  fromAssetAddress: string,
+  clientAddress: string,
+  contractAddress: string,
+): Promise<AllowanceTransaction | null> => {
+  if (!clientAddress) {
+    reportOrWarn('Unable to set allowance', null, 'error');
```
IAmKio, line 128: Could we change this to something like `Unable to set allowance: the clientAddress was falsey`?

src/utils/exchange.js
```diff
@@ -95,3 +105,46 @@ export const parseOffer = (
...
+
+    const txCount = await ethProvider().getTransactionCount(clientAddress);
+
+    return {
+      nonce: txCount.toString(),
+      to: fromAssetAddress,
+      chainId: '1',
+      data: encodedContractFunction,
+    };
+  } catch (e) {
```
IAmKio, line 146: I'm suspecting that `encodeContractMethod` is the reason for the `try / catch`es here

src/utils/__tests__/assets.test.js
```diff
@@ -164,5 +166,27 @@ describe('Assets utils', () => {
         });
       });
     });
+
+    describe('recognize Synthetix transactions', () => {
```
IAmKio, line 170: Can we add at least 1 failure scenario here please?

IAmKio (changes requested):
**Fantastic work!** Just a few small requests around error reporting and some questions - but other than that - can't wait to get this out! Thanks for your efforts.

src/screens/Exchange/ExchangeConfirm.js
poocart: @jdnmr this should be replaces with new estimate flow.

src/services/synthetix.js
jdnmr: Moved them to envConfig, as that's what we do with other services' addresses

src/services/1inch.js
jdnmr: Potential errors are caught and reported inside `createAllowanceTx `

src/services/synthetix.js
jdnmr: We're explicitly throwing an error in case `!data`, also we're calling a function on ethProvider which might error

src/services/synthetix.js
jdnmr: Good catch. I've never seen it fail (and we do it the same way for other services too), but I'm adding a try-catch to log the potential error and return false.

src/utils/exchange.js
jdnmr: And `ethProvider().getTransactionCount()`

src/services/synthetix.js
jdnmr: There's plenty going on here, calling contract methods and working on BNs, so there are multiple error possibilities

IAmKio (approved):
Looks good now and thank you for your explanations - all good there. Cheers!

PR #2312, opened 2020-11-04
Translations/fetching set up [PIL-839]

vyvil:
PR to hook up Transifex provided translations, by using firebase remote config's translations' base url and version strings.

It also provides test coverage for translations feature.


src/actions/cacheActions.js
```diff
@@ -25,11 +25,11 @@ import type { CachedUrls } from 'reducers/cacheReducer';
 import { CACHE_STATUS, REMOVE_URL_CACHE } from 'constants/cacheConstants';
 import { saveDbAction } from './dbActions';
 
-// TODO: should prevent status === CACHE_STATUS.DONE ?
 const canStartCaching = (urlAsKey: string, cachedUrls: CachedUrls) => {
   const { status } = cachedUrls[urlAsKey] || {};
   return !status // haven't yet cached it
     || status !== CACHE_STATUS.PENDING // not being cached
+    || status !== CACHE_STATUS.DONE // not already cached
     || status === CACHE_STATUS.FAILED; // allow to retry if failed
```
poocart, line 33: If there are 4 types of `CACHE_STATUS` so it means it can be cached either on `status === CACHE_STATUS.FAILED` or `status === CACHE_STATUS.CACHE_REQUESTED`, wouldn't that be easier to read, wdyt?

poocart (approved):
Code wise LGTM 👍

src/actions/localisationActions.js
```diff
@@ -60,65 +62,82 @@ type SetLngAndBundle = {
...
   } = props;
   let resources;
-  // TODO: pass in versioning;
   let version = '';
+  let baseUrl = '';
+
+  try {
+    version = firebaseRemoteConfig.getString(FEATURE_FLAGS.APP_LOCALES_LATEST_TIMESTAMP);
+    baseUrl = firebaseRemoteConfig.getString(FEATURE_FLAGS.APP_LOCALES_URL);
+  } catch (e) {
```
IAmKio, line 103: Just wondering if the try / catch is necessary here? I'm not sure that `getString` will throw an Error if the value isn't found...

App.js
```diff
@@ -160,32 +159,8 @@ class App extends React.Component<Props, *> {
...
-     *
-     * @url https://rnfirebase.io/reference/remote-config#activate
-     */
-
-    remoteConfig()
-      .activate()
-      .then((r) => log.info('Firebase Config: Activation result was:', r))
-      .catch(e => log.error('Firebase Config: An error occured loading defaults:', e));
+
+    await initFirebase();
```
IAmKio, line 163: Did we mean to `await` initFirebase here? Just some context: when we added feature we intended it to be completely non-blocking especially in the `App.js` file. I also checked `initFirebase` method in `src/services/firebase.js` and it appear that whilst it is an `async` function - we don't return a value or `await` anything in the method...

src/services/firebase.js
```diff
@@ -22,11 +22,48 @@ import '@react-native-firebase/crashlytics';
...
 
 
 export const firebaseAnalytics = firebase.analytics();
 export const firebaseIid = firebase.iid();
 export const firebaseCrashlytics = firebase.crashlytics();
 export const firebaseMessaging = firebase.messaging();
 export const firebaseRemoteConfig = firebase.remoteConfig();
+
+
+export const initFirebase = async () => {
```
IAmKio, line 38: Just continuing from the comment in `App.js` - We don't `await` or `return` a value here, is `async` needed?

src/services/firebase.js
```diff
@@ -22,11 +22,48 @@ import '@react-native-firebase/crashlytics';
...
+   * @url https://rnfirebase.io/reference/remote-config#activate
+   */
+
+  try {
+    const r = remoteConfig().activate();
+    log.info('Firebase Config: Activation result was:', r);
+  } catch (e) {
+    log.error('Firebase Config: An error occurred while activating:', e);
+  }
+};
```
IAmKio, line 69: Are the try / catches necessary here? Just wanted to explain why i keep pointing this out... from what i recall, .activate is a promise (just checked - it is: https://rnfirebase.io/reference/remote-config#activate) so we should be able to .catch this to handle our errors, instead of try / catch.

To me personally... try / catch is the "nuclear" option - like the last resort (and also the best solution) that we absolutely know that we cannot handle the error using built-in methods. JSON.parse is a good example when it errors, you need to use try / catch. The above isn't "wrong" per-se ... just feels a little bit overkill when we _could_ use `.promise`...

IAmKio (changes requested):
🌐  🎉  **aš susijaudinęs!!!** 🎉 🌐 

Honestly this will be a really defining feature of the app that will allow Pillar to traverse the globe, communities and cultures that were not as accessible before. It's a really exciting feature coming and makes the end of a long journey for both of us (and you especially!).

Just a few small comments, of all of them i'm only truly bothered by the `await` in `App.js`. Thanks for all your hard work! 🏆

src/actions/localisationActions.js
vyvil: It prevented crashing if firebaseRemoteConfig wasn't yet initiated. 
I'm waiting for it to have default values set in order to execute any initial app actions. So this check is no longer needed.

App.js
vyvil: I've reverted the change and I'm waiting of only default values to be set now. 
this means that if there are no baseUrl (did not fetch it in time, no cached value), local translations are used.
Later, when remote config values are fetched, translations are updated if needed.

src/services/firebase.js
vyvil: Reverted back - so no more waiting on both actions. Just on the first one to make sure remote config is initialised.

src/services/firebase.js
vyvil: Same here - reverted it back :)

IAmKio (approved):

PR #2316, opened 2020-11-05
feature/i18n-am-locale

IAmKio:
- Added AM (Amharic) locale.

### DIFF
--- a/src/configs/localeConfig.js
+++ b/src/configs/localeConfig.js
@@ -43,7 +43,7 @@ export default {
   // pairs of language code and language name in native language
   supportedLanguages: {
     [DEFAULT_LANGUAGE_CODE]: DEFAULT_LANGUAGE,
-    am: 'Amharic',
+    am: 'አማርኛ',
   },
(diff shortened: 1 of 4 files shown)


src/configs/localeConfig.js
```diff
@@ -38,6 +43,7 @@ export default {
   // pairs of language code and language name in native language
   supportedLanguages: {
     [DEFAULT_LANGUAGE_CODE]: DEFAULT_LANGUAGE,
+    am: 'Amharic',
```
vyvil, line 46: Value (language name) should be in native language (for better UX) :)

vyvil (approved):
Well done! 🚀 
Just one note :)

src/configs/localeConfig.js
IAmKio: Thanks for pointing that out - I'll get this changed 👍🏾

IAmKio: All done ✅

PR #2324, opened 2020-11-06
Fix eth estimations

graszka22:
Fixes estimation when sending ETH via exchange / aave / etc
Omg I was debugging that when working on Rari for hours, I thought that there is something wrong with my code  🤦‍♀️  And the recent estimations fix just showed this error.

### DIFF
--- a/src/services/smartWallet.js
+++ b/src/services/smartWallet.js
@@ -488,7 +488,7 @@ class SmartWallet {
     } = transaction;
     let { data, recipient } = transaction;
     const decimals = get(assetData, 'decimals');
-    const assetSymbol = get(assetData, 'token');
+    const assetSymbol = get(assetData, 'token', ETH);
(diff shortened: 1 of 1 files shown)


IAmKio (approved):

poocart (approved):
Looks good! 👍 My little concern would be why latest changes caused this piece to fail while it wasn't faiing before estimates refactor and remained untouched after it was merged. So just thinking maybe the root of the problem lays somewhere else 🤔

graszka22:
@poocart ok, it isn't because the previous estimations fix, I've reproduced this bug also before the sequential estimations fix. Sorry for the confusion, it was because I've noticed this while working on rari and before seq tx fix it was not estimating eth tx at all.

poocart:
Sorry for my confusion as well, I meant the estimates refactor PR in general and that this piece we made change was untouched and I'm not sure why it started failing at this point 🤔

PR #2329, opened 2020-11-09
PIL-861: The Graph retry message

wjkg:
In case the user is on a screen which depends heavily on data from The Graph and that query failed, displays a message prompting the user to try again.

Designs linked here: https://linear.app/pillarproject/issue/DES-323

- in general, the changes make it so handling errors that occur during The Graph API calls is lifted in the call stack from the service to the actions, which allows to mark errors in the Redux state;
- extra action types are only for documentation purposes, except for the Sablier ones -- those were necessary for reducer tests to pass type checking;
- there is one extra commit, excluded from this PR, on branch [`test/retry-the-graph-query`](https://github.com/pillarwallet/pillarwallet/tree/test/retry-the-graph-query) for testing, with a toggle that forces `callSubgraph` to always throw an error (note that the Pool Together action has a timeout after a successful query, though);
- @vyvil related to theming: this adds a new button with the old props. I believe you could make flow catch it by setting `<Button>` props to be exact type, so there's a warning after merging.


graszka22 (commented):
LGTM, two questions:

1.
- change the graph url in `callSubgraph` function so that every call to graph throws an error (that's the most reliable method I know 😄 )
- go to sablier, pool together, whatever
- you'll get an error box, that's ok, press cancel
- do something on the screen, refresh, the call will throw an error again but no error box anymore :(
Is that an intended behaviour?

2. The graph is called heavily on the home screen. I suspect it may be the cause of most of the graph errors (so many calls that they timeout or something? I often see such errors on app startup). There are no error boxes on home screen, but are they necessary here? The user would see just outdated data here. What do you think?

wjkg:
@graszka22 thank you for your feedback 😃 

1. Good point, it wasn't specified, but refreshing should reset cancellation.

2. The box is meant to show up when the user takes an action that doesn't succeed due to a failed API call, but not for background updates. Given that they mostly fill in a part of some screen, a particular call may not fall clearly into either category. So while home screen updates are not supposed to trigger this prompt, I'm open to suggestions if you think there's some other screen I should add -- for now, the only errors not reflected in the UI should be calls that add extra info to transactions in the transaction history.

src/services/poolTogether.js
```diff
@@ -365,7 +365,8 @@ export async function getPoolTogetherTransactions(symbol: string, address: strin
   const { unitType, poolContractAddress: contractAddress } = getPoolTogetherTokenContract(symbol);
   let deposits = [];
   let withdrawals = [];
-  const rawHistory = await fetchPoolTogetherHistory(contractAddress, address);
+  const rawHistory = await fetchPoolTogetherHistory(contractAddress, address)
+    .catch(() => null);
```
IAmKio, line 369: Could we report this at least and then return `null`?

src/services/theGraph.js
```diff
@@ -20,6 +20,20 @@
...
 import { reportLog } from 'utils/common';
 
+export class GraphQueryError extends Error {
+  subgraphName: string;
+  query: string;
+  responseError: Error;
+
+  constructor(subgraphName: string, query: string, responseError: Error) {
+    // eslint-disable-next-line i18next/no-literal-string
+    super(`The Graph API call to subgraph "${subgraphName}" failed`);
```
IAmKio, line 30: Just wondering if this would prevent `i18next/no-literal-string`?

```suggestion
    super(`The Graph API call to subgraph \"${subgraphName}\" failed`);
```

Maybe? Not sure if that works...

src/services/theGraph.js
```diff
@@ -29,6 +43,6 @@ export const callSubgraph = (subgraphName: string, query: string) => {
     .then(({ data: response }) => response.data)
     .catch((error) => {
       reportLog(`The Graph subgraph "${subgraphName}" API call failed`, { error, query });
```
IAmKio, line 45: Okay i can see what has happened here (and i did suspect that you may do this). I'll resolve the comments above asking to report to Sentry.

IAmKio (changes requested):
Fantastic work! Just a few small points but otherwise - well done! This will greatly help our user experience from within the app.

src/actions/poolTogetherActions.js
```diff
@@ -59,8 +62,20 @@ export const fetchPoolPrizeInfo = (symbol: string, sequence?: boolean) => { // s
       type: SET_POOL_TOGETHER_FETCHING_STATS,
       payload: true,
     });
-    if (Date.now() - 15000 > lastSynced[symbol]) {
-      const newPoolStats = await getPoolTogetherInfo(symbol, activeAccountAddress);
+
+    if (poolStatsGraphQueryFailed || Date.now() - 15000 > lastSynced[symbol]) {
+      const newPoolStats = await getPoolTogetherInfo(symbol, activeAccountAddress)
+        .catch(error => {
+          if (error instanceof GraphQueryError) {
```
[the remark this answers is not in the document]
IAmKio: Resolving - saw what happened later with `GraphQueryError`.

src/actions/sablierActions.js
```diff
@@ -58,7 +59,14 @@ export const fetchUserStreamsAction = () => {
     if (!smartWalletAccount) return;
 
     dispatch({ type: SET_FETCHING_STREAMS });
-    const streams = await fetchUserStreams(getAccountAddress(smartWalletAccount));
+    const streams = await fetchUserStreams(getAccountAddress(smartWalletAccount))
+      .catch(error => {
+        if (error instanceof GraphQueryError) {
```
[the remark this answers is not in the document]
IAmKio: Resolving - saw what happened later with `GraphQueryError`.

src/actions/exchangeActions.js
```diff
@@ -338,8 +341,30 @@ export const getExchangeSupportedAssetsAction = (callback?: () => void) => {
...
+    const uniswapAssetsSymbols = fetchUniswapSupportedTokens(supportedAssets.map(({ symbol }) => symbol))
+      .then(result => {
+        dispatch({
+          type: SET_UNISWAP_TOKENS_QUERY_STATUS,
+          payload: { status: UNISWAP_TOKENS_QUERY_STATUS.SUCCESS },
+        });
+        return result;
+      })
+      .catch(error => {
+        if (error instanceof GraphQueryError) {
```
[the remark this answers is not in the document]
IAmKio: Resolving - saw what happened later with `GraphQueryError`.

graszka22 (approved):

src/actions/exchangeActions.js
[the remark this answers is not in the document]
wjkg: Every error of this type is already reported, inside the function that makes direct requests to The Graph API.

src/services/poolTogether.js
wjkg: It's already reported in `callSubgraph`. This is one of the places where we don't want any additional handling -- that is, no changes to the UI, so it follows the previous behaviour of `callSubgraph` (returning `null` on error), only a bit higher in the call stack.

src/services/theGraph.js
wjkg: I checked to be sure, but no -- it's the template literal that causes the warning, and it's correct.

IAmKio: Super - thanks for clearing that up!

IAmKio (approved):

PR #2331, opened 2020-11-09
Add Rari deposit flow

graszka22:
There are a couple of cases when depositing assets to rari:
- **user wants to deposit an accepted currency** (currently only USDC, but it's fetched from the contract) -> call directly `deposit` method

in all other cases we'd need to exchange the token somehow to accepted token

- **user wants to deposit a stablecoin (DAI, USDT etc)** -> use mStable contract to get data necessary for the swap and call `exchangeAndDeposit` method
beware that exchanging via mStable takes huuuge transaction fee!
and mStable swaps on 1:(1-fee) basis, that is for example for 1 DAI you'll get exactly 1 USDC minus mStable fee (even though stablecoins aren't worth exactly 1$)
- **user wants to deposit mUSD token** -> use mStable, but use different method for getting swap data - I haven't implemented it because Pillar doesn't support mUSD yet
- **user wants to deposit other ERC-20 token** -> use 0x service to obtain orders, process them (the processing part is taken from Rari app - I don't know the details of the calculations, they're undocumented and scary 😨 ) and call `exchangeAndDeposit`
exchange fee for 0x is sent in ETH as transaction value
- **user wants to deposit ETH** -> use 0x as above, just in addition to exchange fee send also deposit in ETH

[more detailed docs](https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/master/USAGE.md)


IAmKio (changes requested):
**GREAT WORK!**

Another feature that our community will be really happy about! Just a few comments and suggestions but other than that - code-wise this good. Also great job on that complex order code! 🚀

IAmKio (approved):
Rocking and rolling 🎸

src/actions/transactionEstimateActions.js
```diff
@@ -59,6 +59,12 @@ export const resetEstimateTransactionAction = () => {
   };
 };
 
+export const setEstimatingTransactionAction = (isEstimating: boolean) => {
+  return (dispatch: Dispatch) => {
+    dispatch({ type: SET_ESTIMATING_TRANSACTION, payload: isEstimating });
+  };
+};
```
poocart, line 66:
```suggestion
export const setEstimatingTransactionAction = (isEstimating: boolean) => ({
    type: SET_ESTIMATING_TRANSACTION,
    payload: isEstimating,
});
```

src/actions/rariActions.js
```diff
@@ -80,3 +82,31 @@ export const fetchRariUserDataAction = () => {
...
+
+export const calculateRariDepositTransactionEstimateAction = (
+  rariDepositNeededTransactions: Object[],
+) => {
+  return (dispatch: Dispatch, getState: GetState) => {
+    const { accounts: { data: accounts } } = getState();
+    const smartWalletAccount = findFirstSmartAccount(accounts);
+    if (!smartWalletAccount) return;
+
+    dispatch({ type: SET_ESTIMATING_TRANSACTION, payload: true });
```
poocart, line 94: Can be replaced with yours added `setEstimatingTransactionAction`.

src/services/0x.js
```diff
@@ -0,0 +1,195 @@
...
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import axios from 'axios';
+import { BigNumber as EthersBigNumber } from 'ethers';
+import { reportErrorLog } from 'utils/common';
+
+// based on https://github.com/Rari-Capital/rari-dApp/blob/6d07a47206f409ee6c52690f43c172c25a26071b/src/rari-sdk/0x.js
```
poocart, line 25: 👍

src/services/0x.js
```diff
@@ -0,0 +1,195 @@
...
+  outputTokenAddress: string,
+  maxInputAmountBN: EthersBigNumber,
+  maxMakerAssetFillAmountBN?: EthersBigNumber,
+) => {
+  const { data: decoded } = await axios.get(
+    `https://api.0x.org/swap/v0/quote?sellToken=${
+      inputTokenAddress
+    }&buyToken=${
+      outputTokenAddress
+    }${maxMakerAssetFillAmountBN !== undefined && maxMakerAssetFillAmountBN !== null
```
poocart, line 37: Maybe `BigNumber.isBigNumber` would help with this check?

src/services/0x.js
```diff
@@ -0,0 +1,195 @@
...
+
+// based on https://github.com/Rari-Capital/rari-dApp/blob/6d07a47206f409ee6c52690f43c172c25a26071b/src/rari-sdk/0x.js
+export const get0xSwapOrders = async (
+  inputTokenAddress: string,
+  outputTokenAddress: string,
+  maxInputAmountBN: EthersBigNumber,
+  maxMakerAssetFillAmountBN?: EthersBigNumber,
+) => {
+  const { data: decoded } = await axios.get(
+    `https://api.0x.org/swap/v0/quote?sellToken=${
```
poocart, line 33: Wdyt if making this little more readable? Like contructing the url before the call 👀

src/services/0x.js
```diff
@@ -0,0 +1,195 @@
...
+      ? 1
+      : -1,
+  );
+
+  const orders = [];
+  let inputFilledAmountBN = EthersBigNumber.from(0);
+  let takerAssetFilledAmountBN = EthersBigNumber.from(0);
+  let makerAssetFilledAmountBN = EthersBigNumber.from(0);
+
+  for (let i = 0; i < decoded.orders.length; i++) {
```
poocart, line 66: Any way to simplify this cycle, putting some operators as methods, lots of DRY here, wdyt? 👀

src/utils/rari.js
```diff
@@ -0,0 +1,237 @@
...
+import { DAI, USDC, USDT, TUSD, mUSD, ETH, WETH, USD } from 'constants/assetsConstants';
+import { reportErrorLog, parseTokenBigNumberAmount } from 'utils/common';
+import RARI_FUND_MANAGER_CONTRACT_ABI from 'abi/rariFundManager.json';
+import RARI_FUND_PROXY_CONTRACT_ABI from 'abi/rariFundProxy.json';
+import ERC20_CONTRACT_ABI from 'abi/erc20.json';
+import MSTABLE_CONTRACT_ABI from 'abi/mAsset.json';
+import type { Asset, Rates } from 'models/Asset';
+
+
+const getRariAcceptedCurrencies = async () => {
```
poocart, line 33:
```suggestion
const getRariAcceptedCurrencies = () => {
```

src/utils/rari.js
```diff
@@ -0,0 +1,237 @@
...
+              acceptedCurrency,
+            ]),
+          txValue,
+          exchangeFeeBN: utils.parseEther((mStableFeeUSD / ethRates[USD]).toFixed(18)),
+          slippage: 0, // mStable has no slippage
+          rariContractAddress: getEnv().RARI_FUND_PROXY_CONTRACT_ADDRESS,
+        };
+      }
+    }
+  } else if (token.symbol === mUSD) {
```
poocart, line 134: What's up here?

src/utils/rari.js
```diff
@@ -0,0 +1,237 @@
...
+  const slippage = (Math.abs(price - guaranteedPrice) / price) * 100;
+
+  return {
+    depositTransactionData: encodeContractMethod(
+      RARI_FUND_PROXY_CONTRACT_ABI,
+      'exchangeAndDeposit(address,uint256,string,(address,address,address,address,uint256,uint256,' +
+        'uint256,uint256,uint256,uint256,bytes,bytes,bytes,bytes)[],bytes[],uint256)',
+      [
+        token.symbol === ETH
+          ? '0x0000000000000000000000000000000000000000'
```
poocart, line 182: `ethers.constants.AddressZero`

src/screens/Rari/RariAddDeposit.js
```diff
@@ -0,0 +1,247 @@
...
+          ..._transactionPayload,
+          sequentialSmartWalletTransactions: depositTransactions.slice(1), // take the rest except first,
+        };
+      }
+      setTransactionPayload(_transactionPayload);
+      setExchangeFee(_exchangeFeeBN);
+      setSlippage(_slippage);
+      calculateRariDepositTransactionEstimate(depositTransactions);
+    }).catch((error) => {
+      reportErrorLog('Rari service failed: Error creating transaction payload', { error });
```
poocart, line 146: Wdyt of showing `toast.rariServiceFailed'` toast?

poocart (approved):
Left some comments, besides that LGTML 👍

src/services/0x.js
graszka22: I've tried to simplify it a little, but it's not so simple. Flow will complain if I refactor `!== null`, `!== undefined` checks to `isBigNumber`. Extracting code from this loop to some functions would require functions to have several arguments and return values and I can see now literally only 3 lines that are repeating.

src/utils/rari.js
graszka22: The comment should explain everything... The flow is a little bit different for mUSD, but since mUSD is not supported by Pillar wallet, there is no way for the user to select it. So for now it makes no sense to implement it, but I added a comment just in case it will be added in the future.

poocart: So I think we can add this at the top of logic as "TODO" and remove empty "if" statement as it looks that something is missing even though comment is inside, wdyt?

PR #2333, opened 2020-11-12
Rename "feature flags" to "remote config"

wjkg:
With time, more remote config values were added, and it no longer makes sense to have them all under the name of "feature flags". 

The issue came up after the refactor of modals, and this is a separate PR containing only the renaming of files and variables.


vyvil (approved):
🎉

PR #2339, opened 2020-11-16
Add v2 rari pools

graszka22:
Adds rari Yield and ETH pool, according to the new design.


src/services/rariPoolsAPY.js
```diff
@@ -68,21 +69,27 @@ export const getAaveApyBNs = async () => {
...
     }
   }`;
   /* eslint-enable i18next/no-literal-string */
   const data = await callSubgraph(getEnv().AAVE_SUBGRAPH_NAME, query);
   if (!data) return null;
-  const result = data.reserves.reduce((APY, reserve) => {
+  const result = data.reserves
+    .filter(reserve => {
+      return reserve.symbol !== ETH ||
+        reserve.id === '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee0x24a42fd28c976a61df5d00d0599c34c4f90748c8';
```
wjkg, line 85: Maybe this could be extracted to a named constant? I'm suggesting mainly because it's not quite clear why this particular id is special, so a comment would be fine, too.

wjkg (commented):
The inline comments are merely stylistic suggestions, but the main issue I've noticed is that I get multiple errors -- they seem to be mostly "Can't get Rari fund balance" and "Can't get user account deposit in RSPT" with "Error: failed to meet quorum" nested inside. The loading indicator on the main screen doesn't disappear after dismissing the errors, too.

Other than that, there's only a minor difference with the design for the info screen: the spacing between list cards and headers.

wjkg (commented):
Looks good, though I see one thing that needs to be fixed: the way the message about a failed graph query works is, in case of an error, `callSubgraph` with throw an exception (of type `GraphQueryError` exported from `services/theGraph`) that needs to be caught at the action level. This is for easy identification of this particular error, so that the message is not shown to the user in other circumstances (because then it would be misleading -- sending the request again wouldn't help) without adding a special return value to every function it goes through. If there's a call to `callSubgraph`, this exception has to be caught somewhere above, whether the error is going to be marked in the redux state or not.

I believe Sablier has the most straightforward example, here: https://github.com/pillarwallet/pillarwallet/blob/develop/src/actions/sablierActions.js#L53

This is the a toggle I used for debugging: 
```diff
diff --git a/App.js b/App.js
index 974813dd9..173f7eb8c 100644
--- a/App.js
+++ b/App.js
@@ -65,6 +65,8 @@ import Button from 'components/Button';
 import PercentsInputAccessoryHolder from 'components/PercentsInputAccessory/PercentsInputAccessoryHolder';
 import Modal from 'components/Modal';
 
+import { GraphFailureToggle } from 'services/theGraph';
+
 // utils
 import { getThemeByType, defaultTheme } from 'utils/themes';
 import { log } from 'utils/logger';
@@ -303,6 +305,7 @@ class App extends React.Component<Props, *> {
                 language={i18next.language}
                 onNavigationStateChange={this.closeModalsOnScreenChange}
               />
+              <GraphFailureToggle />
               {!!getEnv().SHOW_THEME_TOGGLE &&
               <Button
                 title={`THEME: ${current}`} // eslint-disable-line i18next/no-literal-string
diff --git a/src/services/theGraph.js b/src/services/theGraph.js
index 5de3dd3e2..d919aa8c5 100644
--- a/src/services/theGraph.js
+++ b/src/services/theGraph.js
@@ -17,9 +17,29 @@
     with this program; if not, write to the Free Software Foundation, Inc.,
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
+import React, { useState } from 'react';
+
+import Button from 'components/Button';
 import axios from 'axios';
 import { reportLog } from 'utils/common';
 
+class RetryDebug {
+  static failRequest = false;
+}
+
+export const GraphFailureToggle = () => {
+  const [fail, setFail] = useState(true);
+  RetryDebug.failRequest = fail;
+
+  return (
+    <Button
+      // eslint-disable-next-line i18next/no-literal-string
+      title={`The Graph query error: ${fail ? 'on' : 'off'}`}
+      onPress={() => setFail(!fail)}
+    />
+  );
+};
+
 export class GraphQueryError extends Error {
   subgraphName: string;
   query: string;
@@ -36,6 +56,11 @@ export class GraphQueryError extends Error {
 }
 
 export const callSubgraph = (subgraphName: string, query: string) => {
+  if (RetryDebug.failRequest) {
+    console.log('\x1b[7;91m debug query error \x1b[m', subgraphName);
+    return Promise.reject(new GraphQueryError(subgraphName, query, new Error()));
+  }
+
   // eslint-disable-next-line i18next/no-literal-string
   const url = `https://api.thegraph.com/subgraphs/name/${subgraphName}`;
   return axios
```

(I tested, and it should apply cleanly on this branch)

src/actions/rariActions.js
```diff
@@ -62,23 +54,28 @@ export const fetchRariUserDataAction = () => {
     if (!smartWalletAccount) return;
     const smartWalletAddress = getAccountAddress(smartWalletAccount);
 
-    dispatch({ type: SET_FETCHING_RARI_USER_DATA });
+    dispatch({ type: SET_FETCHING_RARI_DATA });
 
-    const [userDepositInUSD, userDepositInRSPT, userInterests] = await Promise.all([
+    const [userDepositInUSD, userInterests, rariAPY] = await Promise.all([
```
poocart, line 59: Wdyt if we validate the promise output first and then destruct? Since if any of these calls fail then output is first rejection which may not be array and therefore it would crash here.

poocart (approved):

PR #2346, opened 2020-11-23
PIL-914: 1inch API update, other exchange fixes

poocart:
This includes multiple fixes improvements for exchange:
1. 1inch new API version updates (ref – https://linear.app/pillarproject/issue/PIL-914/update-1inch-api-implementation).
2. Reduced offer object as some params such as `nonce`, `gasPrice`, etc. are not needed and retrieving `nonce` even makes request longer.
3. Fixed issue where exchange transaction `data` is not propagated into `estimateTransactionAction`.
4. Fixed issue where exchange rate is displayed as exponential on `ExchangeConfirm` screen.
5. Fixed issue where `fromAsset` and `toAsset` objects were mutating between different provider requests.


src/services/1inch.js
```diff
@@ -111,23 +116,21 @@ export const create1inchOrder = async (
...
-  const txObject = {
-    data: response.data,
-    nonce: txCount.toString(),
-    to: response.to,
-    gasLimit: response.gas || '0',
-    gasPrice: response.gasPrice || '0',
-    chainId: '1',
-    value: response.value,
-  };
+  if (!response || response?.tx) {
```
jdnmr, line 119: Are you sure we want to log error and return null if `response?.tx` is truthy? Didn't you mean `!response?.tx` ?

src/utils/1inch.js
```diff
@@ -80,8 +80,7 @@ export const getResponseData = async (url: string, errorMessage: string, toastMe
   return response.data;
 };
 
-export const parseAssets = (assets: Asset[]) => {
-  assets.forEach(asset => {
-    asset.code = asset.symbol;
-  });
-};
+export const parseAssets = (assets: Asset[]): any => assets.map((asset) => ({
```
jdnmr, line 83: Returned type can be `Asset[]` instead of `any`, because the Asset model contains an optional `code` field as far as I remember

src/utils/uniswap.js
```diff
@@ -66,12 +66,11 @@ export const getAskRate = (trade: Trade): string => {
   return excutionPriceBN.toFixed();
 };
 
-export const parseAssets = (assets: Asset[]) => {
-  assets.forEach(asset => {
-    asset.address = toChecksumAddress(asset.symbol === ETH ? WETH[chainId]?.address : asset.address);
-    asset.code = asset.symbol;
-  });
-};
+export const parseAssets = (assets: Asset[]): any => assets.map((asset) => ({
```
jdnmr, line 69: Same as above: `Asset[]` instead of `any`

jdnmr (approved):

PR #2350, opened 2020-11-25
Add Rari transfer flow

graszka22:


src/locales/en/common.json
```diff
@@ -1467,7 +1469,9 @@
       "youAreWithdrawingFromYieldPool": "You are withdrawing from Yield pool",
       "youAreWithdrawingFromEthPool": "You are withdrawing from ETH pool",
       "withdrawDetails": "Withdraw details",
-      "rariWithdrawalFee": "Rari withdrawal fee"
+      "rariWithdrawalFee": "Rari withdrawal fee",
+      "youAreTransfering": "You are transfering",
```
IAmKio, line 1473:
```suggestion
      "youAreTransferring": "You are transferring",
```

src/screens/Rari/RariInfo.js
```diff
@@ -101,17 +97,13 @@ const RariLogo = styled(CachedImage)`
...
   rates,
+  rariFundBalance,
 }: Props) => {
-  useEffect(() => {
-    fetchRariFundBalance();
-  }, []);
-
   const fiatCurrency = baseFiatCurrency || defaultFiatCurrency;
+  const totalRariFundBalance = (Object.values(rariFundBalance): any).reduce((sum, balance) => sum + balance, 0);
+  const totalRariFundBalanceInFiat = convertUSDToFiat(totalRariFundBalance, rates, fiatCurrency);
```
IAmKio, line 106: Just curious to know what happens if this fails? Do we have a fallback value or logic?

src/screens/Rari/RariTransfer.js
```diff
@@ -0,0 +1,352 @@
...
+    [amount, selectedContact, useGasToken],
+  );
+
+  useEffect(() => {
+    updateTxFeeDebounced();
+    return updateTxFeeDebounced.cancel;
+  }, [updateTxFeeDebounced]);
+
+  const handleFormSubmit = async () => {
+    if (submitPressed) return; // double press
```
IAmKio, line 150: Very nice 👍

IAmKio (changes requested):
Just a minor point but otherwise LGTM!

src/screens/Rari/RariTransfer.js
```diff
@@ -0,0 +1,352 @@
...
+  isEstimating,
+  estimateErrorMessage,
+  userDepositInRariToken,
+  contacts,
+  addContact,
+  estimateTransaction,
+  useGasToken,
+  balances,
+  resetEstimateTransaction,
+}) => {
```
wjkg, line 107: I'd only like to add that it looks like prop types are missing here: 
```suggestion
}: Props) => {
```

src/screens/Rari/RariInfo.js
graszka22: `rariFundBalance` is kept in redux store and persisted, so if fetching fund balance fails (or any other rari data, like user's balances, prices etc.), 0 will be showed or older value that was fetched recently.
Additionally every time it fails a toast is displayed like `Cannot fetch Rari data. Please try again` or graph error box, if graph call failed.

PR #2353, opened 2020-11-26
uniswap subgraph cached endpoint

poocart:
**Note: still waiting for DevOps to provide production url which is the only subject to change in this PR.**

Changes heavy Uniswap subgraph request/response into microservice direct request that returns cached and mapped version of Uniswap subgraph response.


src/actions/__tests__/authActions.test.js
```diff
@@ -165,6 +167,7 @@ describe('Auth actions', () => {
       .then(() => {
         const actualActions = store.getActions();
         expect(actualActions).toEqual(expectedActions);
+        expect(actualActions).toEqual(expectedActions);
```
zincoshine, line 170: Line 169 and 170 are the same?

jdnmr (approved):

src/actions/__tests__/authActions.test.js
poocart: 👍  mistype, fixed, thanks

zincoshine (approved):

PR #2354, opened 2020-11-26
Fix/app launch theme fix

poocart:
Fixes app launch crash when `withTheme` middleware is not ready for `Spinner` component.

Crash can be reproduced by disabling "JS Dev Mode" on React Native dev settings.


graszka22 (approved):
👍  I think you can also remove `withTheme` hoc in Spinner component on the bottom.

Edit: and the spinner is not placed correctly, it's now on the top of the screen on Android at least

PR #2359, opened 2020-12-02
Fix/1inch exchange address fix

poocart:
Includes:
1. Fix to include latest 1inch exchange contract address that's returned by 1inch exchange API V2.
2. Some chore around.

### DIFF
--- a/src/services/1inch.js
+++ b/src/services/1inch.js
@@ -156,7 +156,7 @@ export const create1inchAllowanceTx =
     const exchangeAddress = await get1InchSpenderAddress();
     if (!exchangeAddress) {
       reportLog('create1inchAllowanceTx -> get1inchApproveAddress failed');
-      return false;
+      return null;
     }
     return createAllowanceTx(fromAssetAddress, clientAddress, exchangeAddress);
   };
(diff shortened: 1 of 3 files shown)


src/services/1inch.js
```diff
@@ -59,11 +58,29 @@ const ethProvider = () => {
...
+};
+
 const getAllowanceSet = async (clientAddress: string, safeFromAddress: string, fromAsset: Asset) => {
   let allowanceSet = true;
   if (fromAsset.code !== ETH) {
     const assetContract = new ethers.Contract(safeFromAddress, ERC20_CONTRACT_ABI, ethProvider());
-    const allowance: BigNumber = await assetContract.allowance(clientAddress, EXCHANGE_ADDRESS);
+    const exchangeAddress = await get1InchSpenderAddress();
+    if (!exchangeAddress) {
+      reportLog('getAllowanceSet -> get1inchApproveAddress failed');
```
IAmKio, line 80: Could we report the error also coming back from `/approve/sender`? From what i can see being return in the `.catch` block from `get1InchSpenderAddress`, you can just report `exchangeAddress`?

jdnmr (approved):

IAmKio (approved):
Just one suggestion but it's optional in the interest of patching this issue 👍

src/services/1inch.js
poocart: The error is reported as part of response on `get1InchSpenderAddress`.

PR #2364, opened 2020-12-04
[PIL-766] Feature/WBTC.Cafe

jdnmr:
WBTC.Cafe integration. Can be tested with kovan/testnet BTC - let me know if you need a hand with this.

Because the exchange flow can be very long (usually at least an hour on mainnet, shorter on kovan), we're mocking an incoming WBTC transaction that sits in History like a regular pending tx.

Also refactored ExchangeConfirm screen - moved the table to a separate component.


src/screens/Exchange/RenFeeIcon.js
```diff
@@ -0,0 +1,66 @@
...
+  text-align: center;
+`;
+
+const RenFeeIconText = styled(BaseText)`
+  color: ${themedColors.inactiveTabBarIcon};
+  font-size: 10px;
+  text-align: center;
+`;
+
+const RenFeeIcon = () => {
```
graszka22, line 49: Just letting you know that this component (? button with a tooltip) repeats throughout the design so I've added it as a prop to `TableLabel` component.

src/screens/Exchange/WBTCCustomSlippageModal.js
```diff
@@ -0,0 +1,101 @@
...
+
+  return (
+    <Modal ref={modalRef}>
+      <Container>
+        <CustomTitle>{t('wbtcCafe.enterCustom')}</CustomTitle>
+        <TextInput
+          inputProps={inputProps}
+          numeric
+          hasError={!isValidCustomValue()}
+          errorMessage={t('error.amount.invalidNumber')}
```
graszka22, line 88: z-index issue 
Screenshot 2020-12-04 at 16 12 17

src/screens/Exchange/WBTCCustomSlippageModal.js
```diff
@@ -0,0 +1,101 @@
...
+
+  const isValidCustomValue = () => {
+    const converted = convertToNumber(customValue);
+    return (!!converted || converted === 0) && converted <= 100;
+  };
+
+  const handleCustomChange = (input: string) => {
+    if (!input) input = '0';
+    // 01 => 1, 00 => 0
+    if (input[1] && input[0] === '0' && !['.', ','].includes(input[1])) input = input.substring(1);
```
graszka22, line 66: When user types value with a comma, for example `1,5` he gets a NaN:
Screenshot 2020-12-04 at 16 16 55

src/actions/exchangeActions.js
```diff
@@ -379,3 +384,56 @@ export const getExchangeSupportedAssetsAction = (callback?: () => void) => {
...
+  });
+  dispatch(saveDbAction('pendingWbtcTransactions', { pendingWbtcTransactions }, true));
+};
+
+// for each new WBTC.Cafe tx detected, remove 1 pending
+export const removeWbtcPendingTxsAction = (quantity: number) => (dispatch: Dispatch, getState: GetState) => {
+  if (quantity < 1) return;
+  const { exchange: { data: { pendingWbtcTransactions: pending } } } = getState();
+  if (pending.length >= quantity) {
+    const updatedPendingTransactions = pending.slice(0, pending.length - quantity);
```
graszka22, line 436: What if we have several pending transactions? Do we have a way to know which one needs to be removed? Especially since I can add a transaction without actually sending anything

src/actions/exchangeActions.js
```diff
@@ -379,3 +384,56 @@ export const getExchangeSupportedAssetsAction = (callback?: () => void) => {
...
+      jsonrpc: '2.0',
+      method: 'ren_queryFees',
+      params: {},
+    }),
+  })
+    .then(res => res.json())
+    .then(res => {
+      dispatch({ type: SET_WBTC_FEES, payload: res.result });
+    })
+    .catch(e => reportErrorLog('Failed to fetch WBTC fees', e));
```
graszka22, line 403: Do we have any feedback for user in case of fail?

graszka22 (commented):
Hey, I tested only exchanging BTC to WBTC (the confirmation times are so looong 😢 ) and I've found a bunch of bugs:

1. The dark theme needs to be updated. Rari PR has a similar `Insight` with numbers, that is already updated, so you can skip that one. 
2. The footer should stick to the bottom? 

Screenshot 2020-12-04 at 15 30 33

3. When trying to convert 999 BTC (obviously I don't have that much I was just messing around :D) 

Screenshot 2020-12-04 at 15 32 36

4. The image is not centered horizontally

Screenshot 2020-12-04 at 15 35 48

5. On main btc screen, when history is empty, it shows just History subtitle. Imho it would be good to hide it if history is empty. Also the history should have a card-like style (there should be a prop for that)

6. The BTC -> WBTC has a broken navigation:
- I don't have any btc yet
- I want to convert 1 BTC
- I click "I've sent BTC"
- I'm brought back to Exchange screen o.o 
What is funny for me is that I have a new pending transaction now in the history even though I haven't sent anything - but as I understand there is no way to confirm that I've sent anything, right?

7. If I want to exchange less than 1 btc (for example 0.001 btc) it's just stuck like this:

Screenshot 2020-12-04 at 16 24 19

src/actions/exchangeActions.js
jdnmr: We can't tell this, there's no way we can match a mocked pending tx object to the actual transaction. That's why, for n new transactions, we're deleting n pending txs. There's also no way we can verify if the user actually sent the BTC, so once they press on "I've sent BTC", a mocked pending tx is generated (and it's deleted when a WBTC.Cafe tx arrives or after 12 hrs - whichever happens first).

That said, in the code snippet above we're removing pending txs from the end of their array, while we should delete from the beginning - I'll change that.

jdnmr:
> 2. The BTC -> WBTC has a broken navigation:
> 
> * I don't have any btc yet
> * I want to convert 1 BTC
> * I click "I've sent BTC"
> * I'm brought back to Exchange screen o.o
>   What is funny for me is that I have a new pending transaction now in the history even though I haven't sent anything - but as I understand there is no way to confirm that I've sent anything, right?

This is the desired flow. The wallet can't know if you have any BTC, or if you ever sent any to the address provided. So if a user presses "I've sent BTC", we trust them and assume they really did it (that's the purpose of having the button)- then we create a mock upcoming WBTC tx and navigate back, because there's no point in staying on this screen anymore. 

Perhaps we could use some better, clear way of informing the user that the process has begun - but that's more of a design issue. Also, maybe instead of navigating back to Exchange, we should navigate to WBTCCafe screen, where the user can see the mocked transaction straight away - what do you think? 
@graszka22

graszka22:
> Perhaps we could use some better, clear way of informing the user that the process has begun - but that's more of a design issue. Also, maybe instead of navigating back to Exchange, we should navigate to WBTCCafe screen, where the user can see the mocked transaction straight away - what do you think?

Yes, navigating back to Exchange was the main issue here, navigating to the WBTCCafe screen would be much more logical

src/actions/exchangeActions.js
jdnmr: Added handling this in Exchange screen

jdnmr:
> 1. If I want to exchange less than 1 btc (for example 0.001 btc) it's just stuck like this:
> 
> Screenshot 2020-12-04 at 16 24 19

Turns out that WBTC.Cafe will only work for amounts of 0.001 BTC or more. Added low amount as an error

IAmKio (commented):
**Great work!** This really is a game changer for the Pillar app and a heavily requested feature. I appreciate it's been a pain to get to this point so i really appreciate the efforts! 🚀  I've left a few questions - but other than that, from a code perspective - LGTM! Just incase there are any changes incoming, i'm just adding comments for now but if not - please re-request a review from me and we'll take it from there 👍

src/components/ValueInput/ValueInput.js
```diff
@@ -36,13 +36,14 @@ import Input from 'components/Input';
 import { Spacing } from 'components/Layout';
 import Modal from 'components/Modal';
 
-import { formatAmount, isValidNumber } from 'utils/common';
+import { formatAmount, isValidNumber, noop } from 'utils/common';
```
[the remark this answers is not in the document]
jdnmr: `noop` is a dummy function (`() => {}` or `() => null`), we've been using this pattern where Flow needs us to provide a function - so null wouldn't suffice

src/constants/exchangeConstants.js
```diff
@@ -41,7 +44,10 @@ export const SLOW = 'min';
 export const NORMAL = 'avg';
 export const FAST = 'max';
 
+// WBTC.Cafe
 export const ALLOWED_SLIPPAGE = 3; // 3%
+export const WBTC_PENDING_TRANSACTION = 'WBTC_PENDING_TRANSACTION';
+export const MIN_WBTC_CAFE_AMOUNT = 0.0011; // WBTC.Cafe won't work for lower amounts
```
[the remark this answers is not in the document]
jdnmr: I don't know that. This isn't even officially stated - the service just won't work for any lower amount. While their web app doesn't handle it in any way, I thought it would make sense to display an error message so that the user knows why it's not working. 

I doubt it will go lower, because the fees would be higher than the transaction's amount. We might want to check if it went higher every now and then though.

src/actions/exchangeActions.js
```diff
@@ -379,3 +384,55 @@ export const getExchangeSupportedAssetsAction = (callback?: () => void) => {
     dispatch(saveDbAction('exchangeSupportedAssets', { exchangeSupportedAssets }, true));
   };
 };
+
+export const getWbtcFeesAction = () => (dispatch: Dispatch) => {
+  fetch('https://lightnode-mainnet.herokuapp.com', {
```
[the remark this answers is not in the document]
jdnmr: I'll move this to `envConfig`, thanks

src/constants/exchangeConstants.js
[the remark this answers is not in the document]
IAmKio: Let's leave it as it is for now - we can always patch this later in the event that something does change.

src/actions/exchangeActions.js
```diff
@@ -379,3 +385,55 @@ export const getExchangeSupportedAssetsAction = (callback?: () => void) => {
     dispatch(saveDbAction('exchangeSupportedAssets', { exchangeSupportedAssets }, true));
   };
 };
+
+export const getWbtcFeesAction = () => (dispatch: Dispatch) => {
+  fetch(getEnv().WBTC_FEES_API, {
```
poocart, line 390: We stopped using `fetch` due timeouts not working and all the API calls are made with `axios` (except HTML content load).

poocart: Error is not thrown on `fetch` timeout as well, it will return as successful promise, you might want to change this to `axios` for more control.

src/actions/exchangeActions.js
```diff
@@ -379,3 +385,55 @@ export const getExchangeSupportedAssetsAction = (callback?: () => void) => {
...
+    .then(res => {
+      dispatch({ type: SET_WBTC_FEES, payload: res.result });
+    })
+    .catch(e => reportErrorLog('Failed to fetch WBTC fees', e));
+};
+
+export const getWbtcGatewayAddressAction = (params: WBTCGatewayAddressParams) =>
+  async (dispatch: Dispatch, getState: GetState, api: SDKWrapper): Promise<WBTCGatewayAddressResponse | null> => {
+    const { user: { data: user } } = getState();
+    const walletId = user?.walletId || '';
```
poocart, line 410: Will it go through if `walletId` is empty?

src/screens/Exchange/ConfirmationTable.js
```diff
@@ -0,0 +1,112 @@
...
+import Button from 'components/Button';
+import Icon from 'components/Icon';
+import { BaseText } from 'components/Typography';
+import { Spacing } from 'components/Layout';
+import type { OfferOrder } from 'models/Offer';
+import { ALLOWED_SLIPPAGE } from 'constants/exchangeConstants';
+import type { Theme } from 'models/Theme';
+
+type Props = {
+  errorMessage: string,
```
poocart, line 35:
```suggestion
  errorMessage: ?string,
```

src/screens/Exchange/WBTCCafeAddress.js
```diff
@@ -0,0 +1,91 @@
...
+// utils
+import { fontStyles } from 'utils/variables';
+import { themedColors } from 'utils/themes';
+
+// models
+import type { Theme } from 'models/Theme';
+
+import t from 'translations/translate';
+import WBTCCafeWarning from './WBTCCafeWarning';
+
```
poocart, line 38:
```suggestion

```

src/screens/Exchange/WBTCCafeAddress.js
```diff
@@ -0,0 +1,91 @@
...
+  ${fontStyles.regular};
+  color: ${({ color }) => color};
+`;
+
+type Props = {
+  amount: string;
+  theme: Theme;
+  address?: string;
+  error?: boolean;
+}
```
poocart, line 67:
```suggestion
type Props = {
  amount: string,
  theme: Theme,
  address?: string,
  error?: boolean,
};
```

src/screens/Exchange/WBTCCafeInfo.js
```diff
@@ -0,0 +1,235 @@
...
+
+type Props = {
+  wbtcData: ?WBTCFeesWithRate,
+  extendedInfo?: boolean,
+  navigation: NavigationScreenProp<*>,
+  amount?: string,
+  address?: string,
+  error?: boolean,
+  addWbtcPendingTx: (tx: PendingWBTCTransaction) => void,
+}
```
poocart, line 61:
```suggestion
};
```

src/screens/Exchange/WBTCCafeInfo.js
```diff
@@ -0,0 +1,235 @@
...
+import { EXCHANGE_CONFIRM, WBTC_CAFE } from 'constants/navigationConstants';
+import type { Dispatch } from 'reducers/rootReducer';
+import { LIGHT_THEME } from 'constants/appSettingsConstants';
+
+// partials
+import WBTCSlippageModal from './WBTCSlippageModal';
+import WBTCCafeAddress from './WBTCCafeAddress';
+import { ExchangeIcon, TableWrapper } from './ConfirmationTable';
+import RenFeeIcon from './RenFeeIcon';
+
```
poocart, line 52:
```suggestion


```

src/screens/Exchange/WBTCCafeAddress.js
poocart, line 38:
```suggestion


```

src/screens/Exchange/WBTCCafeAddress.js
```diff
@@ -0,0 +1,91 @@
...
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+/* eslint-disable i18next/no-literal-string */
```
poocart, line 20: I think this is not needed here?

```suggestion
```

src/screens/Exchange/WBTCCafeInfo.js
```diff
@@ -0,0 +1,235 @@
...
+  const handleNextPress = () => {
+    if (!extendedInfo && !!navigation) {
+      return navigation.navigate(EXCHANGE_CONFIRM, {
+        wbtcTxData: { maxSlippage: maxSlippage / 100, amount: Number(amount) },
+        wbtcEstData: wbtcData,
+      });
+    }
+    Clipboard.setString(address || '');
+    return Toast.show({ message: t('toast.addressCopiedToClipboard'), emoji: 'ok_hand' });
```
poocart, line 124:
```suggestion
    Toast.show({ message: t('toast.addressCopiedToClipboard'), emoji: 'ok_hand' });
```

src/screens/Exchange/WBTCCafeInfo.js
```diff
@@ -0,0 +1,235 @@
...
+  const {
+    wbtcData, extendedInfo, navigation, amount, address, error, addWbtcPendingTx,
+  } = props;
+
+  const handleNextPress = () => {
+    if (!extendedInfo && !!navigation) {
+      return navigation.navigate(EXCHANGE_CONFIRM, {
+        wbtcTxData: { maxSlippage: maxSlippage / 100, amount: Number(amount) },
+        wbtcEstData: wbtcData,
+      });
```
poocart, line 121:
```suggestion
      navigation.navigate(EXCHANGE_CONFIRM, {
        wbtcTxData: { maxSlippage: maxSlippage / 100, amount: Number(amount) },
        wbtcEstData: wbtcData,
      });
      return;
```

src/screens/Exchange/WBTCCafeInfo.js
```diff
@@ -0,0 +1,235 @@
...
+    return Toast.show({ message: t('toast.addressCopiedToClipboard'), emoji: 'ok_hand' });
+  };
+
+  const handleSent = () => {
+    addWbtcPendingTx({ amount: wbtcData?.estimate || 0, dateCreated: Date.now() });
+    navigation.navigate(WBTC_CAFE);
+  };
+
+  const getFeeNumber = () => {
+    if (!wbtcData) return 0;
```
poocart, line 133: User would be left behind, maybe we can show at least some Toast, wdyt?

src/screens/Exchange/WBTCCafeWarning.js
```diff
@@ -0,0 +1,58 @@
...
+`;
+
+const Text = styled(BaseText)`
+  ${fontStyles.regular};
+  color: ${({ color }) => color};
+`;
+
+interface Props {
+  theme: Theme;
+}
```
poocart, line 47:
```suggestion
type Props {
  theme: Theme,
};
```

src/screens/Exchange/WBTCCustomSlippageModal.js
```diff
@@ -0,0 +1,103 @@
...
+const CustomTitle = styled(MediumText)`
+  ${fontStyles.regular};
+  margin-bottom: 20px;
+`;
+
+interface Props {
+  theme: Theme;
+  onSubmit: (val: number) => void;
+  activeValue: number;
+}
```
poocart, line 50:
```suggestion
type Props {
  theme: Theme,
  onSubmit: (val: number) => void,
  activeValue: number,
};
```

src/screens/Exchange/WBTCSlippageModal.js
```diff
@@ -0,0 +1,176 @@
...
+  bottom: 0;
+`;
+
+// 0 is for custom
+const PRESET_VALUES = [0.5, 1, 3, 0];
+
+interface Props {
+  theme: Theme;
+  onModalWillHide: (val: number) => void;
+}
```
poocart, line 114:
```suggestion
type Props {
  theme: Theme,
  onModalWillHide: (val: number) => void,
};
```

src/screens/WBTCCafe/WBTCCafe.js
```diff
@@ -0,0 +1,185 @@
...
+  history: Transaction[];
+  pendingWbtcTransactions: PendingWBTCTransaction[];
+  accounts: Accounts;
+  fetchSmartWalletTransactions: () => void;
+  setWbtcPendingTxs: (txs: PendingWBTCTransaction[]) => void;
+}
+
+interface State {
+  showIntro: boolean;
+}
```
poocart, line 75: What's up with these `interfaces` Jan? 😂 

```suggestion
type Props = {
  navigation: NavigationScreenProp,
  hasSeenWbtcCafeIntro: boolean,
  updateHasSeenIntro: () => void,
  theme: Theme,
  history: Transaction[],
  pendingWbtcTransactions: PendingWBTCTransaction[],
  accounts: Accounts,
  fetchSmartWalletTransactions: () => void,
  setWbtcPendingTxs: (txs: PendingWBTCTransaction[]) => void,
};

type State = {
  showIntro: boolean,
}
```

src/screens/WBTCCafe/WBTCCafeIntro.js
```diff
@@ -0,0 +1,135 @@
...
+import { fontStyles, fontSizes } from 'utils/variables';
+import { themedColors } from 'utils/themes';
+import t from 'translations/translate';
+import Button from 'components/Button';
+
+interface Props {
+  theme: Theme;
+  onButtonPress: () => void;
+  backgroundColor: string;
+}
```
poocart, line 35:
```suggestion
interface Props {
  theme: Theme,
  onButtonPress: () => void,
  backgroundColor: string,
};
```

src/screens/WBTCCafe/WBTCCafe.js
```diff
@@ -0,0 +1,185 @@
...
+  height: 64px;
+  margin-vertical: 32px;
+  align-self: center;
+`;
+
+const logo = require('assets/images/exchangeProviders/wbtcLogo.png');
+
+const getBackgroundColor = (theme: Theme) => theme.colors.basic070;
+
+class WBTCCafe extends React.Component<Props, State> {
```
poocart, line 88: Just curious why you didn't choose function component with React hooks for this?

src/services/wbtcCafe.js
```diff
@@ -0,0 +1,117 @@
...
+    const curve = new Contract(
+      getEnv().WBTC_CURVE,
+      CURVE_ABI,
+      getEthereumProvider(isProdEnv ? 'homestead' : 'kovan'),
+    );
+    const renVMFee = Number(amount) * dynamicFeeRate;
+    const amountAfterMint =
+        Number(amount - renVMFee - fixedFee) > 0
+          ? Number(amount - renVMFee - fixedFee)
+          : 0;
```
poocart, line 64: This is the equal, but less noise:

```suggestion
    const amountAfterMint = Number(amount - renVMFee - fixedFee) || 0;
```

src/services/wbtcCafe.js
```diff
@@ -0,0 +1,117 @@
...
+
+  try {
+    const isProdEnv = getEnv().NETWORK_PROVIDER === 'homestead';
+    let exchangeRate;
+    let total;
+
+    const curve = new Contract(
+      getEnv().WBTC_CURVE,
+      CURVE_ABI,
+      getEthereumProvider(isProdEnv ? 'homestead' : 'kovan'),
```
poocart, line 58:
```suggestion
      getEthereumProvider(getEnv().NETWORK_PROVIDER),
```

poocart, line 58:
```suggestion
    let exchangeRate;
    let total;

    const curve = new Contract(
      getEnv().WBTC_CURVE,
      CURVE_ABI,
      getEthereumProvider(getEnv().NETWORK_PROVIDER),
```

src/services/wbtcCafe.js
```diff
@@ -0,0 +1,117 @@
...
+import t from 'translations/translate';
+/* eslint-disable i18next/no-literal-string */
+
+// much of this function's body is copy-pasted from wbtc.cafe web app code
+export const gatherWBTCFeeData = async (
+  amount: number,
+  fees: ?WBTCFeesRaw,
+  fromAssetCode: string,
+): Promise<?WBTCFeesWithRate> => {
+  if (!fees || !fromAssetCode) return null;
```
poocart, line 40: Do you think we might need to log this?

poocart (approved):
No critical issues just some general issues re logging and project code healthcare. Great job! 💯

src/screens/Exchange/WBTCCafeInfo.js
jdnmr: Good catch. We'll still display 0, but we'll show a Toast and disable going any further (i.e. won't display BTC address)

PR #2365, opened 2020-12-04
Add Rari claim fees

graszka22:


src/actions/transactionEstimateActions.js
```diff
@@ -64,6 +64,12 @@ export const setEstimatingTransactionAction = (isEstimating: boolean) => ({
   payload: isEstimating,
 });
 
+export const setEstimatingErrorAction = (errorMessage: string) => {
+  return (dispatch: Dispatch) => {
+    dispatch({ type: SET_TRANSACTION_ESTIMATE_ERROR, payload: errorMessage });
+  };
```
IAmKio, line 70: Could be:

```suggestion
  return (dispatch: Dispatch) => dispatch({ type: SET_TRANSACTION_ESTIMATE_ERROR, payload: errorMessage });
```

... maybe?

src/screens/Rari/RariClaimRgtReview.js
```diff
@@ -65,10 +66,28 @@ const RariClaimRgtReview = ({
...
+        Toast.show({
+          message: t('toast.rariFetchDataFailed'),
+          emoji: 'hushed',
+          supportLink: true,
+          autoClose: false,
+        });
+      }
+      setClaimFee(fee);
+      setIsFetchingFee(false);
+    }).catch(() => null);
```
IAmKio, line 85: Just wondering what happens here from a UX perspective? I think the returning of `null` is what is throwing me off...

src/utils/rari.js
```diff
@@ -837,12 +837,35 @@ export const getWithdrawalFeeRate = (rariPool: RariPool) => {
     });
 };
 
-export const getRariClaimRgtTransaction = (senderAddress: string, amount: number, txFeeInWei?: BigNumber) => {
+export const getClaimRtgFee = async () => {
+  const distributionStartBlock = 11094200;
+  const distributionPeriod = 390000;
+  const distributionEndBlock = distributionStartBlock + distributionPeriod;
+  const initialClaimFee = 0.33;
```
IAmKio, line 844: Just wondering, do any of these need to go into our envConfig / remote config?

```
  const distributionStartBlock = 11094200;
  const distributionPeriod = 390000;;
  const initialClaimFee = 0.33;
```

IAmKio (changes requested):
Just a few questions but great progress all in all! 👍

src/utils/rari.js
graszka22: No, because they're not going to change. Unless there is another liquidity mining program :D

IAmKio: Haha, fair! Thanks for clearing that up.

IAmKio (approved):

PR #2367, opened 2020-12-07
fix wallet connect navigation flows

poocart:
Fixes Wallet Connect transaction navigation where last screen doesn't belong to regular flow.


IAmKio (approved):
Interesting! Thank you for the fix ❤️

PR #2377, opened 2020-12-10
display error message in exchange confirm screen

poocart:
Fixes issue where error message is not displayed on exchange confirm screen and user is left behind with only disabled submit button.


IAmKio (approved):
Best fix ever.

PR #2393, opened 2020-12-21
Fix/Fiat on Exchange screen

jdnmr:
Fix several issues with "display fiat" option on Exchange screen: add error handling, limit fiat amounts to 2 decimals, add fiat updates for disabled input ("to" field).

### DIFF
--- a/src/components/ValueInput/ValueInput.js
+++ b/src/components/ValueInput/ValueInput.js
@@ -169,7 +169,7 @@ export const ValueInputComponent = (props: Props) => {
     if (displayFiatAmount) {
       const split = newValue.split('.');
       // only allow 2 decimals in fiat mode
-      if (split.length <= 2 && split[1] ? split[1].length <= 2 : true) {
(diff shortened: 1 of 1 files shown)


src/components/ValueInput/ValueInput.js
```diff
@@ -156,24 +156,29 @@ export const ValueInputComponent = (props: Props) => {
...
   const handleValueChange = (newValue: string) => {
-    let errorMessage = null;
     // ethers will crash with commas, TODO: we need a proper localisation
     newValue = newValue.replace(/,/g, '.');
     if (displayFiatAmount) {
-      setValueInFiat(newValue);
-      const convertedValue =
+      const split = newValue.split('.');
+      // only allow 2 decimals in fiat mode
+      if (split.length <= 2 && split[1] ? split[1].length <= 2 : true) {
```
graszka22, line 172: maybe it'd be bettter to avoid nested if like this?
`if(split.length <= 2 && !(split[1] && split[1].length > 2) {`

graszka22 (approved):

PR #2395, opened 2021-01-01
[PIL-846] Liquidity pools

graszka22:
A quick walkthrough on how liquidity pools work:
- you deposit some tokens in some fixed ratio to liquidity pools contract (that's adding liquidity), for now we only support uniswap liquidity pools. 
- the ratio  is calculated in fiat value, so for example to the pool 50% PLR and 50% ETH, you deposit a lot of PLR and a little ETH - because 1 ETH is worth much more than 1 PLR
- for deposit you obtain some pool tokens (for now it's UNI-V2) - it represents your share in the pool
-  you can burn pool tokens and obtain PLR+ETH back (that's removing liquidity) note that you can get a different amount of tokens that you've deposited (it depends on price changes)
- a GREAT resource on calculating how much tokens you pay and obtain is [here](https://hackmd.io/@HaydenAdams/HJ9jLsfTz), there is also [uniswap documentation](https://uniswap.org/docs/v2/smart-contracts/router02/)
- now you have a unipool contract that's only for rewards basically - that's a different contract than uniswap
- you can deposit some pool tokens (UNI-V2) to the unipool contract (that's staking)
- you can also get them back (that's unstaking)
- you can also of course claim the rewards for staking, the rewards are given by the contract creator and on kovan that's me 😄 

Some more remarks:
- liquidity pools fully work on kovan ❤️  (well, at least PLR+ETH works - I can add a second one but why bother). If you want PLRs or rewards just message me.
- the implementation of liquidity pools is designed so that it's flexible - it's supposed to be easy to add other liquidity pools. It isn't needed to add any additional screens, components or make drastic changes to existing ones. It should be also possible to have more tokens than 2 (for example 9.31% DAI + 39.36% USDC + 21.86% USDT + 29.46% PAX from the design) but I haven't tested that.
- for now there is ETH+PLR and PLR+DAI on mainnet and PLR+ETH on kovan
- to add a new uniswap+unipool liquidity pool just add its metadata in the constants and add liquidity pool token to supported tokens on the backend
- to add a completely new type of liquidity pool (curve.fi? some other pools that I'm not aware of?) add its metadata and the implementation to `utils/liquidityPools` and `actions/liquidityPoolsActions`. The components just give a pool metadata and liquidity pools redux store and expect the methods like `getPoolStats`, `get.*Transaction` etc. to conform to this "interface", handle different pools and just return the result.
- in the future maybe the metadata will not be hardcoded and will be fetched from the backend

Differences from the design:
- there is no [this](https://zpl.io/VkgDq4Z) screen - you can't change the amount you claim in unipool contract and to be honest it's not necessary, you always just cliam everything
- when you remove liquidity you don't have an option to get only one token - maybe it'll be added in the future

TODO:
- [x]  merge graph PR 

- [x]  add image in info screen (waiting for asset)

- [x] are the rewards still relevant? (edit: for now they're not, I've added a flag for this)


jdnmr (changes requested):
Github won't allow me to upload screenshots for some reason, but there are issues on smaller devices (iPhone 8 in my case):

- On add liquidity screen, keyboard and absolute card cover the bottom part of the screen; the screen should be scrollable;
- On add liquidity review, you get stuck and can't go any further for the same reason - you can't scroll down to the bottom;

Perhaps this is an issue on other screens too, but I can't go any further :)

jdnmr (approved):
LG! Just one tiny comment.

Also, after adding liquidity, nothing is displayed under "Active" (on Kovan), is this intentional?
Screenshot 2021-01-07 at 13 33 06

graszka22:
> Also, after adding liquidity, nothing is displayed under "Active" (on Kovan), is this intentional?

Yes, this is valid for now. The "Active"/"Available" tab shows all liquidity pools that you can add liquidity to, without those in purchased / staked tab. There will be design for this edge case soon.

PR #2402, opened 2021-01-12
Fix rari withdrawals

graszka22:
To reproduce try to withdraw max of currencies like BAT, DAI, PLR, LINK, 0x, etc (generally not in USD in the name :D). Sometimes the estimation will fail with "call failed transaction revert".

And I think one of the problems is with 0x exchange api. Each 0x order has fields `makerAssetAmount` (= how much output tokens we obtain from exchange) and `takerAssetAmount` (= how much input tokens we give to exchange). 
When we calculate max amount we can withdraw, we use 0x api and tell it "I want to sell x of input tokens". The exchange api replies "For x input tokens you'll get y output tokens" (it gives `makerAssetAmount` and `takerAssetAmount`).
Ok, so y tokens is the max amount we can obtain. We note that. 
User wants to withdraw the max amount. Now we ask 0x exchange "I want to buy this max amount of tokens". And sometimes the exchange replies that it needs more input than before. That is, the previous `takerAssetAmount` is slightly less than the current `takerAssetAmount` (even though current `makerAssetAmount` is the same or less than the previous one). I don't know why, maybe price changes, maybe rounding errors on 0x side 🤷‍♀️ . Anyway this leads to tx revert because we want to withdraw more than we can. And now make sure we don't withdraw more than we can.

The second problem is where I added this condition:
```
if (inputFillAmountUsdBN.gte(usdLeftBN.sub(scaleBN(16)))) {
...
   .mul(usdLeftBN.gt(inputFillAmountUsdBN) ? inputFillAmountUsdBN : usdLeftBN)
```
sometimes `usdLeftBN` is slightly larger than `inputFillAmountUsdBN` and calculating `thisOutputAmountBN` is for handling the case where  `inputFillAmountUsdBN` is larger than `usdLeftBN` due to `takerFee` (which is usually 0 as far as I see). This condition fixes that.

There is one more problem with the UI. 
- select bat
- select max value
- wait for the estimations etc
- select chainlink token
- there is tx revert but now it's due to the fact that value input value is larger than max value and there is a tooltip saying that's the case. So it's rather a minor ui issue rather than a bug. The fix is not in this PR.


src/utils/rari.js
```diff
@@ -526,6 +538,23 @@ export const getRariWithdrawTransactionData = async (
...
+            : token.address,
+          inputTokenBN,
+        ).catch(error => {
+          reportErrorLog("Rari service failed: Can't get 0x swap orders", { error });
+          return null;
+        });
+      }
+
+      if (!_0xdata) {
+        continue;
```
poocart, line 555: Not sure if we should log this or this is something too common?

poocart (approved):

src/utils/rari.js
graszka22: it'll be logged in `get0xSwapOrders` function

PR #2404, opened 2021-01-13
Liquidity pools fixes

graszka22:
1. try to add liquidity without uni-v2 asset added - you won't see updated balance and you can't stake tokens. Fixed by fetching balance from the uniswap subgraph.
2. Add liquidity and immediately check event details modal. Crash.


poocart (approved):
LGTM

PR #2416, opened 2021-01-15
PIL-912: Include services in total balance

wjkg:


graszka22 (approved):

graszka22:
LGTM!

poocart (approved):

src/models/Asset.js
```diff
@@ -60,6 +60,11 @@ export type Balance = {
   symbol: string,
 };
 
+export type MixedBalance = {
```
poocart, line 63: I'm not sure if we should add this as `MixedBalance` type or rather reuse existing `Balance` and when displayed convert those to numeric values if needed. By adding `MixedBalance` we build extra debt. Wdyt?

wjkg: It was for passing balances from the services, which could be either type depending on the service, so I wanted to delay conversion until necessary -- it could be removed, but then for services which keep their balance as a number it would mean converting number to string then back to number. Seems there would be more room for error that way.

PR #2421, opened 2021-01-15
[PIL-837] Wrap Recovery Portal access with feature flag

graszka22:


src/screens/Menu/WalletSettings.js
```diff
@@ -111,6 +117,7 @@ class WalletSettings extends React.Component<Props, State> {
       this.setState({ supportedBiometryType });
     });
     this.retrieveWalletObject();
+    isRecoveryPortalDisabled = firebaseRemoteConfig.getBoolean(REMOTE_CONFIG.RECOVERY_PORTAL_DISABLED);
```
jdnmr, line 120: Why `disabled` instead of `enabled`? If you look in `Services`, we tend to use `enabled` for feature flags everywhere (e.g. `isPoolTogetherEnabled`, `isSablierEnabled`...)

jdnmr (approved):

poocart (changes requested):
There's additional piece that needs to be wrapped which is Recovery Portal option during onboading.

src/screens/Menu/WalletSettings.js
graszka22: because the task specified the flag to be `disabled` https://linear.app/pillarproject/issue/PIL-837/wrap-recovery-portal-access-with-feature-flag

poocart (approved):

PR #2424, opened 2021-01-15
PIL-926 Fix/Exchange amounts

jdnmr:
A user was supposed to receive ~11k USD from a Uniswap exchange, but only received ~5k. We can't reproduce the issue, but we know that ~5k amount was provided to Uniswap contract as the minimum output and that our UI claimed the amount would be ~11k.
We calculate the exchange amounts twice: once to generate the offer, and once more to place the order. My guess is that the rates changed between the two calculations (we know that liquidity was dramatically low at the time of the event). This situation is very unlikely, but technically possible. Moreover, the amount displayed on `ExchangeConfirm` screen is passed in navigation params from the offer (and not the order), which explains why ~11k was displayed, but ~5k was provided to execute the contract method.
Changes in this PR make sure this can't happen again: if the Order receive amount is lower than Offer's, the swap will be stopped. 
Changes to `ExchangeConfirm` screen were not required because the user won't be navigated to the screen if the issue occurs.


poocart (approved):

src/utils/exchange.js
```diff
@@ -151,3 +152,22 @@ export const createAllowanceTx = async (
...
+  order: { expectedOutput?: string },
+): boolean => {
+  // no need to do anything if expectedOutput isn't provided - e.g. for Synthetix
+  if (!order.expectedOutput) return false;
+  // askRate is provided by offer
+  const offerAmount = calculateAmountToBuy(askRate, fromAmount);
+  const offerAmountBN = new BigNumber(offerAmount);
+  const orderAmountBN = new BigNumber(order.expectedOutput);
+  // round both values and stop swap if order < offer
+  return offerAmountBN.toFixed(8) > orderAmountBN.toFixed(8);
```
poocart, line 172: Maybe just `BN.lt`?

PR #2425, opened 2021-01-15
PIL-922 Fix/fetching exchange supported assets

jdnmr:
Calls for Uniswap supported assets frequently fail due to backend issues. A while ago I provided a fix where we fall back to store's `exchangeSupportedAssets` if a call fails. This left an edge case for new/reimported wallets: there are no assets to fall back to. 
This PR aims to solve this by setting exchange supported assets basing on 1 source only (i.e. 1inch in most cases) if there are no `exchangeSupportedAssets` in store.
While most commonly Uniswap fails and 1inch works fine, this covers the opposite scenario too.


src/actions/exchangeActions.js
```diff
@@ -376,11 +381,21 @@ export const getExchangeSupportedAssetsAction = (callback?: () => void) => {
...
+    if (fetchSuccess) {
+      fetchedAssetsSymbols = uniq(assetsSymbols[0].concat(assetsSymbols[1]));
+    } else if (fetchOneInchSuccess) {
+      fetchedAssetsSymbols = uniq(assetsSymbols[0]);
+    } else if (fetchUniswapSuccess) {
+      fetchedAssetsSymbols = uniq(assetsSymbols[1]);
+    }
 
-    const fetchedAssetsSymbols: string[] = fetchSuccess ? uniq(assetsSymbols[0].concat(assetsSymbols[1])) : [];
+    if (!fetchedAssetsSymbols.length) return;
```
poocart, line 396: Maybe send this case Sentry? Just monitor cases when users fail to fetch and doesn't have any set, wdyt?

poocart (approved):

PR #2438, opened 2021-01-19
feat: react native 0.63.4

mdjastrzebski:
Updated RN deps and configs based on diff between 0.61.5 -> 0.63.4 initial projects: https://react-native-community.github.io/upgrade-helper/?from=0.61.5&to=0.63.4

Most important changes:
- [x] Added flipper debugging by default
- [x] Switched `android:launchMode` to `"singleTask"` (from non-default `singleInstance`)
- [x] Resolved some runtime errors by patching (no minor library updates would solve these, either there were no release fixes or required major version update)
- [x] Resolved some runtime warnings
- [x] Updated snapshot tests


jdnmr (approved):
LG and seems to work fine!
What's with the giant patches? Do we hope to get rid of them anytime in the future?

mdjastrzebski:
@jdnmr The patches are actually quite simple, as they delete two files (CocoaAsyncSocket), but patch contains the content of these two files. I hope that dep update (react-native-tcp, react-native-udp) will solve it, but these are major version changes and have to be done carefully.

poocart (approved):
LGTM, well done! 👍  One note: maybe patches for `react-native-tcp` and `react-native-udp` can be either removed or the previous changes applies since now it contains patch files that basically are changes between dep versions.

PR #2442, opened 2021-01-19
Rari fixes

graszka22:
- fixes decimals issues
- adds a toast when there is insufficient liquidity on 0x
- removes dai from mstable (mstable currently doesn't support dai anyway, so this is rather a performance fix)


poocart (approved):
LGTM 👍

PR #2444, opened 2021-01-20
PIL-934: added archanova smart wallet sdk session healthcheck

poocart:
https://linear.app/pillarproject/issue/PIL-934/smart-wallet-session-not-renewing-after-expireddisconnected

While I was unable to exactly replicate the issue I take this as first fix that should at least start logging and/or if session/sdk(?) is lost – navigate user to PIN flow that will initialise Smart Wallet SDK again. 30 min check interval is something I came up with, but will double check with back-end on what's the actual lifespan of Archanova session token.

Additional thing discovered and fixed in this PR – Android background state change handler was not working. Changed that to completely rely on `AppState` events.


graszka22 (approved):
LGTM!

PR #2450, opened 2021-01-21
PIL-938 Fix/exchange max amount

jdnmr:
- Fix a bug when using the "max amount" button on Exchange screen didn't trigger search;
- When passing `fromAmount` around the exchange flow, we commonly parsed it from `string` to `number`, which made JS arbitrarily round it to meet the maximum digits limit. This caused minimal differences between desired amounts and the outputs. The differences were too small for users to notice, however it could ruin the "sell max" feature - a user could attempt to sell a higher amount of token than they actually had, which resulted in "transaction reverted".
- Because we no longer round `fromAmount`, users can sell exactly 100% of their token amount, leaving no dust, which is cool. 
- Handle edge case with users going wild with `fromAmount` decimals (>18);
- Also did some minor refactoring to the exchange flow and added more tests.


mdjastrzebski (approved):
Overall looks good.

As a general comment, probably for the future, it seems that we have a lot of repeated parsing from string/number to BigNumber and then some repeated checks with BigNumber. I wonder if we could replaced string/number amounts with BigNumber.

Alternatively we could also create our Amount class or module (based on BigNumber), to encapsulate common amount/price operations, so that they don't need to be inlined in each use-cases. Wdyt?

jdnmr:
> Overall looks good.
> 
> As a general comment, probably for the future, it seems that we have a lot of repeated parsing from string/number to BigNumber and then some repeated checks with BigNumber. I wonder if we could replaced string/number amounts with BigNumber.
> 
> Alternatively we could also create our Amount class or module (based on BigNumber), to encapsulate common amount/price operations, so that they don't need to be inlined in each use-cases. Wdyt?

I generally agree. One of the points of this PR was to eliminate numbers from the exchange flow and limit ourselves to BN and string. The problem is, api calls and some libs require strings, while any checks and calculations are obviously safer with BN. Over time, this resulted with our flow being quite messy in this area. I don't mind trying to refactor this further, but on the other hand, as far as I know we're thinking of moving the logic to the backend, in which case we should hold on with that. cc @IAmKio @poocart

poocart (approved):
PR looks good.

Re comments above – I do agree and especially I like the idea of wrapping number as class where different outputs/formats can be retrieved according to need. We need to keep this in mind for further development/maintenance @IAmKio.

PR #2453, opened 2021-01-25
Release Candidate 2.32.1 (patch)

poocart:
Release Candidate 2.32.1

Fixes Rari estimation requests throttle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,6 +170,7 @@
     "tcomb-form-native": "^0.6.11",
     "timers-browserify": "^1.0.1",
     "url": "~0.10.1",
+    "use-debounce": "^5.2.0",
     "util": "~0.10.3",
     "uuid": "^3.3.2",
     "vm-browserify": "0.0.4",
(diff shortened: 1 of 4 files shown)


mdjastrzebski (commented):
It seems that the way throttling is applied there is a change when user is typing long number, we will query Rari with intermediate input and not the final one. So the estimate might not be accurate. 

This is probably small problem if input is 0,1234 (intermediate) instead of 0,12345678 (final value), but potentially a bigger issues for integer numbers: 1234 (intermediate) vs 12345678 (final value).

I do not know enough context to know if this is feasible. So just raising this to your attention.

As alternative that should avoid this problem, we could apply throttle to text input field, so that no query is done until users stops typing digits.

jdnmr:
> It seems that the way throttling is applied there is a change when user is typing long number, we will query Rari with intermediate input and not the final one. So the estimate might not be accurate.
> 
> This is probably small problem if input is 0,1234 (intermediate) instead of 0,12345678 (final value), but potentially a bigger issues for integer numbers: 1234 (intermediate) vs 12345678 (final value).
> 
> I do not know enough context to know if this is feasible. So just raising this to your attention.
> 
> As alternative that should avoid this problem, we could apply throttle to text input field, so that no query is done until users stops typing digits.

@graszka22

poocart:
Guys, this was updated by @mdjastrzebski with debounce for input value.

mdjastrzebski (approved):
LGTM ;-)

PR #2455, opened 2021-01-26
[PLR-940] Hide PLR & related tokens from Rari

mdjastrzebski:
Quick solution for hiding PLR & PLR/ETH, PLR/DAI from Rari screens.


poocart (approved):
LGTM 👍

jdnmr (approved):

PR #2460, opened 2021-02-02
PIL-944 Fix/display amounts

jdnmr:
- Unify token amounts displaying across the app - decimals vary between different tokens;
- Special exception is made for deposits - for very small gains, I thought it would still be reasonable to display them (e.g. +0.001 DAI instead of rounding to 0) so that the users see they're actually earning something;
- Refactored `assetsConstants` and `Home` screen - moved parts to a separate component;
- Fixed Sablier `OutgoingStream` screen - button alignment was wrong.


src/utils/common.js
```diff
@@ -229,6 +222,9 @@ export const formatAmount = (amount: string | number, precision: number = 6): st
   return new BigNumber(roundedNumber).toFixed(); // strip trailing zeros
 };
 
+export const getDisplayedTokenAmount = (amount: string | number, assetSymbol: ?string): string =>
```
mdjastrzebski, line 225:
```suggestion
export const formatTokenAmount = (amount: string | number, assetSymbol: ?string): string =>
```
Consider renaming to improve readability (getXxx may suggest some computation, while formatXxx is very clear)

mdjastrzebski (commented):
LG, small suggestion about naming to improve readability

IAmKio (approved):

PR #2463, opened 2021-02-03
PIL-948: replace react-native-cached-image with react-native-fast-image

mdjastrzebski:
* Replaced all references to CachedImage with custom Image component using FastImage inside
* Remove dep to `react-native-cached-image` (and their deps like `react-native-fetch-blob`)

* Updated Fast Image version (minor update)
* Tested on iOS & Android


src/actions/authActions.js
```diff
@@ -493,9 +492,6 @@ export const resetAppServicesAction = () => {
     // reset data stored in keychain
     await resetKeychainDataObject();
 
-    // clear images cache
-    ImageCacheManager().clearCache().catch(() => null);
```
mdjastrzebski, line 497: FastImage does not contain method for forcing clearing cache. This should not be a problem, as AFAIK our images do not contain private information.

poocart (approved):
LGTM  👍

PR #2465, opened 2021-02-03
Fix/2fa pipeline apple issue

antonioned:
Fixes the 2FA issue we had in our ios build pipeline. I have removed the USER/PASSWORD combination and we will now use an App Store Connect API key to authenticate.

The key is stored in  private s3 buckets in our AWS accounts and fetched in our pipeline.


vnovacev (approved):

mdjastrzebski (approved):
LGTM

IAmKio (approved):
Lovely

PR #2466, opened 2021-02-03
PIL-941: fixed sign support for wc calls

poocart:
https://linear.app/pillarproject/issue/PIL-941/wallet-connect-and-collabland-bot-issue

Includes:
- Covers incompatibility between different versions of EIP1271 proposal.
- Added support for `eth_signTypedData` calls.
- Wallet Connect chore.

This works only with these changes in any Wallet Connect (WC) supported dapp: https://github.com/WalletConnect/walletconnect-example-dapp/pull/63/files.

We are in contact with Collab.Land to add support according to WC changes in PR above.


src/screens/WalletConnect/WalletConnectPinConfirm.js
```diff
@@ -164,6 +184,23 @@ class WalletConnectPinConfirmScreeen extends React.Component<Props, State> {
...
 
+  handleSignTypedData = async (request: CallRequest, wallet: Object) => {
+    const { approveCallRequest, rejectCallRequest } = this.props;
+    try {
+      const message = request.params[1]; // eslint-disable-line
+      const result = await signTypedData(message, wallet);
+      await approveCallRequest(request.callId, result);
+    } catch (error) {
+      await rejectCallRequest(request.callId, error.toString());
+    }
```
jdnmr, line 195: Shouldn't we log the error?

src/screens/WalletConnect/WalletConnectPinConfirm.js
```diff
@@ -164,6 +184,23 @@ class WalletConnectPinConfirmScreeen extends React.Component<Props, State> {
...
+    this.setState(
+      {
+        isChecking: false,
+      },
+      () => this.handleDismissal(),
+    );
```
jdnmr, line 201:
```suggestion
    this.setState({ isChecking: false }, this.handleDismissal)
```

jdnmr (approved):
Codewise LG, see just small comments

src/screens/WalletConnect/WalletConnectPinConfirm.js
poocart: Not sure because this can be simple reject on dapp end, but it also includes Private Key in signing  process which we don't want to expose by any mean 🤔

src/utils/wallet.js
```diff
@@ -82,19 +84,77 @@ export function signTransaction(trx: Object, walletInstance: Object): Promise<st
...
+
+  const actualMessage = ethers.utils.toUtf8String(messageHex);
+  const data = hashPersonalMessage(actualMessage);
+
+  return wallet.signMessage(ethers.utils.arrayify(data));
+}
+
+export function encodeTypedDataMessage(message: string): string {
+  const useV4 = true;
+  const data = TypedDataUtils.sanitizeData(JSON.parse(message));
```
IAmKio, line 130: Just wondering if `message` in `JSON.parse` is always parsable? JSON.parse fails and throws spectacularly when it cannot parse. Might be worth wrapping in a try / catch?

IAmKio (approved):
One small comment but other than that - great work!

src/utils/wallet.js
poocart: Yes, it's always ABI, otherwise it's failure on dapp end, but we can wrap for sure 👍

poocart: Actually, it's a chain of sign calls and the top method call (2 of them for separate chain of calls) that's waiting for result has exception wrapper so we're fine.

https://github.com/pillarwallet/pillarwallet/blob/99cf16c30426394fefb478af8348e3280b3de127/src/screens/WalletConnect/WalletConnectPinConfirm.js#L163-L202

jdnmr (approved):

src/utils/wallet.js
IAmKio: Thanks for the clarification! 👍

PR #2467, opened 2021-02-04
Fix/Uniswap env switching

jdnmr:
Because of how `envConfig` works, Uniswap didn't work well while using prod env on a staging build or vice-versa (if anyone does that). This PR changes `chainId` and `ethProvider` to be generated dynamically, which solves the issue.


src/utils/uniswap.js
```diff
@@ -24,28 +24,23 @@ import {
...
 import { ALLOWED_SLIPPAGE } from 'constants/exchangeConstants';
 
 import type { Asset } from 'models/Asset';
 
-const isMainnet = isProdEnv;
-export const chainId = isMainnet ? ChainId.MAINNET : ChainId.KOVAN;
+type UniswapChainId = ChainId.MAINNET | ChainId.KOVAN;
+
+export const getChainId = (): UniswapChainId => {
+  return getEnv().NETWORK_PROVIDER === 'homestead' ? ChainId.MAINNET : ChainId.KOVAN;
```
mdjastrzebski, line 39:
```suggestion
  return isProdEnv() ? ChainId.MAINNET : ChainId.KOVAN;
```
This check seems to be based on implementation detail. My suggestion would be to encapsulate the prod/staging env check into a app-wide function (either replacing isProdEnv or a new function) located around `configs/envConfig.js`.

src/utils/uniswap.js
```diff
@@ -24,28 +24,23 @@ import {
...
 import { ETH } from 'constants/assetsConstants';
 import { ALLOWED_SLIPPAGE } from 'constants/exchangeConstants';
 
 import type { Asset } from 'models/Asset';
 
-const isMainnet = isProdEnv;
-export const chainId = isMainnet ? ChainId.MAINNET : ChainId.KOVAN;
+type UniswapChainId = ChainId.MAINNET | ChainId.KOVAN;
+
+export const getChainId = (): UniswapChainId => {
```
mdjastrzebski, line 38:
```suggestion
export const getUniswapChainId = (): UniswapChainId => {
```
`getChainId` seems overly general name for uniswap-related function.

src/screens/Exchange/ExchangeConfirm.js
```diff
@@ -145,8 +145,8 @@ const ExchangeConfirmScreen = ({
   const fetchTransactionEstimate = () => {
     const isConvertedTx = isWethConvertedTx(symbol, contractAddress);
 
-    // for WETH converted txs on homestead, we need to provide ETH data or else estimation is always 0$
-    const contractAddressForEstimation = isProdEnv && isConvertedTx
+    // for WETH converted txs on homestead, we need to provide ETH data or else estimation fails
+    const contractAddressForEstimation = getEnv().NETWORK_PROVIDER === 'homestead' && isConvertedTx
```
mdjastrzebski, line 149:
```suggestion
    const contractAddressForEstimation = isProdEnv() && isConvertedTx
```
Consider chaning isProdEnv into function or adding a new function with different name. This check seems to really on implementation detail.

src/utils/exchange.js
```diff
@@ -105,7 +105,7 @@ export const parseOffer = (
 };
 
 export const isWethConvertedTx = (fromAssetSymbol: string, contractAddress: string): boolean => {
-  return fromAssetSymbol === ETH && contractAddress === WETH[chainId].address;
+  return fromAssetSymbol === ETH && contractAddress === WETH[getChainId()].address;
 };
```
mdjastrzebski, line 109: It seems like we have general purpose (exchange) `isWethConvertedTx` function, relying on uniswap specific `getChainId` function. Either we want to have both function under `uniswap` or both functions under `exchanges`. Wdyt?

mdjastrzebski (commented):
Overall looks good. Suggested some improvements:
* encapsulate prod env check into isProdEnv() fn instread of `NETWORK_PROVIDER === 'homestead'` check
* unclear relation between uniswap-specfic `getChainId()` and general exchange `isWethConvertedTx`.

src/utils/uniswap.js
jdnmr: I didn't want to change `isProdEnv` initially because it relied on build type, not env vars loaded. I thought there might have been some reason to that, but it seems it's safe to just check envConfig, not the build type. Changed to a function.

mdjastrzebski (approved):
LGTM! 👍🏻

PR #2469, opened 2021-02-04
PIL-950 issue additional approval reset transaction if needed

mdjastrzebski:
Fix occasional LP add liquidity error, when user has previously approved non-zero amount (but too small) to add to LP, and token contract (incl. our PLR token contract) rejects such `approve` call because it risking spending both old and new approve amount.


src/utils/liquidityPools.js
```diff
@@ -123,18 +123,33 @@ export const getAddLiquidityTransactions = async (
...
           symbol: ETH,
         },
         ...addLiquidityTransactions,
       ];
+
+      // ERC20 token contracts (incl. PLR itself) can throw error when trying to change non-zero approved allowance
+      // in order to prevent spedning both old and new allowance. As a way to mitigate it, we are first sending
+      // zeroing approve transaction.
+      // See: https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#IERC20-approve-address-uint256-
+      if (approvedAmountBN?.gt(0)) {
```
mdjastrzebski, line 141: I wonder if we should actually apply it only for PLR token. Wdyt?

jdnmr: As far as I understood the ERC20 docs you cited yesterday, it's good practice to do this even if the token's contract doesn't explicitly require it like PLR. So I'd leave it for all tokens, wdyt @poocart ?
```
Beware that changing an allowance with this method brings the risk that someone may use both the old and the new allowance by unfortunate transaction ordering. One possible solution to mitigate this race condition is to first reduce the spender's allowance to 0 and set the desired value afterwards
```

jdnmr (approved):
LG codewise. If there's no rush, I'd just wait for @poocart's opinion on your question before we go

poocart (approved):

PR #2470, opened 2021-02-05
PIL-951: fix: prevent estimation on too little token holding

mdjastrzebski:
Prevent calling estimate function if input is larger than available funds, as this always leads for estimation error with "execution reverted: Call failed".

* refactored from estimation call debouncing to input value debouncing (this allowed for onFormValid value to catch up)
* improved supported asset token check (old one did check only one token in pair)


src/screens/LiquidityPools/AddLiquidity.js
```diff
@@ -122,27 +122,24 @@ const AddLiquidityScreen = ({
...
-    calculateAddLiquidityTransactionEstimate(
-      pool,
-      assetsValues,
-      poolTokenAmount,
-      tokensData,
-    );
-  }, [assetsValues, fieldsValid]);
+    if (!tokensData.every((token) => token != null)) return;
+    if (!assetsValues.every((f) => !!parseFloat(f))) return;
+    if (!fieldsValid.every((f) => f)) return;
```
jdnmr, line 140: Can we squeeze these into one `if` statement? Also, `!=` => `!==`

jdnmr (approved):
LG! See one small comment

src/screens/LiquidityPools/AddLiquidity.js
mdjastrzebski: Re `!=` operator: `!= null` is equal to `!== null || !== undefined`, so that's intended.

mdjastrzebski: I've actually thought that separate simple `if(xxx) return`s where easier to read, but can change them to `if (xxx || yyy || zzz) return`, if you find it other way around.

PR #2471, opened 2021-02-05
PIL-949: Feature/update styled components

mdjastrzebski:
* Updated Styled Components to ver 5.x
* This version requires all css measures to use units (namely `px`) as in regular css otherwise it generates a lot of warnings
* Manual changes are mostly adding `px` units back to styles tagged strings
* However, SC5 also generates warning when spreading style objects without units (like regular RN styles) - in our case it's `fontStyles` and `getTextStyle`
* There were some migration changes: `innerRef`s => `ref`s, `font-family` in `"quotes"`.
* SC5 supports functional components so I could unroll some weird class components used only to wrap not supported FCs


src/components/Modals/SlideModal/SlideModal.js
```diff
@@ -327,21 +327,11 @@ class SlideModal extends React.Component<Props, State> {
...
-type ThemedProps = {| ...OwnProps, innerRef: Ref<SlideModal> |};
-const ThemedSlideModal: AbstractComponent<ThemedProps, SlideModal> = withTheme(SlideModal);
+const ThemedSlideModal = withTheme(SlideModal);
 
 // Exported because some calls to React.createRef need this to be passed as type parameter.
 export type SlideModalInstance = SlideModal;
 
 export default React.forwardRef<OwnProps, SlideModalInstance>((props, ref) => (
-  <ThemedSlideModal {...props} innerRef={ref} />
+  <ThemedSlideModal {...props} ref={ref} />
```
wjkg, line 336: It's good to see that most of this can finally be removed 😄 However, at the time of writing, there were two separate concerns:

1. `withTheme` wasn't typed, so any component it wrapped lost prop type validation. This is the reason for `AbstractComponent<...>` and still true after the update, which means the default export should still be annotated, only with `AbstractComponent<OwnProps, SlideModal>`.
2. The use of `innerRef` instead of `ref` by styled components. Since this is fixed in the new version, there's no reason to use `React.forwardRef` on our side, and the wrapper component can be removed completely.

To summarise: export `withTheme(SlideModal)` directly, but keep the type annotation.

src/components/SelectorOptions/SelectorOptions.js
```diff
@@ -490,11 +490,8 @@ class SelectorOptions extends React.Component<Props, State> {
...
 
-const ThemedSelectorOptions = (withTheme(SelectorOptions): React.AbstractComponent<{|
-    ...OwnProps,
-    innerRef: { current: null | SelectorOptions } | (null | SelectorOptions) => mixed,
-|}>);
+const ThemedSelectorOptions = withTheme(SelectorOptions);
 
 export default React.forwardRef<OwnProps, SelectorOptions>((props, ref) => {
-  return <ThemedSelectorOptions {...props} innerRef={ref} />;
+  return <ThemedSelectorOptions {...props} ref={ref} />;
```
wjkg, line 496: The same thing as with `<SlideModal>` applies here.

wjkg (changes requested):

wjkg (approved):

PR #2473, opened 2021-02-08
PIL-952: Fix liquidity pool inexact max value

mdjastrzebski:
Fix using InputValue max send feature which frequently caused transaction estimation errors because of rounding issues. Current logic migrates the key calculations to BigNumber.

Related screens:
- [x] Remove liquidity
- [x] Stake/Unstake liquidity

Additionally:
- [x] Properly handling remove liquidity when use max pair token amount (not only pool amount)
- [x] Fixed LP list filter for pool with both staked and unstaked amount


src/screens/LiquidityPools/LiquidityPoolDashboard.js
```diff
@@ -191,16 +191,20 @@ const LiquidityPoolDashboard = ({
...
-  const balance = poolStats.userLiquidityTokenBalance;
+  const balance = poolStats.userLiquidityTokenBalance.toNumber();
   const formattedBalance = formatTokenAmount(balance, pool.symbol);
   const hasBalance = balance > 0;
 
   const fiatCurrency = baseFiatCurrency || defaultFiatCurrency;
   const fiatBalance = formatFiat(convertUSDToFiat(balance * poolStats.currentPrice, rates, fiatCurrency), fiatCurrency);
-  const stakedAmountInFiat = convertUSDToFiat(poolStats.stakedAmount * poolStats.currentPrice, rates, fiatCurrency);
+  const stakedAmountInFiat = convertUSDToFiat(
+    poolStats.stakedAmount.toNumber() * poolStats.currentPrice,
```
jdnmr, line 201: Again, is `toNumber()` safe here?

src/screens/LiquidityPools/LiquidityPools.js
```diff
@@ -288,7 +288,8 @@ const LiquidityPoolsScreen = ({
 
   const renderPurchasedPool = ({ item: pool, index }) => {
     const poolStats = poolsStats[index];
-    const balance = poolStats.userLiquidityTokenBalance;
+    const balance = poolStats.userLiquidityTokenBalance.toNumber();
```
jdnmr, line 291: `toNumber`?

src/screens/LiquidityPools/LiquidityPools.js
```diff
@@ -299,13 +300,13 @@ const LiquidityPoolsScreen = ({
...
     const { currentPrice } = poolStats;
-    const stakedAmountInFiat = convertUSDToFiat(currentPrice * poolStats.stakedAmount, rates, fiatCurrency);
+    const stakedAmountInFiat = convertUSDToFiat(currentPrice * poolStats.stakedAmount.toNumber(), rates, fiatCurrency);
     const formattedStakedAmount = formatFiat(stakedAmountInFiat, fiatCurrency);
     const tokenSymbol = pool.rewards[0].symbol;
 
     return (
       <TouchableOpacity onPress={() => goToPoolDashboard(pool)}>
-        {renderPoolListItem(pool, poolStats.stakedAmount, formattedStakedAmount)}
+        {renderPoolListItem(pool, poolStats.stakedAmount.toNumber(), formattedStakedAmount)}
```
jdnmr, line 309: More `toNumber`s :-)

src/screens/LiquidityPools/RemoveLiquidity.js
```diff
@@ -146,54 +147,53 @@ const AddLiquidityScreen = ({
...
+    setObtainedAssetsValues(amounts.pairTokens.map((amount, i) =>
+      i === tokenIndex ? newValue : formatAmount(amount),
+    ));
   };
 
   const renderTokenInput = (tokenIndex: number) => {
     const poolTokenSymbol = poolTokenData?.symbol;
     if (!poolTokenSymbol) return null;
-    const maxAmountBurned = poolStats?.userLiquidityTokenBalance || 0;
+    const maxAmountBurned = poolStats?.userLiquidityTokenBalance.toNumber() ?? 0;
```
jdnmr, line 174: `toNumber`?

src/selectors/balances.js
```diff
@@ -135,9 +135,9 @@ const liquidityPoolsBalanceListSelector = createSelector(
       stakedAmount,
     } = getPoolStats(pool, liquidityPoolsState) ?? {};
 
-    if ([currentPrice, userLiquidityTokenBalance, stakedAmount].includes(undefined)) return null;
+    if (currentPrice === undefined) return null;
     return {
-      balance: (userLiquidityTokenBalance + stakedAmount) * currentPrice,
+      balance: (userLiquidityTokenBalance.toNumber() + stakedAmount.toNumber()) * currentPrice,
```
jdnmr, line 140: So are we sure `userLiquidityTokenBalance` and `stakedAmount` can never be `undefined`? Also, shouldn't we use BN to calculate?

jdnmr (commented):
LG codewise. I'm just wondering if it's safe to use JS numbers for calculations here and there - see comments

src/components/ValueInput/ValueInput.js
```diff
@@ -195,9 +195,9 @@ export const ValueInputComponent = (props: Props) => {
     );
     const maxValueInFiat = getBalanceInFiat(fiatCurrency, newMaxValue, ratesWithCustomRates, assetSymbol);
     if (percent === 100) {
-      onValueChange(newMaxValue);
+      onValueChange(newMaxValue, percent);
     } else {
-      onValueChange(toFixedString(parseFloat(newMaxValue) * (percent / 100)));
+      onValueChange(toFixedString(parseFloat(newMaxValue) * (percent / 100)), percent);
```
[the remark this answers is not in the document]
mdjastrzebski: The goal here was to expose the underlying percent "action" from the user, as an escape hatch for situations like that. I would keep it as is, because it's fairly safe change for the rest of codebase.

src/screens/Home/Home.js
```diff
@@ -411,7 +411,7 @@ class HomeScreen extends React.Component<Props> {
 
   renderLiquidityPool = ({ item: { pool, poolStats } }) => {
     const { rates, baseFiatCurrency } = this.props;
-    const tokenBalance = poolStats.userLiquidityTokenBalance + poolStats.stakedAmount;
+    const tokenBalance = poolStats.userLiquidityTokenBalance.toNumber() + poolStats.stakedAmount.toNumber();
```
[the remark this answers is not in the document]
mdjastrzebski: This call-site is for rough calculation only, so should be safe here.

src/screens/LiquidityPools/RemoveLiquidity.js
```diff
@@ -146,54 +147,53 @@ const AddLiquidityScreen = ({
...
+  const onObtainedAssetValueChange = (tokenIndex: number, newValue: string, newPercent: number | void) => {
+    const amounts = calculateProportionalAssetAmountsForRemoval(pool, newValue || '0', tokenIndex, liquidityPoolsState);
 
-    const formattedAssetsValues = assetsAmounts
-      .map((amount, i) => i === tokenIndex ? newValue : formatAmount(amount));
-    setObtainedAssetsValues(formattedAssetsValues);
-  };
+    // Ensure we will withdraw everything if user want to remove max value.
+    if (newPercent === 100) {
+      onPoolTokenAmountChange(poolStats?.userLiquidityTokenBalance.toFixed() ?? '0');
```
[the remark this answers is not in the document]
mdjastrzebski: This `?? '0'` is to satisfy Flow null check and should happen only when `poolStats` are not available. IMO I would leave it as is, because there is no harm to the user.

src/screens/LiquidityPools/LiquidityPoolDashboard.js
mdjastrzebski: This is only a rough calculation, no harm to be done here.

src/screens/LiquidityPools/LiquidityPools.js
mdjastrzebski: This is only a rough calculation, no harm to be done here.

src/screens/LiquidityPools/LiquidityPools.js
mdjastrzebski: This is only a rough calculation, no harm to be done here.

src/selectors/balances.js
mdjastrzebski: Re null checks: I're right, these could happen for `null` `poolStats` getting coalesced to `{}`. I've re-added them

mdjastrzebski: Re: BN calculation, these are rough calculation for fiat display, so I left them in `number` calculation.

src/screens/LiquidityPools/RemoveLiquidity.js
mdjastrzebski: This is a hard one, as it's hard to get 100% right. My aim is to cover 95% expected use cases.

Expected use cases:
* withdraw 100% amount 
* withdraw partial amount of given pair token value (e.g. withdraw 1 ETH of pool tokens) - while giving user rough estimate of allowed pair token withdrawal

Key findings:
* *maxAmountBurned* is feed into `ValueInput` for *pair* tokens
* `ValueInput` uses `number` math for checking allowed values (`Number(assetBalance) < Number(amount)`)
* I did not want to change `ValueInput` math as it might impact rest of the app
* for 100% withdrawals the pair token input will be truncated to 6 decimal places, which should be below any `number` math rounding down.

@jdnmr wdyt?

jdnmr (approved):
LG! We just need to remember to never use the JS-number-based values for contract executions/API calls

PR #2475, opened 2021-02-09
Fix flow typing for BigNumber.js

mdjastrzebski:
Scope:
* Added flow-typed definitions for BigNumber.js (before that it was treated as `any`)
* Fixed most type errors
* Ignored some remaining errors without clear resolution

Types of changes:
* type changes - these are safe, as they do not change the runtime behaviour
* code changes - this could cause runtime issues, so I've only changed code that seem safe to change  but **please check them carefully**.


src/actions/transactionEstimateActions.js
```diff
@@ -129,7 +129,7 @@ export const estimateTransactionAction = (
       feeInfo = buildTxFeeInfo(estimated, useGasToken);
     }
 
-    if (!errorMessage && feeInfo && !feeInfo.fee.gt(0)) {
+    if (!errorMessage && feeInfo && !feeInfo.fee?.gt(0)) {
```
mdjastrzebski, line 132: Code analysis seem to indicate that null `fee` means that estimation has failed.

src/screens/Exchange/utils.js
```diff
@@ -200,7 +200,7 @@ export const getHeaderRightItems = (
 const isEnoughAssetBalance = (assetBalance: ?string, amount: string): boolean => {
   try {
     const amountBN = new BigNumber(amount);
-    const balanceBN = new BigNumber(assetBalance);
+    const balanceBN = new BigNumber(assetBalance ?? 0);
```
mdjastrzebski, line 203: This change assumes that missing asset balance is equal to balance of 0 which will lead to returning false for all non-zero `amount` values. Previously this was BigNumber(NaN)

src/screens/SendCollectible/SendCollectibleConfirm.js
```diff
@@ -164,7 +164,7 @@ const SendCollectibleConfirm = ({
 
   if (feeInfo) {
     // rinkeby testnet fee check
-    const txFee = utils.formatEther(feeInfo.fee.toString());
+    const txFee = utils.formatEther(feeInfo.fee?.toString() ?? 0);
```
mdjastrzebski, line 167: Previously this threw error iff `fee` == `null`

src/services/uniswap.js
```diff
@@ -163,7 +163,7 @@ export const getUniswapOffer = async (
   const askRate = expectedOutput.dividedBy(quantityBN);
   const allowanceSet = await getAllowanceSet(clientAddress, fromAssetParsed);
 
-  return parseOffer(fromAssetParsed, toAssetParsed, allowanceSet, askRate, PROVIDER_UNISWAP);
+  return parseOffer(fromAssetParsed, toAssetParsed, allowanceSet, askRate.toFixed(), PROVIDER_UNISWAP);
```
mdjastrzebski, line 166: All other parseOffer invocations passed string, produced by `toFixed()` or `toFixed(n)` calls.

src/utils/assets.js
```diff
@@ -199,7 +199,7 @@ export const isEnoughBalanceForTransactionFee = (
     }
   }
 
-  const txFeeInWeiBN = new BigNumber(txFeeInWei.toString()); // compatibility
+  const txFeeInWeiBN = new BigNumber(txFeeInWei?.toString() ?? 0); // compatibility
```
mdjastrzebski, line 202: Otherwise would throw for null txFeeInWei

src/services/smartWallet.js
```diff
@@ -299,7 +299,7 @@ class SmartWallet {
       });
   }
 
-  getAccountStakedAmount(tokenAddress: ?string): BigNumber {
+  getAccountStakedAmount(tokenAddress: ?string): Promise<BigNumber> | BigNumber {
```
jdnmr, line 302: Isn't `Promise<BigNumber>` enough?

src/utils/assets.js
```diff
@@ -137,7 +137,7 @@ export const getFormattedRate = (
 export const calculateMaxAmount = (
   token: string,
   balance: number | string,
-  txFeeInWei: BigNumber,
+  txFeeInWei: ?(BigNumber | string | number),
```
jdnmr, line 140: We have a type `Value` for this in `utils/common`

src/utils/assets.js
```diff
@@ -164,7 +164,7 @@ export const calculateMaxAmount = (
 export const isEnoughBalanceForTransactionFee = (
   balances: Balances,
   transaction: {
-    txFeeInWei: number,
+    txFeeInWei: ?(BigNumber | string | number),
```
jdnmr, line 167: You can use `Value` from `utils/common`

src/utils/common.js
```diff
@@ -182,7 +182,7 @@ export const formatMoney = (
   return (c ? num.replace('.', c) : num).replace(new RegExp(re, 'g'), `$&${s || ','}`);
 };
 
-export const parseNumber = (amount: string = '0') => {
+export const parseNumber = (amount: BigNumber | string | number = '0') => {
```
jdnmr, line 185: `Value`... and in other places in this file too :-)

jdnmr (approved):
LG, just a few small comments

src/services/smartWallet.js
mdjastrzebski: No, as next line there is an early return ` if (!tokenAddress) return new BigNumber(0);` not wrapped in promise.

src/utils/assets.js
mdjastrzebski: Fixed

PR #2478, opened 2021-02-09
feature/appcenter-etherspot-builds

IAmKio:
- Added branch to Circle CI jobs responsible for testing, building for App Centre.

@antonioned could you check that this was all i needed to do? Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -747,6 +747,7 @@ workflows:
               ignore:
                 - develop
                 - master
+                - feature/etherspot
                 
   build_and_deploy_appcenter:
     jobs:
@@ -756,6 +757,7 @@ workflows:
             branches:
               only:
                 - develop
+                - feature/etherspot
       - appcenter_ios:
           requires:
             - build-and-test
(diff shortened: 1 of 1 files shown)


antonioned:
@IAmKio since this is a PR towards develop, it will anyhow trigger appcenter builds... I thought you wanted to trigger them only from the feature branch, is that correct?

IAmKio:
@antonioned so just a bit of context around the `feature/etherspot` branch - this is quite a big architectural change within the app and it won't be merged into develop for quite some time - however we need to start getting the code out to testers with access to App Center sooner than later to test the architectural changes. 

I've just realised something... As i understand, with every App Center build coming from `feature/etherspot` - we get a new, Etherspot version of the app which means when someone merges to develop next time, we end up with a non-Etherspot build. Is there a way we can differentiate this maybe?

poocart (approved):

poocart:
@IAmKio we can differentiate by version and I remember AppCenter supports downgrade so we can even test full update from SW to Etherspot.

PR #2480, opened 2021-02-10
Add a separate build-and-test-etherspot job to set different etherspo…

antonioned:
Since we set our build numbers in the `build-and-test` job, turns out we only need 1 additional job in order to tag the builds with `-etherspot`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,6 +208,7 @@ aliases:
 
 jobs:
 
+  # Run tests and set build numbers
   build-and-test:
     executor: node10-executor
     steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -745,10 +784,35 @@ workflows:
           filters:
             branches:
               ignore:
+                - feature/etherspot
                 - feature/etherspot-transactions
```
antonioned, line 788: @IAmKio @poocart I was not sure about this branch in ignore, is this maybe a leftover and can be removed?

vnovacev (approved):

IAmKio (approved):

PR #2485, opened 2021-02-15
PIL-959: Refactor/restore react native flow typing

mdjastrzebski:
Scope:
- [x] Remove flow config ignores for React Native
- [x] Fix most type issues
- [x] Disable remaining type issues
- [x] Encapsulate `react-native/Libraries` type imports as they refer to implementation details

I've tried to avoid any non-idiomatic code changes, in order to minimise potential runtime issues. Most changes relate to types or are idiomatic and safe. Occasionally type-checking unveiled a clear issues and I've fixed these as well.


src/utils/ui.js
```diff
@@ -73,7 +74,7 @@ export function hexToRgba(hex: string, opacity: ?number) {
     : null;
 }
 
-export const measure = (ref: View): Promise<Measurements> =>
-  new Promise(resolve => ref.measureInWindow((x, y, w, h) => resolve({
+export const measure = (ref: ?React.ElementRef<typeof View>): Promise<Measurements> =>
+  new Promise(resolve => ref?.measureInWindow((x, y, w, h) => resolve({
```
wjkg, line 78: I'm not sure if `ref` being set to `undefined` actually occurs in practice, but if it does then the returned promise will hang where previously it would throw. Looking at usages of `measure`, nothing would change, but rejected promises are expected, so it's probably best to return one in that case for consistency.

wjkg (approved):
I've run into the issue with the lack of RN types a few times, so I'm glad it's being fixed 🎉. Good idea with the proxy file for the types from `react-native/Libraries`. I have some light suggestions, but otherwise looks fine.

src/utils/ui.js
mdjastrzebski: removed nullable check

src/screens/Menu/CommunitySettings.js
```diff
@@ -62,6 +62,7 @@ class CommunitySettings extends React.Component<{}, State> {
       Linking.openURL(link.web);
     } else {
       Linking.canOpenURL(link.app)
+        // $FlowFixMe: already checked
         .then(supported => supported ? Linking.openURL(link.app) : Linking.openURL(link.web))
         .catch(() => {});
```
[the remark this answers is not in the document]
mdjastrzebski: fixed as suggested

PR #2491, opened 2021-02-18
PIL-967: coingecko and rari eth-usd rate fixes

poocart:
https://linear.app/pillarproject/issue/PIL-967/rari-data-failed

Includes:
1. Added check for failed Coingecko ETH price to not include it in updated rates if it failed.
2. Increased Coingecko timeout as it seems this has been common report on Sentry.
3. Added check to Rari ETH pool balance calculation for ETH/USD rate if it's present, however, it should occur anymore based on this PR 1st fix. I didn't add extra logging over ETH/USD rate missing because this would be extra noise and this should never repeat based on 1st fix.

### DIFF
--- a/src/services/coinGecko.js
+++ b/src/services/coinGecko.js
@@ -44,7 +44,7 @@ const BTC_ID = 'bitcoin';
 const WBTC_ID = 'wrapped-bitcoin';
 
 const requestConfig = {
-  timeout: 5000,
+  timeout: 10000,
   headers: {
     Accept: 'application/json',
     'Content-Type': 'application/json',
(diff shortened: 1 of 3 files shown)


src/services/coinGecko.js
```diff
@@ -44,7 +44,7 @@ const BTC_ID = 'bitcoin';
 const WBTC_ID = 'wrapped-bitcoin';
 
 const requestConfig = {
-  timeout: 5000,
+  timeout: 10000,
```
IAmKio, line 47: If we remove the timeout parameter altogether, i believe this will result in the request continuously trying... is this a better behaviour than a 10 second timeout maybe?

mdjastrzebski (approved):
LG

IAmKio (approved):

src/services/coinGecko.js
poocart: Yes, if we remove it it will do it until gets dropped by native handler, but as you're saying, it would go for longer, however, downside is that we lose control here.

PR #2494, opened 2021-02-19
PIL-969 Fix/home performance

mdjastrzebski:
Summary:
Improve home screen performance when user has a long list of past activities. This change reduces activity feed processing from ~2500ms to just ~20ms for ~500 activity item entries.

Scope:
* optimise used data structure in `groupSectionsByDate`
* optimise `Date` operations to take into account observed issues with `Date` operations on iOS/JavaScriptCore
  * very fast `Date.format`
  * very slow `Date` mutation ops like `setHours` and hence `startOfDay`, `isSameDay`, `isToday`, `isYesterday`, etc
* remove unnecessary memoization in ActivityFeed (I did measure timing in actual app and noticed no improvements)
* delay component did mount optimisations in EventDetails to allow smooth animation
* fix css/color-related crash & warnings

### DIFF
--- a/src/components/EventDetails/EventDetails.js
+++ b/src/components/EventDetails/EventDetails.js
@@ -1633,7 +1633,7 @@ export class EventDetail extends React.Component<Props> {
         {!!title && (
           <DetailRow color={titleColor}>
             {title}
-            {!!actionIcon && <ActionIcon name={actionIcon} iconColor={statusIconColor} />}
+            {!!actionIcon && <ActionIcon name={actionIcon} iconColor={this.getColor(statusIconColor)} />}
           </DetailRow>
         )}
         {customActionTitle}
--- a/src/components/EventDetails/EventDetails.js
+++ b/src/components/EventDetails/EventDetails.js
@@ -19,7 +19,7 @@
 */
 
 import * as React from 'react';
-import { View, Linking } from 'react-native';
+import { View, Linking, InteractionManager } from 'react-native';
 import { connect } from 'react-redux';
 import { createStructuredSelector } from 'reselect';
 import styled, { withTheme } from 'styled-components/native';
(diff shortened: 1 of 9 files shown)


src/components/EventDetails/EventDetails.js
```diff
@@ -361,13 +361,15 @@ export class EventDetail extends React.Component<Props> {
...
-    const { event } = this.props;
-    const { type } = event;
-    if (!(type === TRANSACTION_EVENT || type === COLLECTIBLE_TRANSACTION)) return;
-    const txInfo = this.findTxInfo(event.type === COLLECTIBLE_TRANSACTION);
-    if (!txInfo) return;
-    this.syncEnsRegistry(txInfo);
-    this.syncTxStatus(txInfo);
+    InteractionManager.runAfterInteractions(() => {
+      const { event } = this.props;
+      const { type } = event;
```
jdnmr, line 366:
```suggestion
      const { type } = this.props.event;
```

src/components/EventDetails/EventDetails.js
```diff
@@ -361,13 +361,15 @@ export class EventDetail extends React.Component<Props> {
...
+      if (!(type === TRANSACTION_EVENT || type === COLLECTIBLE_TRANSACTION)) return;
+      const txInfo = this.findTxInfo(event.type === COLLECTIBLE_TRANSACTION);
```
jdnmr, line 368:
```suggestion
      const txInfo = this.findTxInfo(type === COLLECTIBLE_TRANSACTION);
```

src/screens/Tank/SettleBalance/SettleBalance.js
```diff
@@ -245,7 +241,7 @@ class SettleBalance extends React.Component<Props, State> {
     } = this.props;
     const { txToSettle } = this.state;
     const showSpinner = !isFetched;
-    const formattedFeedData = groupAndSortByDate(availableToSettleTx, 0);
+    const formattedFeedData = groupSectionsByDate(availableToSettleTx, 1);
```
jdnmr, line 244: Why did 0 change to 1 here?

jdnmr (approved):
LGTM, great optimizations! See just minor suggestions and question

src/screens/Tank/SettleBalance/SettleBalance.js
mdjastrzebski: `groupSectionsByDate` second param is the unit multiplier for date field. Its default value is set to 1000, meaning it expects the date field to contain value in seconds (since epoch), then multiplying it by 1000 to get milliseconds, which are feed into `Date` constructor.

In this case however, the date field already contains value in milliseconds, so the multiplier needs to be 1. The previous code contained a check that did no multiplication when unit multiplier equals 0. I've simplified the code, by removing zero check and using unit multiplier value of 1, which should give the same effect.

src/components/EventDetails/EventDetails.js
mdjastrzebski: Fixed

src/components/EventDetails/EventDetails.js
mdjastrzebski: Fixed

PR #2503, opened 2021-03-02
PIL-973 Feature/contacts gravatar identicon

mdjastrzebski:
Scope:
- [x] Replace contacts' initials with identicons
- [x] Update storbybook


wjkg (approved):
LGTM, one additional thing that could be done is cleaning up some leftover props of `ProfileImage` -- flow should give you a list if you change the `ProfileImageProps` to exact object type.

mdjastrzebski:
Removed leftover props.

PR #2504, opened 2021-03-03
persist activity feed header section item if no data

poocart:
Fixes issue on Android(?) when on fresh account Activity produces blank output on Home screen contents.

The fix is just restore of previous piece before performance PR (https://github.com/pillarwallet/pillarwallet/commit/455a8951dee796e51475eb623d962593b8f08af4) where sections had Header component even if feed data is empty.


mdjastrzebski (approved):
Well spotted! 🥇

PR #2505, opened 2021-03-03
PIL-968: Instabug integration

wjkg:


IAmKio:
Hi @wjkg, just to let you know we've made the decision just to merge this without the push  notifications for now and we'll add this as a later PR.

wjkg:
@IAmKio I've left the version with FCM in a separate PR in case it's still needed for debugging later: #2523

android/app/src/main/java/com/pillarproject/wallet/MainApplication.java
```diff
@@ -60,6 +61,11 @@ public void onCreate() {
         // react-native-async-storage custom max storage
         long storageSizeMax = 60L * 1024L * 1024L; // 60 MB
         com.reactnativecommunity.asyncstorage.ReactDatabaseSupplier.getInstance(getApplicationContext()).setMaximumSize(storageSizeMax);
+
+        new RNInstabugReactnativePackage
+            .Builder(getString(R.string.instabug_token), MainApplication.this)
+            .setInvocationEvent("shake")
```
mdjastrzebski, line 67: I wonder if we should add some option in the menu to manually trigger the instaback dialog. Shake gesture is kinda hidden and not evident.

App.js
```diff
@@ -35,6 +35,7 @@ import { NavigationActions } from 'react-navigation';
 import remoteConfig from '@react-native-firebase/remote-config';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 import { QueryClient, QueryClientProvider } from 'react-query';
+import Instabug, { NetworkLogger } from 'instabug-reactnative';
```
mdjastrzebski, line 38: Consider hidding Instabug dependency behind a thin wrapper module, which would have methods for:
1) Initialization
2) MakePrivateView
3) Callbacks/hooks (setColorTheme, onNavigationStateChange)

android/app/build.gradle
```diff
@@ -184,6 +184,8 @@ android {
         resValue "string", "branch_io_app_link_alternate", "abgsz-alternate.test-app.link"
         resValue "string", "branch_io_key_live", ""
         resValue "string", "branch_io_key_test", "key_test_fdVQmXxlMInIG98ByzdqlbdkCxi8QO3a"
+
+        resValue "string", "instabug_token", ""
```
mdjastrzebski, line 188: Do we plan to support instabug in debug builds?

android/app/src/main/java/com/pillarproject/wallet/MainApplication.java
mdjastrzebski, line 67: Consider adding menu option for invoking instabug in a discoverable way. Shake gesture seems rather had to find for regular users.

src/components/MnemonicPhrase/MnemonicPhrase.js
```diff
@@ -22,6 +22,7 @@ import styled from 'styled-components/native';
 import { BaseText, MediumText } from 'components/Typography';
 import { fontSizes } from 'utils/variables';
 import { getColorByTheme } from 'utils/themes';
+import { makeViewPrivate } from 'utils/ui';
```
mdjastrzebski, line 26:
```suggestion
import { excludeFromMonitoring } from 'utils/appMonitoring';
```

I struggle a bit about the name suggestion. However , the important thing would be somehow connect the `makeViewPrivate` method with monitoring feature. Which currently is not obvious when `makeViewPrivate` is imported from `utils/ui`.

mdjastrzebski (commented):
Overall looks good!

Some remarks worth discussing:
1. Consider hiding direct instabug dependency behind a thin module wrapper that would expose are relevant functions (init, markAsPrivateView, onChangeNavigationState, etc).
2. Such module could probably be named `monitoring` or `insights`, etc
3. Consider exposing a more obvious way for triggering user feedback, as current shake gesture seems hard to discover by users.

android/app/src/main/java/com/pillarproject/wallet/MainApplication.java
IAmKio: @mdjastrzebski just to let you know that for now, we just want to test out with the shake gesture. This is to prevent any clashes with the existing Intercom support system, however once we have successfully tested Instabug with a select few users via the shake mechanism, then we'll replace Intercom and make the invocation of this feature more prominent on the home screen.

IAmKio: @mdjastrzebski please see my comment above.

wjkg: In addition, with the default configuration, Instabug shows a message explaining how it can be accessed when launching the app for the first time: https://docs.instabug.com/docs/react-native-onboarding-experience so it's not completely hidden.

android/app/build.gradle
wjkg: On Android, it's the equivalent of setting Instabug token in `buildConfig.js`, because the initialisation differs between platforms.

wjkg:
@mdjastrzebski thank you for the feedback. Given that most calls are one-liners specific to Instabug, there wouldn't be much difference with or without an abstraction. You do make a very good suggestion regarding the view-masking function, however, and some other parts may in this case be moved together.

src/utils/monitoring.js
```diff
@@ -0,0 +1,51 @@
...
+
+export const excludeFromMonitoring = (element: ElementRef<any> | null) => {
+  if (element !== null) Instabug.setPrivateView(element);
+};
+
+export const setInstabugTheme = (appThemeName: string) => {
+  const instabugTheme = {
+    [LIGHT_THEME]: Instabug.colorTheme.light,
+    [DARK_THEME]: Instabug.colorTheme.dark,
+  }[appThemeName];
```
mdjastrzebski, line 46:
```suggestion
  const instabugTheme = themeType === DARK_THEME ?  Instabug.colorTheme.dark : Instabug.colorTheme.light);
```
nit: could be a bit simplified :-)

mdjastrzebski (approved):
Looks good!

PR #2514, opened 2021-03-10
Invite Contacts UX improvements: initial cleanup

mdjastrzebski:
Cleanup Invite Contacts screens before doing other changes. Changes are mostly structural, there should be no behaviour changes (except removed debounce).

Scope:
- [x] migrate `ReferralContacts`, `ReferFriends` to FC
- [x] remove unnecessary search input debounce in `ReferralContacts`

### DIFF
--- a/src/selectors/user.js
+++ b/src/selectors/user.js
@@ -25,6 +25,10 @@ import type { RootReducerState } from 'reducers/rootReducer';
 import type { User } from 'models/User';
 import type { Selector } from 'selectors';
 
+import { useRootSelector } from './selectors';
+
+export const useUser = () => useRootSelector((root) => root.user.data);
+
 export const userSelector = ({ user: { data } }: RootReducerState): User => data;
 
 export const usernameSelector: Selector<?string> =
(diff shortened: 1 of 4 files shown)


src/selectors/user.js
```diff
@@ -25,6 +25,10 @@ import type { RootReducerState } from 'reducers/rootReducer';
 import type { User } from 'models/User';
 import type { Selector } from 'selectors';
 
+import { useRootSelector } from './selectors';
+
+export const useUser = () => useRootSelector((root) => root.user.data);
+
 export const userSelector = ({ user: { data } }: RootReducerState): User => data;
```
wjkg, line 32: Do we need a custom hook where the selector below would suffice? When reading a component, it's clearer what to expect from `useRootSelector(userSelector)` than `useUser()`.

wjkg (approved):

src/selectors/user.js
mdjastrzebski: IMO it make sense for _most common_ use cases. I agree that somewhat duplicates `userSelector`, but on the other hand it's simplier to use in FCs than `userSelector`, which is better for Class Components.

PR #2518, opened 2021-03-12
Tweak some eslint rules to match @react-native-community/eslint-config

mdjastrzebski:
Scope:
- [x] Disable `no-use-before-define` because it enforces unnecessary string order in file obscuring readability
- [x] Disable `no-unused-expressions` because it warns on optional chaining calls `aaa?.()`
- [x] Update `.prettierrc` with rules matching current code style (this is NOT enforced, just for convenience)
- [x] Cleanup changed rules use cases


wjkg (approved):
👍 I would have suggested to disable these rules myself, If you hadn't made this PR first.

mdjastrzebski:
Yeah, these rules were pretty annoying and didn't bring any value.

PR #2520, opened 2021-03-12
PIL-976: Auto-retry Mechanism

wjkg:


src/utils/__tests__/retry.test.js
```diff
@@ -0,0 +1,113 @@
...
+    const value = retry(fn);
+
+    jest.runAllTimers();
+    await expect(value).resolves.toEqual(4);
+    expect(fn).toHaveBeenCalledTimes(4);
+  });
+
+  it('is rejected if none of the attempts succeed', async () => {
+    const fn = jest.fn(async (bail, attemptNo) => { throw new Error(`#${attemptNo}`); });
+    const value = retry(fn, { retries: 4 });
```
mdjastrzebski, line 50:
```suggestion
    const promise = retry(fn, { retries: 4 });
```

src/utils/retry.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// handles errors thrown by requests made using axios
+export function retryOnNetworkError<T>(fn: Retrier<T>, options?: Options): Promise<T> {
+  return retry<T>(async (bail, attemptNo): Promise<T> => {
+    try {
+      const result: T = await fn(bail, attemptNo);
+      return result;
+    } catch (error) {
+      if (!error.isAxiosError || error.response) {
```
mdjastrzebski, line 51: Consider introducing boolean var for readability
```
const isNetworkError = error.isAxiosError && !error.response;
```

src/utils/retry.js
```diff
@@ -0,0 +1,63 @@
...
+  return asyncRetry(fn, {
+    retries: 10,
+    factor: 1,
+    minTimeout: 6000,
+    randomize: false,
+    ...options,
+  });
+}
+
+// handles errors thrown by requests made using axios
```
mdjastrzebski, line 44: Please add code comment describing the behavior of retrying on network error but not on server response error, etc

src/utils/retry.js
```diff
@@ -0,0 +1,63 @@
+// @flow
+/*
+    Pillar Wallet: the personal data locker
+    Copyright (C) 2019 Stiftung Pillar Project
```
mdjastrzebski, line 4: 2021 ;-)

src/utils/retry.js
```diff
@@ -0,0 +1,63 @@
...
+  factor?: number,
+  minTimeout?: number,
+  maxTimeout?: number,
+  randomize?: boolean,
+  onRetry?: (error: Error) => void,
+};
+
+function retry<T>(fn: Retrier<T>, options: Options = {}): Promise<T> {
+  return asyncRetry(fn, {
+    retries: 10,
```
mdjastrzebski, line 36: Ten retries seems excessive. Unless you have some specific technical requireent I would rather consider around 3 retries.

src/services/uniswap.js
```diff
@@ -311,18 +312,18 @@ export const createUniswapAllowanceTx =
     return allowanceTx;
   };
 
-export const fetchUniswapSupportedTokens = (): Promise<?string[]> => axios
-  .get(getEnv().UNISWAP_CACHED_SUBGRAPH_ASSETS_URL, defaultAxiosRequestConfig)
-  .then(({ data: responseData }: AxiosResponse) => {
-    if (!responseData) {
-      reportErrorLog('fetchUniswapSupportedTokens failed: unexpected response', { response: responseData });
+export const fetchUniswapSupportedTokens = (): Promise<?string[]> =>
+  retryOnNetworkError(() => axios.get(getEnv().UNISWAP_CACHED_SUBGRAPH_ASSETS_URL, defaultAxiosRequestConfig))
```
mdjastrzebski, line 316: Instead of wrapping every axios call, consider providing an thin axios abstraction, that would do that in a centralized form.
```
httpClient.get(...)

// Somewhere in utils
const httpClient: {
  get: (...args) => { retryOnNetworkError(() => axios.get(...args)) }
  post: etc
}
```

mdjastrzebski (changes requested):
Nice feature Weronika! 👍🏻 

Couple of points I feel are worth discussing:
1. Please provide clear info on retryOnNetworkError behaviour, as I learned that it does not retry on server errors only after reading the tests. Code comment should be enough.
2. Instead of wrapping all axios calls with repeatOnNetworkError consider providing a thin axios wrapper that would apply retry on a centralised manner. This should be useful if we were to apply any other centralized axios changes in the future, as well as it would encapsulate axios dep.
3. Consider reducing retry number to say 3, current value of 10 seems too much.

src/utils/retry.js
wjkg: It's the value specified in the ticket, but I agree it could be lower. @IAmKio is there a reason it was set to 10?

src/services/uniswap.js
wjkg: I _would_ like something more elegant than a wrapper, but replacing it with an abstraction is not quite straightforward. It's supposed to be opt-in per API call, even if admittedly all GET requests I've found were ok to be wrapped. The same definitely isn't true for POST ones. A replacement also wouldn't work for SDK requests. Given that, we'd end up having some calls wrapped and some abstracted with nothing to suggest the non-standard feature. Add to that that using axios means handling axios-specific behaviour and response format and I'd rather have some light repetition if it means there's also consistency and clearer view of what exactly is being called.

I'll have think about it some more, because the opt-in part would be solved by adding a custom option to the requests, hiding axios by typing, so that leaves SDK calls. I'd most like to avoid retries being both mentioned by name and hidden in requests next to each other. Any thoughts?

src/utils/retry.js
wjkg: You're right, it should be less ambiguous. I meant to indicate what kind of promise it expects from `fn`.

src/services/uniswap.js
mdjastrzebski: 1. You're right that this probably should apply to GET calls but not POST ones, as repeated calls there could lead to unintended consequences.
2. For sure SDK requests would need to be wrapped in `retryOnNetworkErrorDirectly` but that makes only sense if they rethrow original axios error(!), any other error seems not to be retried anyway.
3. Maybe we could make retry behavior default for all GET calls, with option to opt-out. Sounds like a reasonable default. To achieve that we could add option flag to options param. Achieving non-retry behavior should probably only require specifying zero retry count without the need not-to-wrap GET call.

src/utils/retry.js
IAmKio: Nothing other than 10 retries with 6 second minimum timeout = 1 minute. I'm happy for this to change, I myself was debating this one.

src/actions/exchangeActions.js
```diff
@@ -54,6 +53,7 @@ import { getSmartWalletAddress } from 'utils/accounts';
 import { reportErrorLog, reportLog } from 'utils/common';
 import { getAssetsAsList, getAssetData, isSynthetixTx } from 'utils/assets';
 import { isOrderAmountTooLow } from 'utils/exchange';
+import httpRequest from 'utils/httpRequest';
```
mdjastrzebski, line 56: Consider changing naming to more standard one.
```suggestion
import httpClient from 'utils/httpClient';
```

mdjastrzebski (approved):
Looks great! 🥇 

One minor suggestion would be to rename `httpRequest` to `httpClient` as this seem to be more popular name for such module.

PR #2521, opened 2021-03-12
Chore: ESLint apply hooks plugin

mdjastrzebski:
Scope:
- [x] Apply recommended config for react hooks
- [x] Disable all existing violation, so that there are no behavior changes


src/components/CircularProgressBar/CircularProgressBar.js
```diff
@@ -57,6 +57,7 @@ const CircularProgressBar = ({
 
   useEffect(() => {
     Animated.timing(circleAnim, { toValue: progress, duration: 500, useNativeDriver: true }).start();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jdnmr, line 60: If we don't like `react-hooks/exhaustive-deps` that much, can't we disable it globally?

jdnmr (approved):

src/components/CircularProgressBar/CircularProgressBar.js
mdjastrzebski: It's a recommended hook, which this PR enables. However we disable are existing rule violations, so that it will only apply to new code.

PR #2524, opened 2021-03-16
silent etherspot accounts implementation

poocart:
This add silent implementation of Etherspot SDK. Etherspot Smart Wallet accounts will be created in the background for both new and current users.


src/services/etherspot.js
```diff
@@ -0,0 +1,75 @@
...
+      : NetworkNames.Kovan;
+
+    const envName = getEnv().NETWORK_PROVIDER === 'homestead'
+      ? EnvNames.MainNets
+      : EnvNames.TestNets;
+
+    this.sdk = new EtherspotSdk(privateKey, { env: envName, networkName });
+
+    await this.sdk.computeContractAccount({ sync: true }).catch((error) => {
+      reportErrorLog('EtherspotService init computeContractAccount failed', { error });
```
IAmKio, line 52: Should we return something here instead of `undefined`?

src/services/etherspot.js
```diff
@@ -0,0 +1,75 @@
...
+    return this.sdk.getConnectedAccounts()
+      .then(({ items }: EtherspotAccounts) => items)
+      .catch((error) => {
+        reportErrorLog('EtherspotService getAccounts -> getConnectedAccounts failed', { error });
+        return null;
+      });
+  }
+
+  async logout() {
+    if (!this.sdk) return; // not initialized, nothing to do
```
IAmKio, line 66: Just to let you know i'm not suggesting anything here because of the nature of the function 😂 orrr do you think `logout` should return _something_? Like true / false? Is it any help to us?

src/utils/accounts.js
```diff
@@ -85,6 +85,10 @@ export const findFirstSmartAccount = (accounts: Accounts): ?Account => {
   return accounts.find(({ type }) => type === ACCOUNT_TYPES.SMART_WALLET);
 };
 
+export const findFirstEtherspotAccount = (accounts: Accounts): ?Account => {
+  return accounts.find(({ type }) => type === ACCOUNT_TYPES.ETHERSPOT_SMART_WALLET);
+};
```
IAmKio, line 90: Could be shorthanded to:

```suggestion
export const findFirstEtherspotAccount = (accounts: Accounts): ?Account => return accounts.find(({ type }) => type === ACCOUNT_TYPES.ETHERSPOT_SMART_WALLET);
```

Maybe? Just a suggestion.

IAmKio (approved):
Just a few comments but none of them are show-stoppers, great stuff and thanks! 🎉

src/services/etherspot.js
poocart: Not sure, at this moment I don't think we care about the output 🤔  Added `void` return.

src/utils/accounts.js
poocart: Thanks! Added more improved solution 👍

package.json
```diff
@@ -171,9 +172,11 @@
     "redux-devtools-extension": "^2.13.2",
     "redux-persist": "^6.0.0",
     "redux-thunk": "^2.2.0",
+    "reflect-metadata": "^0.1.13",
     "reselect": "^4.0.0",
     "rn-fetch-blob": "^0.12.0",
     "rn-nodeify": "^10.0.1",
+    "rxjs": "^6.6.2",
```
mdjastrzebski, line 179: Is this dep actually used?

src/actions/etherspotActions.js
```diff
@@ -0,0 +1,106 @@
...
+      return;
+    }
+
+    if (!user) {
+      reportErrorLog('importEtherspotAccountsAction failed: no user');
+      return;
+    }
+
+    const etherspotAccounts = await etherspot.getAccounts();
+    if (!etherspotAccounts) {
```
mdjastrzebski, line 73: Is it possible for get accounts to return empty array? If yes:
```suggestion
    if (!etherspotAccounts?.length) {
```

src/actions/etherspotActions.js
```diff
@@ -0,0 +1,106 @@
...
+      ACCOUNT_TYPES.ETHERSPOT_SMART_WALLET,
+      etherspotAccount, // full object as extras
+    ))));
+
+    const { walletId } = user;
+
+    if (!walletId) {
+      reportErrorLog('importEtherspotAccountsAction failed: no walletId', { user });
+      return;
+    }
```
mdjastrzebski, line 91: As `user` is take from state at the begging of the action method, and redux state should be immutable, then either:
1) This check could be moved just after `if (!user)` check
2) `user` should be re-queried from state, so that it contains updated `walletId`

src/services/etherspot.js
```diff
@@ -0,0 +1,75 @@
...
+// config
+import { getEnv } from 'configs/envConfig';
+
+
+class EtherspotService {
+  sdk: EtherspotSdk;
+  subscription: any;
+
+  async init(privateKey: string) {
+    const networkName = getEnv().NETWORK_PROVIDER === 'homestead'
```
mdjastrzebski, line 41: nit: Consider using `isProdEnv` from `utils/environment`

src/services/etherspot.js
mdjastrzebski: Atm we return implicitly return `Promise<void>` which seems fine.

mdjastrzebski (approved):
Overall looks good.

Added some minor comments.

BTW is it possible to have more than one Etherspot account?

src/actions/etherspotActions.js
poocart: No, it cannot produce empty array unless contract address is not computed, but that is done during initialization.

src/services/etherspot.js
```diff
@@ -0,0 +1,75 @@
...
+    });
+  }
+
+  getAccounts(): Promise<?EtherspotAccount[]> {
+    return this.sdk.getConnectedAccounts()
+      .then(({ items }: EtherspotAccounts) => items)
+      .catch((error) => {
+        reportErrorLog('EtherspotService getAccounts -> getConnectedAccounts failed', { error });
+        return null;
+      });
```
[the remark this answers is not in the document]
poocart: Is there any particular reason why you want it to be wrapped with `try...catch` exception handler? If it's because of SDK instance being empty and therefore failing on selected method call then we can just add a check before, wdyt?

src/services/etherspot.js
```diff
@@ -0,0 +1,75 @@
...
+// utils
+import { reportErrorLog } from 'utils/common';
+
+// config
+import { getEnv } from 'configs/envConfig';
+
+
+class EtherspotService {
+  sdk: EtherspotSdk;
+  subscription: any;
```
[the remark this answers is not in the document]
poocart: Thanks, I copied parts of code from full implementation, removed 👍

package.json
poocart: It's required on Etherspot SDK install https://github.com/etherspot/etherspot-sdk#installation, however, it was used in full implementation branch, but just tried to remove it here and it works just fine without – included changes, thanks 👍  P. S. same goes for `reflect-metadata`, however removing it causes missing dependency issues

src/services/etherspot.js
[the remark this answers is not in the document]
mdjastrzebski: This should be exactly the same code as replaced one, just coded as `async/await` instead of `.then().catch()`. This is a bit of nitpicking form my side :-) but I thought that as you used `async/await` in other methods from the same file, it would make it more consistent.

src/services/etherspot.js
poocart: Thanks, that util as it was introduced later, replaced 👍

src/services/etherspot.js
IAmKio: "I don't think we care about the output" - good enough for me!

src/actions/etherspotActions.js
poocart: That's fine, we care about the state of user when the action was fired, however, moved the check upper 👍

src/services/etherspot.js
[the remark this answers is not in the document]
poocart: This is single method that has been copied over, but there are many more in the full implementation branch.

Actually the reason of choosing `async/await` over `try/catch` at this piece is just that it's easier to destruct the response object with less code.

I think let's keep it as it is if that won't bother you too much 😂

PR #2525, opened 2021-03-17
Differentiate instabug tokens between apps

antonioned:
Both tokens are added in our pipeline.


IAmKio (approved):
Thank you for the quick turnaround 🎉 💪

vnovacev (approved):

PR #2527, opened 2021-03-17
PIL-962: Feature/send flow ux improvements

mdjastrzebski:
Scope:
- [x] Tweak sizing of various elements
- [x] Display full address in ContactSelector


wjkg (commented):
I've noticed something after pressing the "send max" link:

image
(this happens when the text field is not focused)

and the estimate disappears until I remove some of the digits from the end. The estimate also seems to cause an error to show up in a toast if one enters amount larger than available funds, though it may not be in the scope of this PR.

mdjastrzebski:
Applied workaround for issue with value input amount breaking into two lines, by removing custom height which caused the issues with reduced font size. Theoretically that text should not break as it's single line TextInput underneath, but the `ValueInput -> TextInput` component structure is so convoluted that it's hard to diagnose the root cause.

Actually fixing root cause that seems like a major undertaking, potentially requiring duplicating/rewriting `TextInput`.

mdjastrzebski:
Re estimate failing on too much value: this seems non-trivial to fix and should be handled separetely. Here is the task for that: https://linear.app/pillarproject/issue/PIL-984/send-asset-estimate-failed-on-too-large-value

wjkg:
What about the disappearance of estimate on pressing "send max"? Is it the same situation as with failing on invalid value?

mdjastrzebski:
This should be fixed now

wjkg (approved):

PR #2528, opened 2021-03-17
PIL-962: Feature/send flow ux asset selector

mdjastrzebski:
Scope:
- [x] SelectorOptions remove horizontal options and other not used props
- [x] AssetSelectorOptions:
  - [x] duplicated from SelectorOptions and focusing only on asset selection
  - [x] updated sort order
  - [x] tweak list item display
- [x] Improve typing: AssetOption, AssetBalance, component props typing, etc

### DIFF
--- a/src/components/ListItem/ListItemWithImage.js
+++ b/src/components/ListItem/ListItemWithImage.js
@@ -40,8 +40,8 @@ import { fontSizes, spacing, fontTrackings, fontStyles, objectFontStyles } from 
 import { getColorByTheme, getThemeColors } from 'utils/themes';
 import { images } from 'utils/images';
 
+import type { AssetBalance } from 'models/Asset';
 import type { Theme, ThemeColors } from 'models/Theme';
-import type { AssetBalance } from 'models/Selector';
 
 
 type Props = {
--- a/src/components/ListItem/ListItemWithImage.js
+++ b/src/components/ListItem/ListItemWithImage.js
@@ -254,13 +254,13 @@ const ItemValue = styled(BaseText)`
   text-align: right;
 `;
 
-const BalanceValue = styled(BaseText)`
+const BalanceFiatValue = styled(BaseText)`
   ${fontStyles.big};
   color: ${({ color, theme }) => color || theme.colors.basic010};
   text-align: right;
 `;
 
-const BalanceFiatValue = styled(BaseText)`
+const BalanceValue = styled(BaseText)`
   ${fontStyles.regular};
   color: ${({ theme }) => theme.colors.basic030};
   text-align: right;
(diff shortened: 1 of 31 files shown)


src/utils/array.js
```diff
@@ -0,0 +1,27 @@
...
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+/**
+ * Performs map operation and filters out `null` and `undefined` results.
+ */
+export function mapNotNil<Item, Result>(items: Item[], selector: (Item) => ?Result): Result[] {
+  // $FlowFixMe: does not infer that `filter` removes nil values
+  return items.map((item) => selector(item)).filter((item) => item != null);
```
wjkg, line 26: `selector` type implies that the function also filters out `undefined` -- which it does, but because it uses `!=` instead of `!==`. I'd rather not rely on that kind of detail about the language, though.

Also, since this is a wrapper for `map`, it could support all arguments that are usually passed to selector.

In general, good idea, since I've seen it repeated a few times.

src/components/ValueInput/ValueInput.js
```diff
@@ -50,23 +50,24 @@ import { visibleActiveAccountAssetsWithBalanceSelector } from 'selectors/assets'
...
-  customAssets?: Option[],
+  customAssets?: AssetOption[],
   customBalances?: Balances,
   selectorOptionsTitle?: string,
-  assetData: Option,
-  onAssetDataChange: (Option) => void,
+  assetData: AssetOption | Collectible,
+  onAssetDataChange: (AssetOption) => mixed,
+  // Called iff selected asset is Collectible
+  onCollectibleAssetDataChange?: (Collectible) => mixed,
```
wjkg, line 68: Looking at the props, it's not entirely clear if on collectible selection both methods are called, or only the second one, since collectibles here also count as assets.

wjkg (commented):
Maybe inputs for token and collectible assets could be split into separate components? Visually they're very different, but you'll have a better idea if it makes sense, since it would take me some time to check how much code is actually shared between them.

Otherwise, lgtm, just a couple of stylistic suggestions.

src/components/ValueInput/ValueInput.js
mdjastrzebski: It is not called. Do you think a comment will be enough?

I struggle here with naming, as `Asset` type (and `AssetOption`) seem to hold only non-collectible assets. Is there any ubiquitous name used in our app for such assets?

src/utils/array.js
mdjastrzebski: Re naming: I used `NotNil` because popular libs like lodash, ramda, etc use `nil` to signify `null` or `undefined` (https://lodash.com/docs/4.17.15#isNil)

mdjastrzebski: Re `!= null`: I consider this to be standard JS idiom meaning `!== null & !== undefined` on par with `!` and `!!` (falsy handling).

mdjastrzebski: Re selector args: I'll add these to match `map`

src/components/ValueInput/ValueInput.js
wjkg: I'm not sure how to solve the naming -- maybe using "token asset" locally, like the type constants? The tabs on the assets screen are also "Tokens" and "Collectibles".

IMO it would be simpler to merge the change events into one and leave checking asset type to the handler, but if that is not possible then a comment that makes it clear which callbacks are called in both cases would be welcome.

mdjastrzebski: `Token` is also problematic as we have `tokenType` field in `Collectible` :-/

mdjastrzebski: I wanted to have separate events because they return two different types. While I wanted to avoid returning `AssetOption | Collectible` from a single method because majority use-cases do not support `Collectibles` and I didn't want to confuse the majority use case.

mdjastrzebski: I'll simply add comment then :-)

src/components/ListItem/ListItemWithImage.js
```diff
@@ -552,28 +552,20 @@ const Addon = (props: AddonProps) => {
       balance: tokenBalance = '',
       token = '',
       value = '',
-      custom,
-      customOnRight,
     } = balance;
-    const decimalPlaces = getDecimalPlaces(token);
-    const roundedBalance = formatAmount(tokenBalance, decimalPlaces);
```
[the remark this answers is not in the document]
mdjastrzebski: It is formatted in all call-sites using `formatTokenAmount`, and applying formatAmount caused issues.

src/utils/array.js
wjkg: No, nil is fine, though now that you pointed it out the whole name sounds a bit like filter is applied first. Re `!=` I must not have seen it used enough to see it as a standard 🙂 but it does look useful as long as one knows what one's doing.

mdjastrzebski:
Re splitting asset & collectible selection: this would be a good idea for the future, however it will be a major undertaking in terms of effort. Additionally this is currently bound by UX which puts these two together.

wjkg (approved):

PR #2539, opened 2021-03-24
Fix/send asset precision issues

mdjastrzebski:
Scope:
- [x] Fix issue when too high precision was passed to tx or tx estimation causing ethers.js "underflow" error
- [x] Improve `ValueInput` rounding when using % selector

### DIFF
--- a/src/components/SendAsset/SendAsset.js
+++ b/src/components/SendAsset/SendAsset.js
@@ -39,14 +39,13 @@ import SendContainer from 'containers/SendContainer';
 import Toast from 'components/Toast';
 
 // utils
-import { wrapBigNumber, reportErrorLog } from 'utils/common';
+import { wrapBigNumber, truncateAmount, reportErrorLog } from 'utils/common';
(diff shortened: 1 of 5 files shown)


src/components/SendAsset/SendAsset.js
```diff
@@ -168,28 +165,9 @@ const SendAsset = ({
...
-    if (assetData.tokenType === COLLECTIBLES) {
-      formattedSelectedAsset = collectibles.find(({ tokenId }) => assetData.id === tokenId);
-    } else {
-      // $FlowFixMe: collectible
-      formattedSelectedAsset = assetsWithBalance.find((asset) => assetData.token === asset.token);
-    }
-
-    if (!formattedSelectedAsset) return;
-
-    handleAmountChange({ selector: formattedSelectedAsset, input: '' });
```
mdjastrzebski, line 192: All of this code seems to actually do nothing because it's run once during component mount.

wjkg (approved):

PR #2546, opened 2021-03-29
PIL-980: Feature: Storyly

wjkg:
The Storyly component doesn't allow much customisation, so the design was changed to display stories in a row above the total balance.


src/constants/remoteConfigConstants.js
```diff
@@ -51,6 +52,7 @@ export const INITIAL_REMOTE_CONFIG = {
   [REMOTE_CONFIG.FEATURE_AAVE]: true,
   [REMOTE_CONFIG.FEATURE_POOL_TOGETHER]: true,
   [REMOTE_CONFIG.FEATURE_ALTALIX]: true,
+  [REMOTE_CONFIG.FEATURE_STORIES]: true,
```
IAmKio, line 55: After a chat with the people making the content, could we please turn this off for now? Thanks!

IAmKio (commented):
Looks good to me, just one small thing re: the flag default value. We'll switch this back on permanently later but for now, it needs to be `false`.

IAmKio (approved):
🎉

src/components/Stories/Stories.js
```diff
@@ -0,0 +1,59 @@
...
+
+  const logStoryOpen = () => dispatch(logEventAction('STORY_OPEN'));
+
+  return (
+    <StorylyWthSpacing
+      storylyId={getEnv().STORYLY_TOKEN}
+      onLoad={setCountOnLoad}
+      onFail={logStorylyError}
+      onStoryOpen={logStoryOpen}
+      $hide={storyGroupCount === 0}
```
mdjastrzebski, line 48: Would it make sense to use conditional rendering here instead of setting height to zero?

mdjastrzebski, line 48: I like the idea of $props for styled components stuff :-)

mdjastrzebski (approved):
Overall looks goods.

Some minor comments in code.

Please let Dmitry know about layout limitations, so he can update his design.

src/components/Stories/Stories.js
wjkg: It's not an idea, exactly, but a feature of `styled-components` that prevents the prop from being passed to the underlying component :)

src/components/Stories/Stories.js
```diff
@@ -0,0 +1,59 @@
...
+    />
+  );
+};
+
+const StorylyWthSpacing = styled(Storyly)`
+  height: ${({ $hide }) => ($hide ? 0 : 140)}px;
+  overflow: hidden;
+  margin-left: ${spacing.layoutSides}px;
```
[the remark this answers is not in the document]
wjkg: I'm not quite sure what do you mean by that. If it's about moving the style to the outer component, I'd prefer to keep stories-related spacing defined in one place.

.circleci/config.yml
```diff
@@ -63,19 +63,6 @@ aliases:
       name: Restore pods cache
       keys:
         - ios-pods-{{ checksum "./ios/Podfile.lock" }}
-  - &fetch_pods
```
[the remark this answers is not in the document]
wjkg: Yes, it only cleans up unused code that was noticed when upgrading CocoaPods.

src/components/Stories/Stories.js
wjkg: We want to show it only after it's known there are stories to display, but to make the initial request it needs to be rendered first. `display: none` is going to be clearer, though.

PR #2547, opened 2021-03-29
Feature/migrate to smart wallet summary

mdjastrzebski:
Scope:
- [x] Migration Review screen
- [X] Re-usable `Table` & `TableFee` components


src/components/modern/FeeTable.js
```diff
@@ -0,0 +1,98 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import * as React from 'react';
```
IAmKio, line 21: Is importing `*` necessary? Could we just import what we need?

src/components/modern/FeeTable.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export default FeeTable;
+
+type FeeItemProps = {|
+  title: string,
+  symbol: string,
+  fee: Value,
+  separator?: boolean,
+|};
```
IAmKio, line 69: Just wondering if we should move all type definitions together at the top fo the file?

src/components/modern/Table.js
```diff
@@ -0,0 +1,96 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import * as React from 'react';
```
IAmKio, line 21: Same as above, do we need to do `*`?

src/components/modern/Table.js
```diff
@@ -0,0 +1,96 @@
...
+  variant?: RowValueVariant,
+  fontVariant?: FontVariant,
+|};
+
+export const RowValue: React.ComponentType<RowValueProps> = styled(Text)`
+  margin-left: ${spacing.extraSmall}px;
+  ${({ fontVariant }) => (fontVariant ? `font-variant: ${fontVariant}` : '')};
+  ${({ variant, theme }) => (variant === 'secondary' ? `color: ${theme.colors.secondaryText}` : '')};
+  ${({ variant, theme }) => (variant === 'positive' ? `color: ${theme.colors.positive}` : '')};
+`;
```
IAmKio, line 95: I won't mention this again, but it seems logical to me that we should move all type definitions to the top of the file and leave exports to the bottom of the file. I can see how you're writing this; define - export, and repeat - but to me that _feels_ like this needs to be two UI components when it's written this way - even though i understand that it's all one <table> HTML element render, hence why you've grouped it all together.

Let me know what you think - it's just a suggestion 👍

src/locales/en/common.json
```diff
@@ -122,6 +122,7 @@
     "needToActivate": "Need to activate",
     "newSmartWalletAccountDevice": "New Smart Wallet account device",
     "notApplicable": "N/A",
+    "notEnoughGas": "Insufficient gas",
```
IAmKio, line 125:
```suggestion
    "notEnoughGas": "Sorry, insufficient gas was provided. Please try again or contact our support team for help",
```

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferChoose.js
```diff
@@ -204,15 +204,15 @@ const KeyBasedAssetTransferChoose = ({
...
-              calculateTransactionsGas(); // start calculating
-              navigation.navigate(KEY_BASED_ASSET_TRANSFER_CONFIRM);
-            }}
-          />
-        </Footer>
+          <FooterContent>
+            <Button
+              title={t('button.next')}
+              onPress={() => {
+                calculateTransactionsGas(); // start calculating
```
IAmKio, line 211: When you say "start calculating", i'm assuming this is actually an "instant-ish" calculation (data payload * pre-fetched gas price from the SDK) or does it require some calculation? If the latter is true, should we not `await` this and then call navigate? 

I'm making some assumptions here about `calculateTransactionsGas()` 😂 but please feel free to clarify!

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferConfirm.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 /*
     Pillar Wallet: the personal data locker
-    Copyright (C) 2019 Stiftung Pillar Project
+    Copyright (C) 2021 Stiftung Pillar Project
```
IAmKio, line 4: Cheers! 😂

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferConfirm.js
```diff
@@ -17,192 +17,227 @@
...
-const FooterInner = styled.View`
-  align-items: center;
-  width: 100%;
-`;
+  const images = useThemedImages();
 
-const NotEnoughFee = styled(BaseText)`
-  margin-top: ${spacing.large}px;
-`;
+  if (isCalculatingGas) {
```
IAmKio, line 70: I think this answered my previous question about gas calculation...

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferConfirm.js
```diff
@@ -17,192 +17,227 @@
...
-
-  const tokensTransfer = keyBasedAssetsToTransfer.filter(
-    ({ assetData }) => assetData?.tokenType !== COLLECTIBLES,
-  );
-  const collectiblesTransfer = keyBasedAssetsToTransfer.filter(
-    ({ assetData }) => assetData?.tokenType === COLLECTIBLES,
-  );
+  const handleSubmit = () => {
+    navigation.navigate(KEY_BASED_ASSET_TRANSFER_UNLOCK);
+  };
```
IAmKio, line 82:
```suggestion
  const handleSubmit = () => navigation.navigate(KEY_BASED_ASSET_TRANSFER_UNLOCK);
```

IAmKio (changes requested):
Just a few minor comments - but overall, some great optimisations and fantastic improvements! Just some questions and some minor requests, but great work all round! 🎉 🎉 🎉

src/components/modern/FeeTable.js
mdjastrzebski: This syntax has become rather standard in modern React dev. Please check the reasoning behind that: https://epicreact.dev/importing-react-through-the-ages/ from KCD

src/components/modern/FeeTable.js
mdjastrzebski: This intentional to colocate prop type def next to component that uses these props.

src/components/modern/Table.js
mdjastrzebski: Personally, I find it easier to read if connected parts are co-located close together instead of sorting them by type, which will require scrolling back and forth.

src/locales/en/common.json
mdjastrzebski: This is the button text as on this screen:

image

Using longer text seems to result in odd design for a button. The design comes from Dmitry, with the reasoning for reduced text, that we assume some level of user knowledge here.

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferConfirm.js
mdjastrzebski: Time flies ;-)

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferConfirm.js
mdjastrzebski: This seem to run the transaction in the simulation mode to get the gas price though Infura/Provider, hence await + UI feedback. BTW the transfer logic did not change, and previous version of the screen also used spinners.

src/components/modern/FeeTable.js
mdjastrzebski: Basically this is to reduce code complexity to anybody reading the code, following a rule of thumb that "Things that change together should be located as close as reasonable." Pls check the following reasoning in more details here: https://kentcdodds.com/blog/colocation

src/components/modern/Table.js
mdjastrzebski: This is similar idea as for FeeTable comment.

src/components/modern/FeeTable.js
IAmKio: Thanks for that, great read!

src/components/modern/FeeTable.js
IAmKio: "Things that change together should be located as close as reasonable" - interesting, that logically makes sense but it goes against my own ethos of separating concerns... at least it feels that way, anyway!

src/components/modern/Table.js
IAmKio: Saw it - cheers!

src/locales/en/common.json
IAmKio: Apologies - i thought this was a toast or something similar.

IAmKio (approved):

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferChoose.js
mdjastrzebski: This is redux action call, which is returns instantianously but there result will be reflected in redux state after a while.

PR #2548, opened 2021-03-29
PIL-983: Feature: Tutorial

jdnmr:
Display tutorial flow to users:
- Each user (current or new) will go through the flow once; they're navigated right after login/onboarding;
- The flow is CMS driven, except of the first screen. Data can be changed, screens can be added and removed entirely on the CMS side - no changes in code should be required, as long as data is valid and follows the agreed pattern;
- The feature requires the.`_prismic_token_` build variable which is not set as for now. For testing purposes, the value can be shared by @IAmKio and hardcoded;
- Components/functions dealing with the feature were renamed to `Tutorial...` to distinguish the feature from the onboarding flow;
- To disable the feature, simply set the prismic data to null/empty array - tutorial won't be displayed.

Note: the hardcoded strings for the initial screen are dummy data as we're still waiting for actual ones to be provided.


src/screens/Tutorial/Tutorial.js
```diff
@@ -0,0 +1,84 @@
...
+} = CMS_DATA_TYPES;
+
+type TUTORIAL_PATH = typeof NATIVES | typeof NEWBIES;
+
+const TutorialScreen = ({ navigation, hasSeenTutorial, tutorialData }: Props) => {
+  const [activePath, setActivePath] = useState<TUTORIAL_PATH>(NEWBIES);
+  const routeName = navigation?.state?.params?.nextNavigationRouteName || HOME;
+
+  if (!tutorialData) {
+    navigation.navigate(routeName);
```
mdjastrzebski, line 54: Here you invoke navigation side-effect in render, this should probably be done in `useLayoutEffect` (or `useEffect`)

src/screens/Tutorial/TutorialCMSView.js
```diff
@@ -0,0 +1,87 @@
...
+        source={{ uri: imageUrl }}
+        style={getImageStyle()}
+        resizeMode="contain"
+      />
+      <Body>{body}</Body>
+    </ScrollView>
+  );
+};
+
+export default withTheme(CMSView);
```
mdjastrzebski, line 87: Pls use `useTheme` hook instead

src/screens/Tutorial/TutorialCMSView.js
```diff
@@ -0,0 +1,87 @@
...
+  const getImageStyle = () => {
+    const width = DEVICE_WIDTH - (2 * spacing.rhythm);
+    const ratio = imageHeight / imageWidth;
+    const height = width * ratio;
+    return { width, height };
+  };
+
+  return (
+    <ScrollView
+      style={{ flex: 1, width: DEVICE_WIDTH }}
```
mdjastrzebski, line 70:
```suggestion
      style={{ flex: 1, width: DEVICE_WIDTH }}
```
Do you really need to explicitly set width as DEVICE_WIDTH. `100%` or `align-self: stretch` would be more future proof.

src/models/CMSData.js
```diff
@@ -0,0 +1,61 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+export type CMSDocument = {
```
mdjastrzebski, line 21: style, nit: could we use CmsDocument and similar naming, in order to match `Url`, `Ens`, `Http`, etc casing?

src/constants/cmsConstants.js
```diff
@@ -0,0 +1,27 @@
...
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+export const CMS_DATA_TYPES = {
+  ONBOARDING_SCREENS_FOR_NATIVES: 'onboarding_screens_for_natives',
+  DEFI_SHOWCASE: 'defi_showcase',
+  ONBOARDING_SCREENS_FOR_NEWBIES: 'onboarding_screens_for_newbies',
+};
```
mdjastrzebski, line 25: Explicit typing should help to get `enum`-like typesafety:
```
ONBOARDING_SCREENS_FOR_NATIVES: ('onboarding_screens_for_natives': 'onboarding_screens_for_natives'),
...
```

src/utils/cms.js
```diff
@@ -0,0 +1,63 @@
...
+import orderBy from 'lodash.orderby';
+import type { CMSDocument, ParsedCMSDocument, TutorialDataObject, CMSData } from 'models/CMSData';
+import { CMS_DATA_TYPES } from 'constants/cmsConstants';
+
+const {
+  ONBOARDING_SCREENS_FOR_NATIVES: NATIVES,
+  ONBOARDING_SCREENS_FOR_NEWBIES: NEWBIES,
+} = CMS_DATA_TYPES;
+
+const getSortedTutorialData = (docs: CMSDocument[], type: string): ParsedCMSDocument[] =>
```
mdjastrzebski, line 30: Naming is a bit misleading, as beside sorting it does parsing and filtering:

```suggestion
const getTutorialDataForType = (docs: CMSDocument[], type: string): ParsedCMSDocument[] =>
```

src/utils/cms.js
```diff
@@ -0,0 +1,63 @@
...
+  orderBy(docs.map(d => parseCMSDocument(d)).filter(d => d.type === type), i => (i.order || 0), ['asc']);
+
+const parseCMSDocument = (doc: CMSDocument): ParsedCMSDocument => {
+  const {
+    id, type, href, slugs, data: {
+      order = 0, title, subtitle, body, image,
+    } = {},
+  } = doc;
```
mdjastrzebski, line 38:
```suggestion
  const {
    id, type, href, slugs, data: {
      order = 0, title, subtitle, body, image,
    } = {},
  } = doc;
```
style: nested destructuring is quite hard to comprehend, could we split it
```suggestion
  const { id, type, href, slugs, data = {} } = doc;
  const { order = 0, title, subtitle, body, image } = data;
```

src/actions/cmsActions.js
```diff
@@ -0,0 +1,49 @@
...
+  return prismicClient.query(Predicates.any(DOCUMENT_TYPE, [NATIVES, NEWBIES]))
+    .then((res: CMSData) => {
+      const tutorialData = getTutorialDataObject(res);
+      if (!isValidTutorialData(tutorialData)) return;
+      dispatch({
+        type: SET_TUTORIAL_DATA,
+        payload: tutorialData,
+      });
+    })
+    .catch(e => reportErrorLog(e));
```
mdjastrzebski, line 48: style, readability: Could we use `await` + `try/catch` instead of `then()` + `catch()`?

src/screens/Tutorial/TutorialSwiper.js
```diff
@@ -0,0 +1,144 @@
...
+        <TutorialButton
+          title={t('button.native')}
+          onPress={() => handleButtonPress(ONBOARDING_SCREENS_FOR_NATIVES)}
+        />
+      </ContentWrapper>
+    </InitialScreenWrapper>
+  );
+
+  return (
+    <ContainerWithHeader headerProps={{ noBack: true, centerItems: [{ title: t('title.welcome') }] }}>
```
mdjastrzebski, line 121: Header title here seems redundant: 
image

src/screens/Tutorial/TutorialSwiper.js
```diff
@@ -0,0 +1,144 @@
...
+      <ScrollView
+        horizontal
+        bounces={false}
+        pagingEnabled
+        style={{ width: DEVICE_WIDTH, flex: 1 }}
```
mdjastrzebski, line 126: Could we try width 100% here?
```suggestion
        style={{ width: `100%`, flex: 1 }}
```

src/screens/Tutorial/TutorialSwiper.js
```diff
@@ -0,0 +1,144 @@
...
+  width: 100%;
+`;
+
+const Img = styled(Image)`
+  margin-top: 40px;
+  width: ${DEVICE_WIDTH * 0.7}px;
+  height: 100px;
+`;
+
+export default withTheme(({
```
mdjastrzebski, line 75: Pls use `useTheme` hook

src/screens/Tutorial/TutorialSwiper.js
```diff
@@ -0,0 +1,144 @@
...
+  onButtonPress, data, onFinish, theme,
+}: Props) => {
+  const scrollViewRef = useRef(null);
+  const [scrollIndex, setScrollIndex] = useState<number>(0);
+  const [userHasScrolled, setUserHasScrolled] = useState<boolean>(false);
+
+  const scrollToIdx = (idx: number) =>
+    scrollViewRef?.current && scrollViewRef.current.scrollTo({ x: idx * DEVICE_WIDTH });
```
mdjastrzebski, line 83:
```suggestion
    scrollViewRef.current?.scrollTo({ x: idx * DEVICE_WIDTH });
```

src/screens/Tutorial/TutorialSwiper.js
```diff
@@ -0,0 +1,144 @@
...
+
+  const handleButtonPress = (val: string) => {
+    onButtonPress(val);
+    scrollToIdx(1);
+  };
+
+  const handleScroll = (e: Object) => {
```
mdjastrzebski, line 90: Could we extract a correct event typing (and expose it thought `utils/types/react-native`?

src/screens/Tutorial/TutorialSwiper.js
```diff
@@ -0,0 +1,144 @@
...
+
+  return (
+    <ContainerWithHeader headerProps={{ noBack: true, centerItems: [{ title: t('title.welcome') }] }}>
+      <ScrollView
+        horizontal
+        bounces={false}
+        pagingEnabled
+        style={{ width: DEVICE_WIDTH, flex: 1 }}
+        ref={scrollViewRef}
+        onScroll={handleScroll}
```
mdjastrzebski, line 128: Should we also use `scrollEventThrottle` prop here?

src/screens/Tutorial/TutorialSwiper.js
```diff
@@ -0,0 +1,144 @@
...
+  };
+
+  const handleScroll = (e: Object) => {
+    const xPosition = e.nativeEvent.contentOffset.x;
+    if (xPosition === 0) {
+      setScrollIndex(0);
+    } else if (xPosition % DEVICE_WIDTH === 0) {
+      setScrollIndex(xPosition / DEVICE_WIDTH);
+    }
+  };
```
mdjastrzebski, line 97: These checks seem a bit fragile with the modulo comparison. Maybe we could use something like this (note it will call useState setter much more frequently, however calling setState with same value theoreticaly should not cause rerender)

```suggestion
   setScrollIndex(Math.floor(xPosition / DEVICE_WIDTH + 0.5))
  };
```

src/screens/Tutorial/TutorialCMSView.js
```diff
@@ -0,0 +1,87 @@
...
+`;
+
+type Props = {
+  theme: Theme,
+  document: ParsedCMSDocument,
+};
+
+const DEVICE_WIDTH = getDeviceWidth();
+
+const CMSView = ({ document }: Props) => {
```
mdjastrzebski, line 56: This does not handle dark mode too well for the images:
image

src/screens/Tutorial/TutorialCMSView.js
```diff
@@ -0,0 +1,87 @@
...
+      bounces={false}
+    >
+      <Title>{title}</Title>
+      <SubTitle>{subtitle}</SubTitle>
+      <Image
+        source={{ uri: imageUrl }}
+        style={getImageStyle()}
+        resizeMode="contain"
+      />
+      <Body>{body}</Body>
```
mdjastrzebski, line 82: Image + Body should benefit from centering them vertically and adding some variable spacing between image and body, as they seem to cramped atm:

image

Suggested spacing:
```
<View style={{ flex: 1}} />
<Image />
<View style={{ flex: 1}} />
<Body />
<View style={{ flex: 1}} />
```

src/screens/Tutorial/TutorialFooter.js
```diff
@@ -0,0 +1,72 @@
...
+  const renderButton = (title: string, handler: ?Function) => (
+    <TouchableOpacity disabled={!handler} onPress={handler} hitSlop={hitSlop10}>
+      <ButtonText>{handler ? title : ''}</ButtonText>
+    </TouchableOpacity>
+  );
+
+  return (
+    <Wrapper>
+      {renderButton(t('button.back'), onBackPress)}
+      {renderButton(t('button.skip'), onSkipPress)}
```
mdjastrzebski, line 68: nit picking as hell: on the last step: skip seems to do exactly the same thing as finish, so it would be beneficial to hide Skip on the last page

image

src/screens/Tutorial/TutorialSwiper.js
```diff
@@ -0,0 +1,144 @@
...
+  const isLast = () => scrollIndex >= data.length;
+
+  const handleNextPress = () => isLast() ? onFinish() : scrollToIdx(scrollIndex + 1);
+
+  const renderInitialScreen = () => (
+    <InitialScreenWrapper>
+      <ContentWrapper>
+        <Title>{t('tutorialTitle')}</Title>
+        <SubTitle>{t('tutorialSubtitle')}</SubTitle>
+        <Img resizeMode="contain" source={images(theme).pillarLogo} />
```
mdjastrzebski, line 108: Pillar Logo would probably benefit from more vertical centering

image

src/utils/cms.js
```diff
@@ -0,0 +1,63 @@
...
+
+const {
+  ONBOARDING_SCREENS_FOR_NATIVES: NATIVES,
+  ONBOARDING_SCREENS_FOR_NEWBIES: NEWBIES,
+} = CMS_DATA_TYPES;
+
+const getSortedTutorialData = (docs: CMSDocument[], type: string): ParsedCMSDocument[] =>
+  orderBy(docs.map(d => parseCMSDocument(d)).filter(d => d.type === type), i => (i.order || 0), ['asc']);
+
+const parseCMSDocument = (doc: CMSDocument): ParsedCMSDocument => {
```
mdjastrzebski, line 33: This function would probably benefit from wrapping in try catch, as e.g. data from 3rd party service could come in wrong format cause app crash: e.g. `title` could not be an array, cause `TypeError` on `title[0]?.text` expression.

mdjastrzebski:
The feature does not seem to work on Android (Pixel 3, Android 11, physical device):

Unable to scroll using swipe gestures, no bottom prev/next buttons:

image

src/screens/Tutorial/TutorialCMSView.js
jdnmr: I believe it's the best way in the current conditions. This component is a part of a horizontal scrollview, so setting width to 100%/stretching ruins the style and texts no longer fit the screen

src/screens/Tutorial/TutorialSwiper.js
jdnmr: That's because all these strings are dummy data I made up. I believe the header might be handy once we have proper strings ready, so I'd leave it. @IAmKio thoughts?

src/screens/Tutorial/TutorialCMSView.js
jdnmr: These are dummy images afaik. I suppose actual ones won't have a transparent filling, or there will be separate ones for light and dark themes like we often do across the app. @IAmKio

mdjastrzebski: Do we support separate images for light/dark mode?

jdnmr: Not directly in this feature as this was not requested and images are provided via links, so I'd expect the images provided to work and look well in both themes. But we do use different images for different themes across the app.
That said, I don't really know how I'd handle this issue, I don't think hard-coding modifications to dynamically fetched images are a good idea here. Like I said, I'd expect the images chosen to looks nicely out of hand. 
Please note we're not launching this feature straight away afaik, this is just a setup, details are still to be settled and some changes will need to be made

src/screens/Tutorial/Tutorial.js
```diff
@@ -0,0 +1,80 @@
...
+type TUTORIAL_PATH = typeof NATIVES | typeof NEWBIES;
+
+const TutorialScreen = ({ hasSeenTutorial, tutorialData }: Props) => {
+  const [activePath, setActivePath] = useState<TUTORIAL_PATH>(NEWBIES);
+  const navigation = useNavigation();
+  const routeName = navigation?.state?.params?.nextNavigationRouteName || HOME;
+
+  useEffect(() => {
+    !tutorialData && navigation.navigate(routeName);
+  }, []);
```
mdjastrzebski, line 54: Using `&&` expression here is non-intuitive.

```suggestion
  useEffect(() => {
    if (!tutorialData) {
      navigation.navigate(routeName);
    }
  }, []);
```

src/screens/Tutorial/TutorialFooter.js
```diff
@@ -0,0 +1,73 @@
...
+  bottom: 0;
+  border-top-width: 1px;
+  border-color: ${({ theme }) => theme.colors.basic060};
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+`;
+
+const Button = styled.TouchableOpacity`
+  flex: 1;
```
mdjastrzebski, line 42: I would go without `flex: 1` here, as buttons looked better when touching the container sides:
`|B-B-B|` rather than `|-B-B-B-|`

mdjastrzebski (approved):
LG 👍🏻 

Added some minor finishing touches though.

src/screens/Tutorial/TutorialFooter.js
jdnmr: Without flex: 1, the buttons "moved" slightly as the footer changed, flex: 1 makes them stay at 1 place. I added changes so that the left and right buttons stay at container sides

IAmKio:
@antonioned @vnovacev @arddluma could you please check the `circleci/config.yml` file please?

arddluma (commented):
@IAmKio  .circleci/config.yml looks good, env variable added to CircleCI: 
image

vnovacev (approved):

arddluma (approved):

PR #2549, opened 2021-03-29
PIL 863 - Sentry fixes

VSaulis:


src/configureStore.js
```diff
@@ -78,11 +79,13 @@ const enhancer = composeWithDevTools({
...
 })(applyMiddleware(...middlewares));
 
+const sentryReduxEnhancer = Sentry.createReduxEnhancer();
+
 const configureStore = (initialState: ?Object): Object => {
   const store = createStore(
     pReducer,
     initialState,
-    enhancer,
+    compose(enhancer, sentryReduxEnhancer),
```
jsidorenko, line 88: I think we should exclude some actions with the private data (private keys, mnemonic, etc) like it's described here: https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/redux/#redux-enhancer-options

jsidorenko: P. S. Overall that change is awesome!

VSaulis: I define one action but i am not sure about all actions that have sensitive data

PR #2550, opened 2021-03-30
Home vNext 1: restructure navigation

mdjastrzebski:
Part of Home vNext feature

Scope:
- [x] Remove tab navigation & restructure top level navigation
- [x] Provide temporary home content and navigation buttons until proper parts are implemented


src/components/modern/Layout.js
```diff
@@ -0,0 +1,36 @@
+// @flow
```
jdnmr, line 1: No license info etc.?

src/screens/Home/Home.js
```diff
@@ -1,788 +1,86 @@
...
-  }
+      <Content>
+        {/* Temporary content */}
+        <MainContent>
+          {/* eslint-disable-next-line i18next/no-literal-string */}
+          <Text>Hello Pillar!</Text>
+        </MainContent>
+
+        {/* Temporary navigation section */}
+        {/* eslint-disable-next-line i18next/no-literal-string */}
```
jdnmr, line 64: nit: we disable this rule plenty of times throughout the codebase, can we disable it globally?

jdnmr (approved):
LG!

src/screens/Home/Home.js
mdjastrzebski: I'm not a fan of this rule, however it seems to be added by intent: https://github.com/pillarwallet/pillarwallet/pull/2213

@IAmKio maybe we could disable this rule it globally, wdyt?

PR #2554, opened 2021-03-31
PIL-991: Feature: Firebase Analytics User Properties

wjkg:


src/actions/authActions.js
```diff
@@ -269,6 +269,9 @@ export const loginAction = (
       }
 
       if (isOnline) {
+        // Send basic app/user info to Firebase Analytics
+        await dispatch(setUserPropertiesAction());
```
IAmKio, line 273: Will this cause a delay when a user is performing the `authActions.js` flow? If so, could we remove the `await` as this feels like it can be a fire-and-forget dispatch?

wjkg: Good point -- the reason is that we'd fetch remote config values for the previous user properties, which could mean having to reload the app twice to see the correct ones depending on the timing, but it makes more sense to group both properties and fetching in one action

IAmKio (approved):
LGTM

PR #2555, opened 2021-03-31
Home vNext 4: floating actions

mdjastrzebski:
Scope:
- [x] Home: floating actions buttons
- [x] FloatingButtons better support for safe area


src/screens/Home/FloatingActions.js
```diff
@@ -0,0 +1,80 @@
...
+import { CONNECT_FLOW, EXCHANGE_FLOW, SEND_TOKEN_FROM_HOME_FLOW } from 'constants/navigationConstants';
+
+// Selectors
+import { useRootSelector } from 'selectors';
+import { totalBalanceSelector } from 'selectors/balances';
+import { useSmartWalletStatus } from 'selectors/smartWallet';
+
+const walletConnectIcon = require('assets/icons/icon-24-wallet-connect.png');
+
+function FloatingActions() {
```
poocart, line 39: Obviously this is by preference, but in our codebase most of the functions or components are written as constant + arrow functions, just wondering if we want to make this consistent. What are your thoughts? 👀

poocart (approved):
LGTM 👍

src/screens/Home/FloatingActions.js
mdjastrzebski: From my perspective this is a minor thing with no impact on code readability, maintainability or usage. So personally I would not impose any hard rules here. @poocart wdyt?

PR #2556, opened 2021-04-01
Fix/cocoapods upgrade

IAmKio:
This is a pre-requisite for Storyly.


.circleci/config.yml
```diff
@@ -759,14 +759,14 @@ workflows:
           filters:
             branches:
               only:
-                - develop
+                - fix/cocoapods-upgrade
```
antonioned, line 762: this needs to be reverted back to develop branch in order for the pipeline to work.

antonioned (changes requested):

.circleci/config.yml
IAmKio: @antonioned thanks mate will update now

antonioned (approved):

mdjastrzebski (approved):
Loogs good

PR #2558, opened 2021-04-01
Home vNext 5: assets

mdjastrzebski:
Scope:
- [x] Display asset category summary
- [x] Calculate actual Archanova balances
- [x] Provide easy way to swap with etherspot balances & profits
- [x] Toggle display side chains
- [x] Common chains and asset categories ui properties


src/models/Home.js
```diff
@@ -0,0 +1,41 @@
...
+*/
+
+import { BigNumber } from 'bignumber.js';
+
+export type WalletInfo = {|
+  ethereum?: ChainInfo,
+  binance?: ChainInfo,
+  xdai?: ChainInfo,
+|};
+export type ChainInfo = {|
```
poocart, line 28:
```suggestion
|};

export type ChainInfo = {|
```

src/models/Home.js
```diff
@@ -0,0 +1,41 @@
...
+  wallet?: BalanceInfo,
+  deposits?: BalanceInfo,
+  investments?: BalanceInfo,
+  liquidityPools?: BalanceInfo,
+  rewards?: BalanceInfo,
+  datasets?: BalanceInfo,
+  collectibles?: number,
+  contacts?: number,
+|};
+export type BalanceInfo = {|
```
poocart, line 38:
```suggestion
|};

export type BalanceInfo = {|
```

src/screens/Home/AssetsSection.js
```diff
@@ -0,0 +1,150 @@
...
+      {renderChain(CHAINS.ETHEREUM, wallet.ethereum)}
+      {renderChain(CHAINS.BINANCE, wallet.binance)}
+      {renderChain(CHAINS.XDAI, wallet.xdai)}
+    </Container>
+  );
+}
+
+export default AssetsSection;
+
+const Container = styled.View``;
```
poocart, line 150: I wonder if this wouldn't cause confusion over having styled components defined level above component in current codebase. What are your thoughts?

poocart (approved):
Good work! 🚀

Few general notes:
1. I've noticed you are adopting 2 new patterns in codebase: defining styled components below component and using regular function over arrow function for components. While I totally understand this preference I'm not sure if this wouldn't cause confusion over existing codebase where it's mostly done based on other preference of styled components used above component definition and all functions being constants (except explicit cases where class is needed).
2. Not sure if this will happen later, but for new formatting methods it would be good to have at least couple tests 👀

src/screens/Home/AssetsSection.js
mdjastrzebski: When reading the existing components, that put styled components first, I've found it hard to find the logic and essence of the actual components when the component size is large (say 200-300+ LOCs). By putting the main component & it's props first, it should be easier to quickly figure out it's external API while leaving the implementation details at the end of the file.

So basically the structure I suggest for new component files is:

1. Import
2. Props & related type defs
3. Main component
4. Helper functions (if there's a lot of them, put them in `utils.js` file or something)
5. Styled components &styles (if there's a lot of them & their reusable, put them in `components.js` file or something)

@poocart wdyt?

mdjastrzebski:
Thanks @poocart 

Regarding tests, there are some added in the next PRs.

PR #2559, opened 2021-04-01
Release Candidate 2.34.11

IAmKio:
## Release Candidate 2.34.11

- Improved: Metrics capture to help better provide and control features within the app
- Improved: Core software libraries updated


src/actions/remoteConfigActions.js
```diff
@@ -41,3 +47,28 @@ export const loadRemoteConfigAction = () => {
...
+  };
+
+  return firebaseAnalytics
+    .setUserProperties(properties)
+    .catch((error) => {
+      reportLog('Firebase Analytics: Failed to set user properties', { error });
+    });
+};
+
+export const loadRemoteConfigWithUserPropertiesAction = () => async (dispatch: Dispatch) => {
```
mdjastrzebski, line 71: It seems that there is no logical connection between two actions, as first sets user context info for firebase analytics while second loads firebase remote config. If there are indeed not connected it would be better to keep them separate.

src/actions/remoteConfigActions.js
```diff
@@ -41,3 +47,28 @@ export const loadRemoteConfigAction = () => {
       .catch(e => { reportLog('Failed to fetch feature flags or initialize with defaults', e); });
   };
 };
+
+export const setUserPropertiesAction = () => (
+  dispatch: Dispatch,
+  getState: GetState,
+) => {
```
mdjastrzebski, line 54: This action seems to be about analytics, and hence would be better to put it into `analyticsActions` file.

mdjastrzebski (approved):
Overall looks safe to release, however I have some concerns about `loadRemoteConfigWithUserPropertiesAction` which seems it mix analytics & remote config actions without any clear connection between the two.

PR #2561, opened 2021-04-07
fix: underflow issue on transaction estimate

mdjastrzebski:
Fix ethers underflow issue that occurred during tx estimation when token amount had decimal places exceeding available decimal places for given token.


jdnmr (approved):
LG
Did the issue occur only during transfers from key based?

mdjastrzebski:
I've noticed the issue for keybased transfers. I guess it can occur in different places as well, but I wanted to limit the change surface area.

PR #2563, opened 2021-04-07
Home vNext 6: icons

mdjastrzebski:
image

Scope:
- [x] Support for SVG icons
- [x] Updated icon files
- [x] Home: copy wallet address


src/components/modern/TextWithCopy.js
```diff
@@ -0,0 +1,83 @@
...
+    </Container>
+  );
+};
+
+const hitSlop = {
+  top: 15,
+  bottom: 15,
+  left: 15,
+  right: 15,
+};
```
jdnmr, line 72: We have some hitSlop objects in `utils/common` already. I think it's a good pattern for the future to either use them or (if you need 15 specifically) add it there instead of declaring them every time

src/components/modern/TextWithCopy.js
```diff
@@ -0,0 +1,83 @@
...
+    if (!textToCopy) return;
+
+    Clipboard.setString(textToCopy);
+    Toast.show({ message: t('toast.addressCopiedToClipboard'), emoji: 'ok_hand' });
+  };
+
+  return (
+    <Container hitSlop={hitSlop} onPress={handleCopyToClipboard} style={style}>
+      {typeof children === 'string' ? <Text>{children}</Text> : children}
+      {textToCopy != null && <CopyIcon name="copy" color={iconColor} width={18} height={18} />}
```
jdnmr, line 62: Why compare to null instead of just `!! textToCopy  && ...`?

src/components/modern/Icon.js
```diff
@@ -0,0 +1,188 @@
...
+
+const components = {
+  // Generic icons
+  'add-cash': IconAddCash,
+  'add-contact': IconAddContact,
+  'arrow-right': IconArrowRight,
+  'arrow-down': IconArrowDown,
+  'arrow-left': IconArrowLeft,
+  'arrow-up': IconArrowUp,
+  bullet: IconBullet,
```
jdnmr, line 101: why are some keys wrapped in `' '`while others aren't?

jdnmr (approved):
LG, see a few questions and suggestions

src/components/modern/Icon.js
mdjastrzebski: legal JS names can be used without quotes when defining an object, however for keys that are not legal JS identifiers you need to put them into quotes.

src/components/FloatingButtons/FloatingButtons.js
```diff
@@ -24,15 +24,15 @@ import styled from 'styled-components/native';
...
+import Icon from 'components/modern/Icon';
 import Text from 'components/modern/Text';
 
 // Utils
 import { spacing } from 'utils/variables';
 
 // Types
 import type { ImageSource } from 'utils/types/react-native';
-import type { IconName } from 'components/Icon';
+import type { IconName } from 'components/modern/Icon';
```
poocart, line 35: I assume `modern` means newer Icon component? If so – what do you think if we replace existing `Icon` into something like `LegacyIcon` and new one can stay `Icon` and be reused with new implementations?

src/components/modern/Icon.js
```diff
@@ -0,0 +1,188 @@
...
+};
+
+export type IconName = $Keys<typeof components>;
+
+type Props = {|
+    name: string;
+    color?: string;
+    width?: number;
+    height?: number;
+|};
```
poocart, line 168:
```suggestion
type Props = {|
  name: string;
  color?: string;
  width?: number;
  height?: number;
|};
```

.svgrrc
```diff
@@ -0,0 +1,5 @@
+{
+  "replaceAttrValues": {
+    "#000": "{props.fill}"
```
poocart, line 3: I assume this wouldn't cause any issues with existing SVG assets?

src/components/modern/TextWithCopy.js
```diff
@@ -0,0 +1,83 @@
...
+const TextWithCopy = ({
+  children,
+  textToCopy,
+  style,
+  iconColor,
+}: Props) => {
+  const { t } = useTranslation();
+
+  const handleCopyToClipboard = () => {
+    if (!textToCopy) return;
```
poocart, line 53: Maybe we can add a toast saying like "Failed to copy" or something, what do you think?

poocart (approved):
Awesome! 💯 

Left few comments and general one would be about vanilla function vs es6 arrow function that I left in other PRs.

src/components/FloatingButtons/FloatingButtons.js
mdjastrzebski: That's a good idea, however I would pause it until the code is actually merged with `develop`, otherwise keeping `vNext` branch in sync would be difficult.

But yeah, the general idea is that existing components keep using the LegacyIcon, while any new components should use modern/Icon.

.svgrrc
mdjastrzebski: This is a replacement color for bundles SVG files, atm the only bundled SVG files are the new icons. Also replacement color is only used if there it's passed through `fill` prop, otherwise it stays the same. So should not cause problems here.

src/components/modern/TextWithCopy.js
mdjastrzebski: Good spot of edge case, however I'd rather disable the copy event if there is no text to copy.

PR #2566, opened 2021-04-09
PIL-998 fix: liquidity pools withdrawal tx fee vs balance check

mdjastrzebski:
Root cause: incorrect check for tx fee vs user ETH balance:

Should be:
* tx fee <= user ETH balance

But was
* tx fee <= user ETH balance - obtained ETH amount (!)

Looks like copy-paste issue from add liqudity screen


IAmKio (approved):
LGTM

PR #2568, opened 2021-04-12
Fee label toggle issue

divyaBacancy:
https://user-images.githubusercontent.com/66996635/114372875-a8db9400-9b9f-11eb-948b-f5d647958795.mov


src/components/FeeLabelToggle/FeeLabelToggle.js
```diff
@@ -100,7 +100,7 @@ const FeeLabelToggle = ({
   isGasTokenSupported,
   hasError,
 }: Props) => {
-  const [isFiatValueVisible, setIsFiatValueVisible] = useState(showFiatDefault);
+  const [isFiatValueVisible, setIsFiatValueVisible] = useState(!showFiatDefault);
```
mdjastrzebski, line 103: This code change is incorrect, as it causes `showFiatDefault` prop to cause token value to be shown(!)

The correct solution would be to:
1) Remove `showFiatDefault` prop from this component and all it's use cases
2) Initialising `isFiatValueVisible` to `true`

mdjastrzebski (changes requested):
The logic change is incorrect, as it just flips the `showFiatDefault` prop meaning. Please check comment below for scope of correct implementation.

src/components/FeeLabelToggle/FeeLabelToggle.js
divyaBacancy: @mdjastrzebski Thanks for bringing this. This has been done.

mdjastrzebski: Please also remove `showFiatDefault` prop from all `FeeLabelToggle` use-cases (~14 files)

divyaBacancy: Sorry. I checked with yarn run flow. It did not show any error so I missed this out. Done

mdjastrzebski:
Thanks for updating the implementation. In order to complete the PR you need to have all automated tests passing on the CI.

Atm snapshot tests are failing and you need to update them (`yarn test -u`). However it seems that the snapshots for the component are outputting `£NaN`. In order to fix it, you probably need to provide rates redux mock data (pls, see example of how to do it in `EventDetails.stories.js` file).

Fyi we also have means of visually texting the components by using Storybook (Menu -> Storybook + and implmentation is in `*.stories` files - the same as used also for snapshot tests), pls check it out.

src/utils/common.js
```diff
@@ -555,7 +555,7 @@ export const getFormattedTransactionFeeValue = (feeInWei: string | number | BigN
 };
 
 export const formatTransactionFee = (feeInWei: string | number | BigNumber, gasToken: ?GasToken): string => {
-  if (!feeInWei) return '';
+  if (!feeInWei) return '0';
```
mdjastrzebski, line 558: IMO it makes more sense to keep the existing code that returned empty string than to return '0'.

storybook/__snapshots__/storyshots.test.js.snap
```diff
@@ -38949,7 +38949,7 @@ exports[`Storyshots FeeLabelToggle default 1`] = `
               ]
             }
           >
-            tokenValue
+            £NaN
```
mdjastrzebski, line 38952: Pls handle these `NaN` outputs by providing the correct redux `rates` state in `FeeLabelToggle.stories.js`, in order for snapshot test to calculate some valid (non-`NaN` value).

mdjastrzebski (changes requested):
Generally looks good.

Pls handle the `NaN` outputs from storyshots by providing correct mocked `rates` redux state in `FeeLabelToggel.stories.js` (you can check how to do it in e.g. `EventDetails.stories.js`).

src/utils/common.js
divyaBacancy: Thank you for reviewing. I will try to have other work around for NaN issue. I changed this thinking empty string is returning NaN ( not a number) issue. I also looked into one of the screen where NaN was coming and it resolved after changing this condition.

storybook/__snapshots__/storyshots.test.js.snap
mdjastrzebski: It seems that the reason for NaN value is that under the tests `formatTransactionFee(txFeeInWei, gasToken)` in FeeLabelToggle.js returns `'tokenValue'` string, which later causes `parseFloat(feeDisplayValue)` to return `NaN`. 

Looking into details it seems that the culprit is that we try to parse user visible string output from `formatTransactionFee`. This parsing is used because we need to go from transaction amount in Wei to amount in ETH. Simplies workaround would be to get the ETH amount by using something like:

```
const feeValue = getFormattedTransactionFeeValue(txFeeInWei, gasToken);
...

const feeInFiat = parseFloat(feeValue) * getRate(rates, gasTokenSymbol, fiatCurrency);
```
While keeping the `feeDisplayValue` for UI display purposes only.

mdjastrzebski (commented):
Added idea of potential workaround for rates calculation issues.

storybook/__snapshots__/storyshots.test.js.snap
divyaBacancy: Thanks. It worked.

divyaBacancy: I have pushed the changes. I am no longer getting NaN. However my 2 test cases from authActions.test.js are failing even in develop branch where no changes have been done. I am assuming this is a machine specific issue. Hence I have skipped for now.

src/actions/__tests__/authActions.test.js
```diff
@@ -124,7 +124,7 @@ describe('Auth actions', () => {
     });
   });
 
-  it('should expect series of actions with payload to be dispatch on checkPinAction execution', () => {
+  it.skip('should expect series of actions with payload to be dispatch on checkPinAction execution', () => {
```
mdjastrzebski, line 127: Please do not skip these tests, they seem to fail only on your machine. I've checked and they pass on mine, and probably on CI as well.

src/components/FeeLabelToggle/FeeLabelToggle.stories.js
```diff
@@ -25,11 +25,32 @@ import FeeLabelToggle from './FeeLabelToggle';
...
 import CenterViewDecorator from '../../../storybook/CenterViewDecorator';
 
+const reduxData = {
+  rates: {
+    BTC: {
+      ETH: 27.228736, EUR: 53175, GBP: 46132, USD: 63591,
+    },
+    ETH: {
+      ETH: 1, EUR: 2017.2, GBP: 1756.24, USD: 2407.1,
+    },
```
mdjastrzebski, line 35: You only need `ETH: { }` top level rates key, please remove the others as they are noise.

src/components/FeeLabelToggle/FeeLabelToggle.stories.js
```diff
@@ -25,11 +25,32 @@ import FeeLabelToggle from './FeeLabelToggle';
...
+      ETH: 27.233989, EUR: 53190, GBP: 46151, USD: 63604,
+    },
+  },
+};
+
 storiesOf('FeeLabelToggle', module)
   .addDecorator(CenterViewDecorator)
   .addDecorator(WithThemeDecorator)
   .add('default', () => (
     <FeeLabelToggle
```
mdjastrzebski, line 52: This `reduxData` is not passed correctly to the `FeeLabelToggle` as you use the FeeLabelToggle that is exported with `react-redux` `connect` HOC, which causes `connect` HOC to overwrite passed `reduxData` props.

In order to pass redux data this way, please add named export in FeeLabelToggle and import that one only for tests. 

You will then need to supply dummy data for remaining redux fields expected by FeeLabelToggle like this:
```
const reduxData = {
  accountAssets: {},
  accountHistory: [],
  baseFiatCurrency: defaultFiatCurrency,
  isGasTokenSupported: true,
  rates: {
    ETH: {
      ETH: 1,
      EUR: 2017.2,
      GBP: 1756.24,
      USD: 2407.1,
    },
  },
};
```

src/components/FeeLabelToggle/FeeLabelToggle.js
```diff
@@ -93,21 +92,20 @@ const FeeLabelToggle = ({
...
-  const [isFiatValueVisible, setIsFiatValueVisible] = useState(showFiatDefault);
+  const [isFiatValueVisible, setIsFiatValueVisible] = useState(true);
 
   if (isLoading) {
     return <Spinner size={20} trackWidth={2} />;
   }
 
   const fiatCurrency = baseFiatCurrency || defaultFiatCurrency;
-  const feeDisplayValue = formatTransactionFee(txFeeInWei, gasToken);
+  const feeDisplayValue = getFormattedTransactionFeeValue(txFeeInWei, gasToken);
```
mdjastrzebski, line 108: This is incorrect as this causes the FeeLabelToggle to display amount without token. You need two separate variables:
`const feeDisplayValue = formatTransactionFee(txFeeInWei, gasToken);` for display label
`const feeValue = getFormattedTransactionFeeValue(txFeeInWei, gasToken);` for `parseFloat` input.

mdjastrzebski (changes requested):
These are still some open issues in the PR:

1. Despite passing `rates` to the component storyshot output shows `£0.00`, please check relevant comment about passing data to component wrapped in `connect` HOC.
2. Please fix the display of token value, as atm it displays just the amount without the token symbol.
3. Please re-enable skipped tests, as these seem to work fine outside of your local machine.

divyaBacancy:
@mdjastrzebski Thank you for reviewing.The changes has been done.

src/components/FeeLabelToggle/FeeLabelToggle.stories.js
```diff
@@ -20,16 +20,33 @@
 
 import * as React from 'react';
 import { storiesOf } from '@storybook/react-native';
+import { defaultFiatCurrency } from 'constants/assetsConstants';
 
-import FeeLabelToggle from './FeeLabelToggle';
+import { FeeLabelToggleComponent as FeeLabelToggleStoryItem } from './FeeLabelToggle';
```
mdjastrzebski, line 25: It seems that there is no reason for renaming import.

storybook/__snapshots__/storyshots.test.js.snap
```diff
@@ -39077,7 +39077,7 @@ exports[`Storyshots FeeLabelToggle not enough token 1`] = `
               ]
             }
           >
-            tokenValue
+            £17562.40
```
mdjastrzebski, line 39080: 👍🏻

mdjastrzebski (approved):
Looks good!

One very minor issue:
instead of `import { FeeLabelToggleComponent as FeeLabelToggleStoryItem } from ...` please use non-renaming import: `import { FeeLabelToggleComponent } from ...`. There seems to be no particular reason for renaming it, while it makes the code harder to read.

src/components/FeeLabelToggle/FeeLabelToggle.stories.js
divyaBacancy: @mdjastrzebski Thank you for reviewing.The changes has been done.

PR #2534, opened 2021-03-22
PIL-891: Etherspot full implementation

poocart:
This is main branch where all smaller PRs will be pointed to.

[] Etherspot Accounts
[] Etherspot transactions + estimates
[] Etherspot ENS
[] Other UI changes
[] Archanova<>Etherspot transaction dispatcher

PR #2572, opened 2021-04-20
Feature/etherspot balances

poocart:
Part of #2534.

This includes:
- Added balances for Etherspot accounts.
- Services are being soft killed (per chat with @IAmKio) by disabling their visibility while Etherspot account is active. Therefore balances of services are now fetched and Home screen widgets are shown only if Archanova account is active.
- Other inevitable changes while refactoring from `smartWallet` to `archanova` naming.

This PR doesn't cover all `smartWallet` to `archanova` naming refactor and Etherspot transactions, history, Wallet Connect.


storybook/__snapshots__/storyshots.test.js.snap
```diff
@@ -2033,7 +2033,7 @@ exports[`Storyshots ActivityFeedItem Collectible sent from KW to SW 1`] = `
                       ]
                     }
                   >
-                    label.sent
```
mdjastrzebski, line 2036: This hints that maybe some logic has been inverted here from sent collectible -> received collectible. Please verify root cause here.

src/utils/accounts.js
```diff
@@ -111,9 +111,15 @@ export const isSmartWalletAccount = (account: ?Account): boolean => [
   ACCOUNT_TYPES.ETHERSPOT_SMART_WALLET,
 ].includes(account?.type);
 
-export const isKeyBasedAccount = (account: Account): boolean => account.type === ACCOUNT_TYPES.KEY_BASED;
+export const isAccountType = (account: ?Account, type: string): boolean => account?.type === type;
 
-export const isArchanovaAccount = (account: ?Account): boolean => account?.type === ACCOUNT_TYPES.SMART_WALLET;
+export const isArchanovaAccount = (
+  account: ?Account,
+): boolean => isAccountType(account, ACCOUNT_TYPES.SMART_WALLET);
```
mdjastrzebski, line 118: Consider renaming the const for clarity
```suggestion
): boolean => isAccountType(account, ACCOUNT_TYPES.ARCHANOVA_SMART_WALLET);
```

src/utils/accounts.js
```diff
@@ -111,9 +111,15 @@ export const isSmartWalletAccount = (account: ?Account): boolean => [
   ACCOUNT_TYPES.ETHERSPOT_SMART_WALLET,
 ].includes(account?.type);
 
-export const isKeyBasedAccount = (account: Account): boolean => account.type === ACCOUNT_TYPES.KEY_BASED;
+export const isAccountType = (account: ?Account, type: string): boolean => account?.type === type;
```
mdjastrzebski, line 114: Consider inlining the function as it's both short and trival

src/selectors/history.js
```diff
@@ -42,15 +43,17 @@ export const accountHistorySelector = createSelector(
...
   (history, accounts) => {
-    const userHasSW = userHasSmartWallet(accounts);
-    if (!userHasSW) return [];
-    const smartAccountId = getActiveAccountId(accounts);
-    if (!smartAccountId) return [];
-    return orderBy(history[smartAccountId] || [], ['createdAt'], ['desc']);
+    const userHasArchanovaAccount = findFirstArchanovaAccount(accounts);
+    if (!userHasArchanovaAccount) return [];
+
+    const archanovaAccountId = getActiveAccountId(accounts);
```
mdjastrzebski, line 53: How do we know that active account is actually archanova?

src/locales/en/common.json
```diff
@@ -1501,7 +1501,8 @@
     "liquidityPoolsRemoveLiquidity": "{{value}} {{token}} removed from liquidity.",
     "liquidityPoolsStake": "{{value}} {{token}} staked.",
     "liquidityPoolsUnstake": "{{value}} {{token}} removed from liquidity.",
-    "liquidityPoolsClaimRewards": "{{value}} {{token}} claimed."
+    "liquidityPoolsClaimRewards": "{{value}} {{token}} claimed.",
+    "noActiveAccountFound": "No active account found. Please"
```
mdjastrzebski, line 1505: This string seems unfinished, please complete the sentence or remove hanging "Please"

src/screens/Services/Services.js
```diff
@@ -126,95 +125,116 @@ class ServicesScreen extends React.Component<Props> {
...
   }
 
-  getServices = (): Service[] => {
-    const {
-      navigation,
-      isActiveAccountSmartWallet,
-      isSmartWalletActivated,
-      accounts,
-    } = this.props;
+  navigateToArchanovaSupportedServiceScreen = (route: string) => {
```
mdjastrzebski, line 128:
```suggestion
  navigateToRouteIfArchanovaWallet = (route: string) => {
```

src/screens/Services/Services.js
```diff
@@ -126,95 +125,116 @@ class ServicesScreen extends React.Component<Props> {
...
+    const { navigation, isArchanovaWalletActivated } = this.props;
+    if (!isArchanovaWalletActivated) return;
```
mdjastrzebski, line 130: Is this check correct? It seems that it will redirect the user if he has archanova account regardles whether it's currently active or not.

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -241,11 +238,10 @@ export class ActivityFeedItem extends React.Component<Props> {
     } = this.props;
 
     if (isForAllAccounts) {
-      const isBetweenAccounts = (isSWAddress(to, accounts) && isKWAddress(from, accounts))
-        || (isSWAddress(from, accounts) && isKWAddress(to, accounts));
-      return (isBetweenAccounts && isReceived)
-        || (!isBetweenAccounts && isKWAddress(to, accounts))
-        || (!isBetweenAccounts && isSWAddress(to, accounts));
+      const isBetweenAccounts = isSmartWalletAccountAddress(to, accounts)
+        && isSmartWalletAccountAddress(from, accounts);
```
mdjastrzebski, line 242: The logic here differs significantly form previous check which was checking whether the transactions was KW <-> SW, and now it checks SW <-> SW. Is this the intent here?

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -880,27 +875,10 @@ export class ActivityFeedItem extends React.Component<Props> {
...
-            } else {
-              additionalInfo.itemImageSource = event.accountType === ACCOUNT_TYPES.KEY_BASED
-                ? keyWalletIcon
-                : smartWalletIcon;
-            }
-            additionalInfo.isBetweenAccounts = true;
-          }
 
-          if (!isTrxBetweenAccounts) {
+          if (isBetweenSmartWalletAccounts) {
```
mdjastrzebski, line 879: Similar to one of the previous comments, this seems to replace KW <-> SW check for SW <-> SW check. Pls make sure this is intended.

src/components/EventDetails/EventDetails.js
```diff
@@ -877,27 +872,31 @@ export class EventDetail extends React.Component<Props> {
...
-            : [
-              {
-                title: t('button.send'),
-                onPress: this.sendSynthetic,
-                secondary: true,
-              },
-              topUpMoreButton,
-            ],
-        };
+        if (isArchanovaAccountActive) {
```
mdjastrzebski, line 881: Consider inverting the if for readability
```suggestion
        if (!isArchanovaAccountActive) break;
```

src/components/EventDetails/EventDetails.js
```diff
@@ -907,27 +906,31 @@ export class EventDetail extends React.Component<Props> {
...
-        eventData = {
-          buttons: [
-            {
-              title: t('button.withdrawMore'),
-              onPress: this.PPNWithdraw,
-              secondary: true,
-            },
-          ],
-        };
+        if (isArchanovaAccountActive) {
```
mdjastrzebski, line 909: Similarly, consider inverting the if for readability here and in next occurrences.

src/components/EventDetails/EventDetails.js
```diff
@@ -1328,9 +1337,13 @@ export class EventDetail extends React.Component<Props> {
           }
         }
     }
+
+    if (!eventData) return null;
```
mdjastrzebski, line 1341: Please check that EventDetails works correctly if null is returned here?

src/screens/Exchange/Exchange.js
```diff
@@ -339,9 +340,10 @@ class ExchangeScreen extends React.Component<Props, State> {
     this.emptyMessageTimeout = null;
   };
 
-  checkIfAssetsExchangeIsAllowed = () => {
-    const { accounts, smartWalletState, isActiveAccountSmartWallet } = this.props;
-    if (!isActiveAccountSmartWallet) return true;
+  isAssetsExchangeAllowed = (): boolean => {
+    const { accounts, smartWalletState, activeAccount } = this.props;
+    if (!isArchanovaAccount(activeAccount)) return true;
```
mdjastrzebski, line 345: Are we not allowing exchange for etherspot atm?

mdjastrzebski (approved):
Looks good man! 🚀 🌔 

I've added some comments in places that I think deserve some extra scrutiny and might contain subtle logic issues.

src/utils/accounts.js
poocart: It's actually renamed in other PR, leaving unchanged here 👍

src/screens/Exchange/Exchange.js
poocart: We are, this is a check `if there's no Archanova account then it returns true, otherwise if it's Archanova then there's additional check below if Archanova is deployed`.

storybook/__snapshots__/storyshots.test.js.snap
poocart: It was old case when we were showing assets movement history between accounts and it was shown as "received" on one account and "sent" on other, however we no longer have this in place for key based so this is why it changes from `sent` (per story it was sent from KW) to `received ` because it was just an asset received to SW (Archanova wallet) and it's no longer marked as asset moving between accounts, hope this makes sense.

src/locales/en/common.json
poocart: Whoopsy! 👀

src/screens/Services/Services.js
poocart: Changed to `navigateToRouteIfArchanovaWalletActivated` since it's actually a check if Archanova wallet is activated (deployed) and can use the service.

src/selectors/history.js
poocart: Remember this piece, it had to be little different, it shouldn't check for active account ID, but check for found Archanova account ID, fixed, thanks 👍

src/screens/Services/Services.js
poocart: Yes, this check is to determine if Archanova wallet is activated (deployed) before it can use any of service (see "requires activation" label for each service), hope it makes sense.

src/components/EventDetails/EventDetails.js
poocart: It does, it just doesn't render the bottom modal. The thing is it's actually set to `null` before `switch` and it would still return null if it didn't change after `switch` which won't happen as `default` still adds some data over it. Just to avoid more noise with additional checks in the code below where `eventData` has final appends and I just went straight forward with this check. You're aware that the code here is very messy, but didn't want to rewrite it (plus the fact that you changed history components with V2). Hope it makes sense.

src/components/ActivityFeed/ActivityFeedItem.js
poocart: We do not support KW anymore and it shouldn't stay in accounts that's why this check leaves us only with Smart Wallets (Archanova<>Etherspot), hope it makes sense.

src/components/ActivityFeed/ActivityFeedItem.js
poocart: KW is not supported, all KW code are leftovers (except migration).

PR #2574, opened 2021-04-20
PIL-993 History: UI (+ archanova integration)

mdjastrzebski:
Scope:

- [x] New History UI for use with Etherspot (mock data atm)
  - [x] History List
  - [x] History Event Detail bottom modal 
- [x] Reuse old ActivityFeed for Archanova
- [x] Reuse HistoryList in Asset / Collectible screen (when data available)


src/components/HistoryEventDetails/layouts/TokenTransactionEventDetails.js
```diff
@@ -0,0 +1,142 @@
...
+    navigation.navigate(SEND_TOKEN_FROM_HOME_FLOW, {
+      contact: {
+        ethAddress: address,
+        name: ensRegistry[address] ?? address,
+        ensName: ensRegistry[address],
+      },
+    });
+  };
+
+  if (event.type === 'tokenReceived') {
```
poocart, line 74: I wonder if we should extract event types as constants, wdyt?

src/components/HistoryEventDetails/layouts/WalletEventDetails.js
```diff
@@ -0,0 +1,123 @@
...
+  const dispatch = useDispatch();
+
+  const colors = useThemeColors();
+
+  const openActivate = () => {
+    Modal.open(() => <SWActivationModal navigation={navigation} />);
+  };
+
+  const openTopUp = () => {
+    const smartWalletAddress = getSmartWalletAddress(accounts);
```
poocart, line 75:
```suggestion
    // TODO: archanova wallet only
    const smartWalletAddress = getSmartWalletAddress(accounts);
```

src/components/HistoryEventDetails/layouts/WalletEventDetails.js
```diff
@@ -0,0 +1,123 @@
...
+    if (!smartWalletAddress) return;
```
poocart, line 76: Wdyt of showing warning toast here?

src/components/HistoryEventDetails/layouts/TokenTransactionEventDetails.js
poocart: Also it would give us full picture of all even types

src/utils/linking.js
```diff
@@ -0,0 +1,53 @@
...
+import { Linking } from 'react-native';
+import t from 'translations/translate';
+
+// Components,
+import Toast from 'components/Toast';
+
+// Services
+import smartWalletService from 'services/smartWallet';
+
+export function viewOnBlockchain(hash: ?string) {
```
poocart, line 30: I would move this to `common` or `transactions` utils as `linking` doesn't say much/reflect the contents and it would most likely will contain this method only

src/utils/format.js
```diff
@@ -163,3 +184,37 @@ export function formatFiatChangeExtended(change: ?BigNumber, initialBalance: ?Bi
...
+
+  const decimalPlaces = getDecimalPlaces(symbol);
+  const formattedValue = formatValue(value.abs(), { decimalPlaces, ...options });
+
+  return value.gte(0)
+    ? t('positiveTokenValue', { value: formattedValue, token: symbol })
+    : t('negativeTokenValue', { value: formattedValue, token: symbol });
+}
+
+export function formatHexAddress(address: string) {
```
poocart, line 218: There's `humanizeHexString` on `utils/common` and `elipsizeAddress` on `utils/feedData`, maybe we should either get rid of others and replace with single method or mark those as deprecated, wdyt?

src/selectors/smartWallet.js
```diff
@@ -78,3 +78,15 @@ export const useSmartWalletStatus = (): SmartWalletStatus => {
   const smartWalletState = useRootSelector((root) => root.smartWallet);
   return getSmartWalletStatus(accounts, smartWalletState);
 };
+
+// Temporary stuff to get info about smart wallet type
```
poocart, line 82:
```suggestion
// Temporary stuff to get info about smart wallet type
// TODO: combine this with etherspot work once it's in
```

src/locales/en/common.json
```diff
@@ -263,6 +268,7 @@
     "depositMore": "Deposit more",
     "explore": "Explore",
     "exchange": "Exchange",
+    "exchangeMore": "Exchange More",
```
poocart, line 271: I assume `More` starts with capital M by design since `depositMore` `more` starts with lowercase 👀

src/components/modern/Button.js
```diff
@@ -0,0 +1,74 @@
...
+
+export default Button;
+
+const TouchableContainer = styled(TouchableOpacity)`
+  justify-content: center;
+  align-items: center;
+  border-radius: 6px;
+  width: 100%;
+  padding: 14px ${spacing.large}px;
+  ${({ theme, variant }) => (variant === 'primary' ? `background-color: ${theme.colors.buttonPrimaryBackground}` : '')};
```
poocart, line 64: This and below can be replaces to just this:
```suggestion
  ${({ theme, variant }) => (variant === 'primary' && `background-color: ${theme.colors.buttonPrimaryBackground}`)};
```

src/components/HistoryList/items/TokenTransactionItem.js
```diff
@@ -0,0 +1,95 @@
...
+    const ensName = findEnsNameCaseInsensitive(ensRegistry, event.fromAddress);
+
+    return (
+      <HistoryListItem
+        iconName="arrow-down"
+        iconColor={colors.positive}
+        iconBorderColor={colors.positiveWeak}
+        title={ensName ?? formatHexAddress(event.fromAddress)}
+        valueComponent={
+          <TokenValueView value={event.value.value} symbol={event.value.symbol} variant="medium" mode="change" />
```
poocart, line 61: I wonder if we can add helper in the above for this (and other items possibly) to destruct `value`, wdyt? Also it looks like mistype, on the helper it could be mentioned that this is intended, wdyt?

src/components/HistoryList/items/HistoryListItem.js
```diff
@@ -0,0 +1,151 @@
...
+  iconComponent,
+  valueComponent,
+  status,
+  onPress,
+}: Props) {
+  const colors = useThemeColors();
+
+  return (
+    <TouchableContainer onPress={onPress} disabled={!onPress}>
+      <LeftColumn>
```
poocart, line 66: Is this ok if there's no `iconUrl`, `iconName` and no `iconComponent` to leave it as it is since it occupies defined with alone?

poocart (approved):
Stunning work! 🚀 Left few notes to take a look 👀

src/components/HistoryList/items/TokenTransactionItem.js
mdjastrzebski: I've added destructuring, indeed it improved readability.

I'm not sure though what do you exactly mean by "mistype" here? Pls elaborate.

src/components/HistoryEventDetails/layouts/WalletEventDetails.js
mdjastrzebski: I could add one, but what kind of message would you display there?

src/utils/linking.js
mdjastrzebski: Linking refers to the fast that it uses `Linking` utility from React Native. I'll try to figure out something more descriptive. I didn't want to move this to `common` as `common` is IMO too messy atm with all the various staff that's there.

src/components/HistoryEventDetails/layouts/WalletEventDetails.js
poocart: Something like "Cannot get wallet address" with support link, wdyt? (cc @IAmKio)

src/components/HistoryList/items/TokenTransactionItem.js
poocart: For the same thing (double value key), but I guess if you destructed it should be fine 👍

mdjastrzebski: Yeah naming here is tricky, so the type is `TokenValue` which is `value: BigNumber` + `symbol: string`, which kinda seems alright, but then you have `TokenReceivedEvent` with `value: TokenValue` field, and then it get's a bit awkward, but I guess this is quite minor and mostly aesthetic thing

src/components/HistoryList/items/HistoryListItem.js
mdjastrzebski: IMO it makes sense, because this is a list item and either we block the place or move rest of the element around. Blocking the space IMO seems better solution.

PR #2578, opened 2021-04-27
PIL-988 Value Input improvements

divyaBacancy:
https://user-images.githubusercontent.com/66996635/116246650-cefc5900-a787-11eb-86b2-8a2d58aaadfd.mov


src/containers/SendContainer/SendContainer.js
```diff
@@ -115,7 +115,7 @@ const SendContainer = (props: Props) => {
     <ContainerWithHeader
       headerProps={{ centerItems: [{ title: customScreenTitle || t('transactions.title.sendScreen') }] }}
       footer={<SendFooter {...footerProps} />}
-      minAvoidHeight={800}
+      minAvoidHeight={900}
```
mdjastrzebski, line 118: Could you explain why have you tweaked this value? Both previous value and this one seem to be quite random.

storybook/__snapshots__/storyshots.test.js.snap
```diff
@@ -47778,7 +47778,7 @@ exports[`Storyshots TextInput default 1`] = `
                     "borderTopWidth": 0,
                     "borderWidth": undefined,
                     "flexDirection": "row",
-                    "height": 54,
+                    "height": 62,
```
mdjastrzebski, line 47781: This change seems to cause input value to break into two lines. AFAIK it should stay in one line, pls consult with Dmitry if necessary.


image

src/components/ValueInput/ValueInput.js
```diff
@@ -220,6 +222,9 @@ export const ValueInputComponent = ({
   const openAssetSelector = () => {
     Keyboard.dismiss();
 
+    if (onValueChange) {
```
mdjastrzebski, line 225: I see two issue with this line:
1. When user opens asset selector, but then uses the back button to dismiss it the value is reset to zero.
2. It does not seem to work, when the ValueInput is in fiat input mode, opening the modal resets the token amount but not fiat value.

Consider clearing both token and fiat value when the selected token actually changes. That should address both of the issues above.

src/components/ValueInput/ValueInput.js
```diff
@@ -188,6 +189,7 @@ export const ValueInputComponent = ({
   };
 
   const handleUsePercent = async (percent: number) => {
+    Keyboard.dismiss();
```
mdjastrzebski, line 192: Pleased add `keyboardShouldPersistTaps` for SendContainer, as now when in Send screen, tapping max value (not percent amount) first cause keyboard dismiss but not applcation of max value, only second tap cause ValueInput to apply max value.

src/components/TextInput/TextInput.js
```diff
@@ -163,7 +163,7 @@ const AddonRegularText = styled(BaseText)`
 const AddonIcon = styled(Icon)`
   font-size: 16px;
   color: ${({ theme }) => theme.colors.basic010};
-  margin-right: 9px;
+  padding-right: 9px;
```
mdjastrzebski, line 166: Is that change necessary? It does not seem to cause any visual effect?

mdjastrzebski (changes requested):
I've noticed couple of issues with changes:
1. When focused out, a long (token) value display in two lines, while it should display in one line, with fron of the numbe being visible (AC 4.)
2. Token resetting occurs too early when opening the token selector and not when actually changing token
3. Token resetting does not seem to work well when `ValueInput` is in fiat input mode
4. On Send Asset screen taping max value (not percent value) causes keyboard dismiss without applying max value, pls fix by applying `keyboardShouldPersistTaps` prop in `SendContainer`
5. Second part of AC 4 is also not implemented: "4. When input contains too large value to fit at once, ... place the cursor at the beginng when focussing".
6. There are some code changes which does not seem to have direct connection with required functionality changes

src/containers/SendContainer/SendContainer.js
divyaBacancy: Indeed both previous and current value is random. ContainerWithHeader checks this value internally with screen height and minAvoidHeight is used for smaller screens if the screen is small then leave footer area locked at the bottom and otherwise if screen is big enough then lock footer items above keyboard. 
Will revert the changes and will find another approach.

storybook/__snapshots__/storyshots.test.js.snap
divyaBacancy: This was as per the discussion with Dmitry @paragramm. Dmitry can you please confirm.

src/components/ValueInput/ValueInput.js
divyaBacancy: I will look into this. Also let me know point 2, fiat input mode.

src/components/TextInput/TextInput.js
divyaBacancy: This was regarding 5 point.

poocart (commented):
@mdjastrzebski comments makes sense, will take a look one more time once those are solved 👌

divyaBacancy:
@poocart  @mdjastrzebski  @IAmKio  All the changes have been done. In 4th point the front part is displayed for large values. However when onFocus have not brought the cursor at beginning.

divyaBacancy:
https://user-images.githubusercontent.com/66996635/116512549-d46fb580-a8e5-11eb-8dbe-4dc51f55cd8b.mov

storybook/__snapshots__/storyshots.test.js.snap
IAmKio: @paragramm could you please confirm this please mate?

src/components/ValueInput/ValueInput.js
```diff
@@ -268,6 +269,14 @@ export const ValueInputComponent = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [errorMessage]);
 
+  React.useEffect(() => {
+    if (onValueChange) {
+      onValueChange('0');
+      setValueInFiat('0');
+    }
```
mdjastrzebski, line 276:
```suggestion
    onValueChange?.('0');
    setValueInFiat('0');
```

divyaBacancy: @mdjastrzebski Thank you for reviewing. The changes has been done.

mdjastrzebski:
Points 1, 2, 3, and 6 from my previous comments are fixed now.

However there are some issues with point 4:
1. Tapping max value when keyboard is open, does only cause keyboard to be dismissed and not max value to be applied. It looks like appling ScrollView with `keyboardShouldPersistTaps` did not help here, please consider removing that change and trying something different. 

https://user-images.githubusercontent.com/6368606/116856738-ea3ded00-abfb-11eb-8015-c89bddad1ad0.mov

2. Pls be aware that this feature previously worked for Exchange screen and now is not working, so it needs to be restored.

https://user-images.githubusercontent.com/6368606/116857183-b1eade80-abfc-11eb-8471-d8c60af0503b.mov

3. ScrollView added to SendContainer causes incorrect scroll view content size to be applied when keyboard is open:

https://user-images.githubusercontent.com/6368606/116856978-53256500-abfc-11eb-9a06-908c7959b473.mov

4. Something weird is happening on blur event now, as it seems to append "." after blur, see move for point 2 and also here:

https://user-images.githubusercontent.com/6368606/116857391-02623c00-abfd-11eb-8291-889350408d41.mov

divyaBacancy:
The scroll view issue in send container also exist in develop branch.

https://user-images.githubusercontent.com/66996635/117031682-73e5db80-ad1e-11eb-9b9d-a9f27eeff671.mov

divyaBacancy:
I was not able to reproduce the dot issue ( append "." after blur). However I have fixed scrolling and tapping max value issue.

https://user-images.githubusercontent.com/66996635/117031972-bb6c6780-ad1e-11eb-9ac9-44c9d305b3bb.mov

https://user-images.githubusercontent.com/66996635/117032056-cfb06480-ad1e-11eb-96ed-31c7df4502b1.mov

mdjastrzebski (approved):
I've run the PR on both iOS and Android.

On iOS everything looks good, there is no "." issue (I think it could be related to my runtime env), clicking max in send container fills the value & dismisses the keyboard in the first try, there is no send container scroll issue.

On Android however the value shows only trailing part of the number, instead of leading one:
image. Otherwise looks good.

divyaBacancy:
I have made changes for Android issue ( the value shows only trailing part of the number, instead of leading one).

https://user-images.githubusercontent.com/66996635/117472134-354a5e00-af76-11eb-86b4-6372416fad6c.mov

Also, I have made some changes in handleBlur of TextInput. It seems on Android handleblur will never be called as every time this condition 
if (Platform.OS === 'android' && e.nativeEvent.text === undefined) {
    return;
}
fails.
Have made some changes. @all Let me know what are your views on this ?

src/components/TextInput/TextInput.js
```diff
@@ -229,22 +230,19 @@ class TextInput extends React.Component<Props, State> {
...
-    if (Platform.OS === 'android' && e.nativeEvent.text === undefined) {
-      return;
-    }
-
-    const { inputProps: { onBlur, selectorValue = {} }, trim } = this.props;
+  handleBlur = () => {
+    const { inputProps: { onBlur, selectorValue = {}, value }, trim } = this.props;
     const { selector } = selectorValue;
-    const value = trim ? e.nativeEvent.text.trim() : e.nativeEvent.text;
+    const value1 = trim ? value.toString().trim() : value;
```
mdjastrzebski, line 240:
```suggestion
    const trimmedValue = trim ? value.toString().trim() : value;
```

mdjastrzebski (approved):
Looks good on both iOS and Android

PR #2580, opened 2021-04-29
PIL-985 Fiat value display

divyaBacancy:


divyaBacancy:
Uploading PIL-985.mov…

divyaBacancy:
https://user-images.githubusercontent.com/66996635/116555996-99d44000-a91a-11eb-967d-fbedcfcfd659.mov

src/utils/common.js
```diff
@@ -277,16 +277,22 @@ export const commify = (
...
 };
 
-export const formatFiatValue = (value: number | string, options?: { skipCents?: boolean }): string => {
-  const formatedValue = commify(value, options);
+export const formatFiatValue = (
+  value: number | string,
+  options?: { skipCents?: boolean, isFormatAmount?: boolean },
+): string => {
+  const num = new BigNumber(value).toFixed(2);
+  const formatedValue = options?.skipCents || options?.isFormatAmount ? commify(value, options) : formatBigAmount(num);
```
poocart: I'm not sure I follow the options :( 
1. `skipCents` and `isFormatAmount` according to ternary check does exactly the same so why keep both?
2. Previously `commify` was not taking `skipCents` into credit, but now it does?

divyaBacancy: @poocart I have added isFormatAmount( name changed to shoudCommify) to prevent the formatting in Detail model and review screen and there skipCents are not true. As per @paragramm in Detail and review screen amount should not be formatted as K,M,B.

poocart: Sorry Divya, I'm getting little confused here looking at this, I wonder how we can simplify, maybe `isFormatAmount` could be changed to something more reasonable that would make more sense? Since the method itself is calling for `format`  🤔

src/components/ReviewSummary/TokenReviewSummary.js
```diff
@@ -61,7 +61,8 @@ export const TokenReviewSummaryComponent = ({
 
   if (asset) {
     assetIcon = { uri: `${getEnv().SDK_PROVIDER}/${asset.iconUrl}?size=3` };
-    fiatAmount = fiatAmount || getFormattedRate(rates, amount, asset.symbol, baseFiatCurrency || defaultFiatCurrency);
+    fiatAmount =
+      fiatAmount || getFormattedRate(rates, amount, asset.symbol, baseFiatCurrency || defaultFiatCurrency, true);
```
poocart, line 65: I know it wasn't in this PR scope, but double different ternary here in single line little confusing, maybe this is easier, wdyt?
```suggestion
    if (!fiatAmount) {
      fiatAmount = getFormattedRate(rates, amount, asset.symbol, baseFiatCurrency || defaultFiatCurrency, true);
    }
```

poocart (approved):
LGTM

src/utils/common.js
poocart: Thanks

src/components/ReviewSummary/TokenReviewSummary.js
divyaBacancy: @poocart It make sense. I will do the changes.

PR #2584, opened 2021-05-01
PIL-1009: New Dashboard UI: requested changes

mdjastrzebski:
Scope:
- [x] Asset categories with nested chains
  - [x] Archanova: omit nested chains
  - [x] Deploy status for not deployed Ethereum Mainnet 
- [x] Charts: swipable pie charts: assets & chains
- [x] Remove side chain toggle 
- [x] Updated floating buttons (with 4 buttons support)


src/screens/Home/AssetsSection.js
```diff
@@ -19,132 +19,175 @@
...
-          <HomeListItem
-            key={`${chain}-services`}
-            title={t('services')}
-            iconName="info"
-            onPress={() => navigation.navigate(SERVICES_FLOW)}
-          />
-        )}
+      <React.Fragment key={`${category}-fragment`}>
+        <CategoryListItem
+          key={`${category}`}
```
poocart, line 107:
```suggestion
          key={category}
```

src/screens/Home/BalanceSection.js
```diff
@@ -35,37 +37,36 @@ import { formatFiatValue, formatFiatChangeExtended } from 'utils/format';
...
-import type { Balance } from 'models/Home';
-
 // Local
 import SpecialButton from './components/SpecialButton';
 
 
 type Props = {|
-  balance: Balance,
+  balanceInFiat: BigNumber,
+  changeInFiat?: ?BigNumber,
```
poocart, line 46: I assume it can either be provided `maybe` BigNumber or not provided at all? So either `changeInFiat: ?BigNumber` or `changeInFiat?: BigNumber`, wdyt?

```suggestion
  changeInFiat: ?BigNumber,
```

poocart (approved):
LGTM, well done! 🤩 

One note would be that maybe var naming of `categoryBalances` for type `CategoryBalance` little misleading on plural vs singular. Maybe here we can adopt something like for chains type `CategoryBalancesPerChain` like `BalancePerCategory`, wdyt?

src/screens/Home/BalanceSection.js
mdjastrzebski: `prop?: BigNumber` - means that the prop is optional and can be BigNumber or undefined
`prop: ?BigNumber` - means that the prop must be present & can be BigNumber or null or undefined.
`prop?: ?BigNumber` - means that the prop is option and can be BigNumber or null or undefined.

The third case is what we want to support atm, so we need both of these `?`.

mdjastrzebski:
I've updated `CategoryBalance` type name to `CategoryBalances`. So the problem with `BalancePerCategory` is that we also would need `BalancePerCategoryPerChain` which IMO gets quite hard to read.

PR #2588, opened 2021-05-04
PIL-1011 - Smart Wallet Activation Crash

divyaBacancy:
https://user-images.githubusercontent.com/66996635/117031357-20738d80-ad1e-11eb-87d1-430e76bdf646.mov


IAmKio (approved):
We'll revisit any issues after.

PR #2596, opened 2021-05-12
PIL-1013: Wallet Connect home screen (vNext)

mdjastrzebski:
Scope:
- [x] New Wallet Connect home screen
- [x] Fetching and parsing Prismic data
- [x] Show storyly stories
- [x] Support scenario of not being deployed on mainnet (banner + disabled apps)
- [x] Floating connect button
- [x] Floating connect apps button 

Screenshots:
image
image


src/components/modern/Layout.js
```diff
@@ -91,6 +91,21 @@ export const ColumnRight: React.ComponentType<ViewProps> = styled.View`
...
 
+type CenterProps = {|
+  children: React.Node,
+  flex?: number,
+  height?: string | number,
+  weight?: string | number,
+|};
+
+export const Center: React.ComponentType<CenterProps> = styled.View`
+  ${({ flex }) => (flex != null ? `flex: ${flex}` : '')};
```
poocart, line 102:
```suggestion
  ${({ flex }) => flex != null && `flex: ${flex};`};
```

src/components/modern/TabBar.js
```diff
@@ -0,0 +1,96 @@
...
+              <Underline />
+            </TabContainer>
+          </TouchableWithoutFeedback>
+        ) : (
+          <TouchableWithoutFeedback key={key} onPress={() => onActiveTabChange(key)}>
+            <TabContainer>
+              <TabTitle>{title}</TabTitle>
+            </TabContainer>
+          </TouchableWithoutFeedback>
+        );
```
poocart, line 62: Not crucial, but I think you will agree that ternary with components never reads nicely, maybe put first first `activeTab === key` into if and then after it return the fallback component?

src/models/Chain.js
```diff
@@ -33,3 +33,11 @@ export type ChainRecord<T> = {|
   xdai?: T,
   ethereum?: T,
 |};
+
+// Based on: https://chainid.network/
+export const CHAIN_ID = {
+  ETHEREUM: '1',
```
poocart, line 39: Chain ID is always number type, maybe we can change it here too, wdyt?

src/screens/WalletConnect/Home/WalletConnectHome.js
```diff
@@ -0,0 +1,230 @@
...
+      </ListHeader>
+    );
+  };
+
+  // Note: in order to achieve multicolumn layout, we group n normal items into one list row item.
+  const renderListRow = (items: WalletConnectCmsApp[]) => {
+    return (
+      <ListRow>{items.map(renderItem)}</ListRow>
+    );
+  };
```
poocart, line 92:
```suggestion
  const renderListRow = (items: WalletConnectCmsApp[]) =><ListRow>{items.map(renderItem)}</ListRow>;
```

src/screens/WalletConnect/Home/WalletConnectHome.js
```diff
@@ -0,0 +1,230 @@
...
+          contentContainerStyle={{ paddingBottom: safeArea.bottom + FloatingButtons.SCROLL_VIEW_BOTTOM_INSET }}
+        />
+      ) : (
+        <>
+          {renderListHeader()}
+          <Center flex={1}>
+            <Spinner />
+          </Center>
+        </>
+      )}
```
poocart, line 130: Maybe change this to:

{isFetching && ...}
{!isFetching && ...}

Totally optional tho, but just I think that ternary looks good only if it's single operator, but if there's more conditional rendering inside ternary then it can become a mess to read, wdyt?

src/screens/WalletConnect/Home/components/ConnectFloatingButton.js
```diff
@@ -0,0 +1,84 @@
...
+  };
+
+  const validateUri = (uri: string): boolean => {
+    return uri.startsWith('wc:') || uri.startsWith('pillarwallet:');
+  };
+
+  const handleUri = (uri: string) => {
+    if (uri.startsWith('wc:')) {
+      connectToConnector(uri);
+    } else if (uri.startsWith('pillarwallet:')) {
```
poocart, line 68: I think scanning result will never ran into deep link flow so it can be removed.

src/services/prismic.js
```diff
@@ -18,7 +18,29 @@
...
+import * as Prismic from '@prismicio/client';
 import { getEnv } from 'configs/envConfig';
 
-export default Prismic.client(getEnv().PRISMIC_ENDPOINT_URL, { accessToken: getEnv().PRISMIC_TOKEN });
+export const DOCUMENT_TYPE = 'document.type';
+
+const prismicClient = Prismic.client(getEnv().PRISMIC_ENDPOINT_URL, { accessToken: getEnv().PRISMIC_TOKEN });
+
+export default prismicClient;
+type QueryOptions = {|
```
poocart, line 29:
```suggestion

type QueryOptions = {|
```

src/components/HistoryEventDetails/layouts/WalletEventDetails.js
```diff
@@ -103,7 +97,7 @@ function WalletEventDetails({ event }: Props) {
         <Text variant="large">{t('label.created')}</Text>
         <Spacing h={spacing.extraLarge} />
 
-        {isActivated ? (
+        {isDeployed ? (
```
poocart, line 100: I'm not sure it should be checked for Etherspot in this place? The "activation" is only meant for Archanova and for Etherspot we will have deployment intro screen with no actual deployment (per our last call with Kieran). So until we don't have that implemented I think it's best not to let users enter that screen from Etherspot since it's meant for Archanova deployment only.

Left the same comment on https://github.com/pillarwallet/pillarwallet/pull/2622

poocart (approved):
Mostly great! 👍  one major thing to discuss re wallet deployment UI incentives

mdjastrzebski:
Re wallet deployment incentives: atm they are implemented as in designs, so we should address these after Dmitry updates the designs to embrace the new deployment approach. @poocart, wdyt?

src/screens/WalletConnect/Home/components/ConnectFloatingButton.js
mdjastrzebski: Removed deep link here

src/components/modern/TabBar.js
mdjastrzebski: Extracted into renderItem function

src/components/HistoryEventDetails/layouts/WalletEventDetails.js
mdjastrzebski: I'll remove this check them as the new history UI is intended only for Etherspot

src/models/Chain.js
mdjastrzebski: Changed

PR #2600, opened 2021-05-13
PIL-1024: exchange allowances per account

poocart:
https://linear.app/pillarproject/issue/PIL-1024/exchange-allowances

Adds exchange allowances per account to support recently adde Etherspot account.


src/models/Offer.js
```diff
@@ -70,18 +62,22 @@ export type OfferOrder = {
...
 export type Allowance = {
   provider: string,
   fromAssetCode: string,
   toAssetCode: string,
   transactionHash: string,
   enabled?: boolean,
-}
+};
+
+export type Allowances = {
```
mdjastrzebski, line 81: Consider renaming to `AccountAllowances` or `AllowancesPerAccount` for readability.

src/actions/exchangeActions.js
```diff
@@ -328,29 +370,31 @@ export const checkEnableExchangeAllowanceTransactionsAction = () => {
...
-            emoji: 'handshake',
-            autoClose: true,
-          });
-          dispatch(enableExchangeAllowanceByHashAction(transactionHash));
-        }
-      });
+    Object.keys(exchangeAllowances).map((accountId) =>
+      exchangeAllowances[accountId]
+        .filter(({ enabled }) => !enabled)
+        .map(({ transactionHash, fromAssetCode, toAssetCode }) => {  // eslint-disable-line
```
mdjastrzebski, line 376: Looks like this should be 'forEach'

```suggestion
        .forEach(({ transactionHash, fromAssetCode, toAssetCode }) => { 
```

src/actions/exchangeActions.js
```diff
@@ -263,49 +270,84 @@ export const addExchangeAllowanceIfNeededAction = ({
...
-    const { exchange: { data: { allowances: _allowances = [] } } } = getState();
-    const allowance = _allowances.find(
+export const enableExchangeAllowanceByHashAction = (transactionHash: string, senderAccountId: string) => {
+  return (dispatch: Dispatch, getState: GetState) => {
+    const allowances = allAccountsAllowancesSelector(getState());
+
+    const combinedAllowances = Object.keys(allowances).reduce((combined, accountId) => [
+      ...combined,
+      ...allowances[accountId],
+    ], []);
```
mdjastrzebski, line 315: This should be the same, but more idiomatic.
```suggestion
    const combinedAllowances = Object.keys(allowances).flatMap((accountId) => allowances[accountId]);
```

src/actions/exchangeActions.js
```diff
@@ -263,49 +270,84 @@ export const addExchangeAllowanceIfNeededAction = ({
...
     });
-    const allowances = _allowances
-      .filter(
-        ({ transactionHash: _transactionHash }) => _transactionHash !== transactionHash,
-      );
-    allowances.push(updatedAllowance);
-    dispatch(saveDbAction('exchangeAllowances', { allowances }, true));
+
+    const updatedAllowances = Object.keys(allowances).reduce((updated, allowancesAccountId) => {
+      if (senderAccountId !== allowancesAccountId) return updated;
```
mdjastrzebski, line 334: This part seems a bit weird. So for all entries in `allowances` map we are interested only in allowanes for `senderAccountId`. IMO it looks like there is no need to use `reduce` then, but just get the `allowances[senderAccountId]` and do the same thing as in `reduce` body.

WDYT?

src/actions/exchangeActions.js
```diff
@@ -263,49 +270,84 @@ export const addExchangeAllowanceIfNeededAction = ({
...
     const updatedAllowance = {
       ...allowance,
       enabled: true,
     };
+
     dispatch({
-      type: UPDATE_EXCHANGE_ALLOWANCE,
-      payload: updatedAllowance,
+      type: UPDATE_ACCOUNT_EXCHANGE_ALLOWANCE,
+      payload: { accountId: senderAccountId, allowance: updatedAllowance },
```
mdjastrzebski, line 330: Seems like we could be duplicating given allowance entry from one account to another, is that the intent?

mdjastrzebski (approved):
Overall looks good, but there are some minor details that require attention.

src/actions/exchangeActions.js
poocart: Yes. Allowance is set per each sender account.

src/services/dataMigration/exchangeAllowances.js
```diff
@@ -0,0 +1,39 @@
...
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import { isArray } from 'lodash';
+import { getAccountId, isArchanovaAccount } from 'utils/accounts';
+
+export default function (storageData: Object) {
```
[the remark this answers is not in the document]
poocart: I think for the matter on this PR let's leave as it is.

src/actions/exchangeActions.js
poocart: Old code, didn't bother to fix as it should have been quick fix, but will change, thanks 👍

src/reducers/exchangeReducer.js
```diff
@@ -141,32 +150,31 @@ export default function exchangeReducer(
...
-          allowances: [
+          allowances: {
             ...state.data.allowances,
-            action.payload,
-          ],
+            [accountId]: [...accountAllowances, allowance],
+          },
           hasNotification: true,
         },
       };
```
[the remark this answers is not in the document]
poocart: Technically we could, but we have lint rule that forbids redeclaring same name vars under single switch.

src/actions/exchangeActions.js
poocart: You're absolutely right, this ticket is little rush of moving existing pieces, changed, thanks, good catch 👍

PR #2613, opened 2021-05-17
fix send token flow mixed param passed to send token component

poocart:
Includes:
- Fixed issue where account address is provided as token address on token send flow from Asset screen therefore failing to provide estimate.
- Additionally fixed issue where asset icon is not shown when navigated through send flow from Asset screen.


mdjastrzebski (approved):
LG

IAmKio (approved):

PR #2617, opened 2021-05-17
fixed amount parse using ethers big number

poocart:
Fixed issue where send amount is parsed multiple ways into Ethers BigNumber. Changed to use single method general Ethers parse util instead.


IAmKio (approved):
LGTM

PR #2621, opened 2021-05-17
PIL-1030: Expose existing account selector (vNext)

mdjastrzebski:
Scope:
- [x] Expose existing account selector on new home screen
- [x] Do not display selector icon if user has only one account

Screenshots:
image


IAmKio (approved):
Great work! 🎉

PR #2622, opened 2021-05-17
Feature/v next etherspot account deployed check

mdjastrzebski:
Scope:
- [x] Check for etherspot account deployment
- [x] Check for active account deployment (etherspot + archanova)
- [x] Apply in new dashboard UI, new history UI, wallet connect,

### DIFF
--- a/src/actions/walletConnectActions.js
+++ b/src/actions/walletConnectActions.js
@@ -162,7 +162,8 @@ export const approveWalletConnectConnectorRequestAction = (peerId: string) => {
       return;
     }
 
-    if (!isActiveAccountDeployedOnEthereumSelector(getState())) {
+    const requiresSmartWalletDeployment = !isActiveAccountDeployedOnEthereumSelector(getState());
+    if (requiresSmartWalletDeployment) {
       Toast.show({
         message: t('toast.walletConnectSmartWalletNotActive'),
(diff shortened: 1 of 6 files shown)


src/components/HistoryEventDetails/layouts/WalletEventDetails.js
```diff
@@ -103,7 +97,7 @@ function WalletEventDetails({ event }: Props) {
         <Text variant="large">{t('label.created')}</Text>
         <Spacing h={spacing.extraLarge} />
 
-        {isActivated ? (
+        {isDeployed ? (
```
poocart, line 100: I'm not sure it should be checked for Etherspot in this place? The "activation" is only meant for Archanova and for Etherspot we will have deployment intro screen with no actual deployment (per our last call with Kieran). So until we don't have that implemented I think it's best not to let users enter that screen from Etherspot since it's meant for Archanova deploymeny only.

poocart (approved):
LGTM, left one bigger note to pay attention to 👀

src/components/HistoryEventDetails/layouts/WalletEventDetails.js
mdjastrzebski: Removed that, as new history UI is only for Etherspot.

PR #2623, opened 2021-05-17
PIL-1017: Wallet Connect: transaction and signature request modals.

mdjastrzebski:
Scope:
- [x] Call request base modal & workaround for using it as navigation screen
- [x] Modal contents for: transaction request, signature request, unsupported request


src/components/modern/Layout.js
```diff
@@ -91,6 +91,21 @@ export const ColumnRight: React.ComponentType<ViewProps> = styled.View`
...
 
+type CenterProps = {|
+  children: React.Node,
+  flex?: number,
+  height?: string | number,
+  weight?: string | number,
+|};
+
+export const Center: React.ComponentType<CenterProps> = styled.View`
+  ${({ flex }) => (flex != null ? `flex: ${flex}` : '')};
```
poocart, line 102:
```suggestion
  ${({ flex }) => flex != null && `flex: ${flex}`};
```

src/components/modern/TabBar.js
```diff
@@ -0,0 +1,96 @@
...
+            </TabContainer>
+          </TouchableWithoutFeedback>
+        ) : (
+          <TouchableWithoutFeedback key={key} onPress={() => onActiveTabChange(key)}>
+            <TabContainer>
+              <TabTitle>{title}</TabTitle>
+            </TabContainer>
+          </TouchableWithoutFeedback>
+        );
+      })}
```
poocart, line 63: Not crucial, but I think you will agree that ternary with components never reads nicely, maybe put first first `activeTab === key` into `if` and then after it return the fallback component?

src/models/Chain.js
```diff
@@ -33,3 +33,11 @@ export type ChainRecord<T> = {|
   xdai?: T,
   ethereum?: T,
 |};
+
+// Based on: https://chainid.network/
+export const CHAIN_ID = {
```
poocart, line 38: ID is always number type, maybe we can change it here too, wdyt?

src/screens/WalletConnect/Home/WalletConnectHome.js
```diff
@@ -0,0 +1,239 @@
...
+      </ListHeader>
+    );
+  };
+
+  // Note: in order to achieve multicolumn layout, we group n normal items into one list row item.
+  const renderListRow = (items: WalletConnectApp[]) => {
+    return (
+      <ListRow>{items.map(renderItem)}</ListRow>
+    );
+  };
```
poocart, line 105:
```suggestion
  const renderListRow = (items: WalletConnectApp[]) => <ListRow>{items.map(renderItem)}</ListRow>)
```

src/screens/WalletConnect/Home/WalletConnectHome.js
```diff
@@ -0,0 +1,239 @@
...
+          contentContainerStyle={{ paddingBottom: safeArea.bottom + FloatingButtons.SCROLL_VIEW_BOTTOM_INSET }}
+        />
+      ) : (
+        <>
+          {renderListHeader()}
+          <Center flex={1}>
+            <Spinner />
+          </Center>
+        </>
+      )}
```
poocart, line 144: Maybe change this to:

{isFetching && ...}
{!isFetching && ...}

Totally optional tho, but just I think that ternary works only if it's single operator, but if there's more conditional rendering inside ternary then it can become a mess to read, wdyt?

src/screens/WalletConnect/Home/components/ConnectFloatingButton.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    Modal.open(() => <QRCodeScanner validator={validateUri} onRead={handleUri} />);
+  };
+
+  const validateUri = (uri: string): boolean => {
+    return uri.startsWith('wc:') || uri.startsWith('pillarwallet:');
+  };
+
+  const handleUri = (uri: string) => {
```
poocart, line 61: I think scanning result will never ran into deep link flow so it can be removed.

src/screens/WalletConnect/Home/utils.js
```diff
@@ -0,0 +1,39 @@
...
+ * If chain is nil it returns all passed apps.
+ */
+export function filterAppsByChain(apps: WalletConnectApp[], chain: ?Chain): WalletConnectApp[] {
+  if (!chain) return apps;
+  return apps.filter(app => app.chains.includes(chain));
+}
+
+// NOTE: replace by proper useWalletConnect hook when available
+// Mock data
+export function useWalletConnect() {
```
poocart, line 35: I think this is already in place?

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
```diff
@@ -0,0 +1,199 @@
...
+    decimals,
+    txFeeInWei: feeInWei,
+    gasToken: feeInfo?.gasToken,
+  });
+
+  const { estimateCallRequestTransaction } = useWalletConnect();
+
+  React.useEffect(() => {
+    estimateCallRequestTransaction(request);
+  }, [request, estimateCallRequestTransaction]);
```
IAmKio, line 151:
```suggestion
  React.useEffect(() => estimateCallRequestTransaction(request), [request, estimateCallRequestTransaction]);
```

Possibility?

IAmKio (approved):
Great stuff!

PR #2627, opened 2021-05-18
onboarding new suername crash fix, mnemonic input back flow fix

poocart:
Includes:
- Fix for non-string username added as default text input value.
- Fix for mnemonic import flow when user presses back and imports new mnemonic.


IAmKio (approved):
🙏🏾👌🏾

PR #2639, opened 2021-05-21
PIL-1027: Feature/vnext real data

poocart:
https://linear.app/pillarproject/issue/PIL-1027/wire-up-live-data-with-v2-ui

Part 1 - balances

"This will be simple PR" – last famous words.

Includes:
- Totals using Zapper API.
- Collectible send flow fixes.
- Etherspot account deployed status fix.
- Removed mock data.
- Other small fixes.

Note: this PR does not contain any protocol items returned (will go under `totals` reducer as `items` or smth) by Zapper and prices are in USD, this will be addressed in separate PR.


src/constants/zapperServiceConstants.js
```diff
@@ -0,0 +1,93 @@
...
+  SUSHISWAP: 'sushiswap',
+  SWERVE: 'swerve',
+  SYNTHETIX: 'synthetix',
+  TOKENSETS: 'tokensets',
+  TOKENS: 'tokens',
+  UNISWAP: 'uniswap',
+  UNISWAP_V2: 'uniswap-v2',
+  UNISWAP_V3: 'uniswap-v3',
+  UNIT: 'unit',
+  VALUE: 'VALUE',
```
IAmKio, line 87: Sorry did i caps this? I think it's supposed to be lower case...

```suggestion
  VALUE: 'value',
```

src/screens/Accounts/Accounts.js
```diff
@@ -81,34 +82,30 @@ type ListItem = {|
...
   keyBasedWalletHasPositiveBalance,
+  accountsTotals,
 }: Props) => {
   const navigation = useNavigation();
   const theme = useTheme();
+  const fiatCurrency = useFiatCurrency();
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  useEffect(() => { fetchAllAccountsBalances(); }, []);
+  useEffect(() => { fetchAllAccountsTotals(); }, []);
```
IAmKio, line 105:
```suggestion
  useEffect(() => fetchAllAccountsTotals(), []);
```

Maybe?

src/screens/Assets/collectibles/utils.js
```diff
@@ -17,17 +17,29 @@
     with this program; if not, write to the Free Software Foundation, Inc.,
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
-
+import { COLLECTIBLES } from 'constants/assetsConstants';
 import type { CollectibleNavigationParams } from 'screens/Collectible';
 import type { CollectibleItem } from './selectors';
```
IAmKio, line 22:
```suggestion
// Constants
import { COLLECTIBLES } from 'constants/assetsConstants';

// Screens
import type { CollectibleNavigationParams } from 'screens/Collectible';

// UI
import type { CollectibleItem } from './selectors';
```

src/screens/Assets/deposits/selectors.js
```diff
@@ -22,19 +22,20 @@ import { BigNumber } from 'bignumber.js';
...
-  const value = useRootSelector(depositsBalanceSelector);
+  const value = useRootSelector(depositsTotalBalanceSelector);
   return { value };
 }
 
-export function useDepositsChainBalances(): ChainRecord<BigNumber> {
-  return { ethereum: BigNumber(10) };
+export function useDepositsChainBalances(): ChainBalances {
+  return useRootSelector(depositsTotalBalanceByChainsSelector);
 }
```
IAmKio, line 39:
```suggestion
export function useDepositsChainBalances(): ChainBalances => useRootSelector(depositsTotalBalanceByChainsSelector);
```

Not essential.

src/screens/Assets/investments/selectors.js
```diff
@@ -22,19 +22,23 @@ import { BigNumber } from 'bignumber.js';
...
-  const value = useRootSelector(investmentsBalanceSelector);
+  const value = useRootSelector(investmentsTotalBalanceSelector);
   return { value };
 }
 
-export function useInvestmentsChainBalances(): ChainRecord<BigNumber> {
-  return { ethereum: BigNumber(10) };
+export function useInvestmentsChainBalances(): ChainBalances {
+  return useRootSelector(investmentsTotalBalanceByChainsSelector);
 }
```
IAmKio, line 42:
```suggestion
export function useInvestmentsChainBalances(): ChainBalances => useRootSelector(investmentsTotalBalanceByChainsSelector);
```

src/screens/Assets/utils.js
```diff
@@ -50,11 +46,11 @@ export function buildAssetDataNavigationParam(asset: Asset, { accountAddress }:
     id: symbol,
     name: name || symbol,
     token: symbol,
-    address: accountAddress,
```
IAmKio, line 53: is it better if we rename this to `accountAddress`? I feel that the term `address` itself is a bit ambiguous in this context...

src/services/zapper.js
```diff
@@ -0,0 +1,228 @@
...
+import type { Chain } from 'models/Chain';
+
+// does not change between envs
+const ZAPPER_CONFIG = {
+  API_URL: 'https://api.zapper.fi/v1',
+  API_KEY: '96e0cc51-a62e-42ca-acee-910ea7d2a241', // public
+};
+
+const requestConfig = {
+  timeout: 10000,
```
IAmKio, line 40: Could we remove this timeout altogether, just incase we have a scenario where network conditions or Zapper server load slows down the response time. I'm always happy to wait _a bit_ longer to get my data - eventually.

src/services/zapper.js
```diff
@@ -0,0 +1,228 @@
...
+import { CHAIN } from 'models/Chain';
+import { PROTOCOLS } from 'constants/zapperServiceConstants';
+
+// types
+import type { Chain } from 'models/Chain';
+
+// does not change between envs
+const ZAPPER_CONFIG = {
+  API_URL: 'https://api.zapper.fi/v1',
+  API_KEY: '96e0cc51-a62e-42ca-acee-910ea7d2a241', // public
```
IAmKio, line 36: Could we move these values to `envConfig.js`?

src/services/zapper.js
```diff
@@ -0,0 +1,228 @@
...
+
+export const getZapperProtocolBalanceOnNetwork = async (
+  addresses: string[],
+  protocol: string,
+  network: string,
+) => {
+  if (!supportedProtocols.includes(protocol)) return null;
+
+  try {
+    const result = await httpRequest.get(
```
IAmKio, line 173: Do we need the try / catch here? Could we not use the .catch method of the promise `httpRequest.get`?

src/services/zapper.js
```diff
@@ -0,0 +1,228 @@
...
+type AvailableChainProtocol = {
+  chain: string,
+  zapperNetworkId: string,
+  zapperProtocolIds: string[],
+};
+
+export const getZapperAvailableChainProtocols = async (
+  addresses: string[],
+): Promise<?AvailableChainProtocol[]> => {
+  try {
```
IAmKio, line 202: Just a note re: the try / catch mentioned above.

IAmKio (changes requested):
Absolutely fantastic work 🎉 the app is **ALIVE** and it feels **great**! ❤️ I just want to note that whilst i have requested changes - at this moment in time i'm not precious about them due to our Monday deadline. Anything that we don't address in terms of code quality and other minor things, we can defer till after. I'll leave it to your judgement mate 👍

src/selectors/balances.js
```diff
@@ -21,39 +21,54 @@ import { createSelector } from 'reselect';
...
+  activeAccountSelector,
 } from './selectors';
 import { availableStakeSelector } from './paymentNetwork';
 
 
+const sumBalancesPerChainChain = (
+  accountTotals: ?CategoryBalancesPerChain,
+  category: string,
+): ChainBalances => {
+  const chains = (Object.keys(accountTotals || {}): any);
```
mdjastrzebski, line 53: Pls try to avoid any here, typing should work with `CategoryBalancesPerChain`
```suggestion
if (!accountTotals) return {};
const chains = Object.keys(accountTotals);
```

src/selectors/balances.js
```diff
@@ -21,39 +21,54 @@ import { createSelector } from 'reselect';
...
-  poolTogetherStatsSelector,
-  sablierSelector,
-  rariSelector,
-  liquidityPoolsSelector,
+  activeAccountSelector,
 } from './selectors';
 import { availableStakeSelector } from './paymentNetwork';
 
 
+const sumBalancesPerChainChain = (
```
mdjastrzebski, line 49: This probably should go to `utils/balances` or some other `utils`.

mdjastrzebski, line 49: It seems you're not summing anything here.
```suggestion
const getChainBalancesForCategory = (
```

mdjastrzebski: I've got pretty similar function implementations in `screens/Home/utils` you might want to moved them to `utils/balances` and re-use them.

mdjastrzebski (approved):
Awesome stuff mate! 💯 Great to see all pieces coming together! 

Some points to consider:
* I've to a bunch of balance calculation functions in `screen/Home/utils` you mighe want to move them to `utils/balances` and re-use them
* Consider using more specific `balances` naming instead of generic `totals`
* Zapper nested query seems huge (like 60+ api calls in one go), does it work fine in decent amount of time? What happens if one of the queries fails?

src/components/HistoryEventDetails/layouts/TokenTransactionEventDetails.js
```diff
@@ -126,7 +126,7 @@ function TokenTransactionEventDetails({ event }: Props) {
         <TransactionStatusText status={event.status} color={colors.basic030} variant="medium" />
         <Spacing h={spacing.extraLarge} />
 
-        <FeeLabel value={event.fee.value} symbol={event.fee.symbol} mode="actual" />
+        {!!event?.fee && <FeeLabel value={event.fee.value} symbol={event.fee.symbol} mode="actual" />}
```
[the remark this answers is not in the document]
poocart: Wdym? 👀

src/screens/Accounts/Accounts.js
poocart: No, `useEffect` return is reserved for `cleanup` before re-render happens to unsubscribe from prev. instances and etc. (ref – https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1).

src/actions/appActions.js
```diff
@@ -194,6 +195,9 @@ export const initAppAndRedirectAction = () => {
       const { contacts = [] } = get(storageData, 'localContacts', []);
       dispatch({ type: SET_CONTACTS, payload: contacts });
 
+      const { balances: totalBalances = {} } = get(storageData, 'totalBalances', {});
```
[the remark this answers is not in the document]
poocart: Just wanted to keep in the code as other calls, but you know what, maybe this is wrong because I keep copying old pattern every time I add so maybe it's good to add this, thanks

src/screens/Assets/utils.js
poocart: It's not needed here at all, I guess it appeared here by mistake because 🤷

src/actions/assetsActions.js
```diff
@@ -371,6 +385,85 @@ export const fetchAccountAssetsBalancesAction = (account: Account) => {
...
+    const availableChainProtocols = await getZapperAvailableChainProtocols(accountsAddresses);
+    if (!availableChainProtocols) {
+      reportErrorLog('fetchAllAccountsTotalsAction failed: no availableChainProtocols', { accountsAddresses });
+      return;
+    }
+
+    try {
+      await Promise.all(availableChainProtocols.map(async ({ chain, zapperProtocolIds, zapperNetworkId }) => {
+        const chainProtocolsBalances = await Promise.all(zapperProtocolIds.map(async (zapperProtocolId) => {
+          const protocolBalancesByAccounts = await getZapperProtocolBalanceOnNetwork(
```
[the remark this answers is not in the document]
poocart: Yes mate, we're aware, thanks for concerns! This will essentially move either to our serverless or Etherspot back-end, but given time window we have until private beta this is the only solution.

src/services/zapper.js
```diff
@@ -0,0 +1,228 @@
...
+    case 'binance-smart-chain': return CHAIN.BINANCE;
+    case 'xdai': return CHAIN.XDAI;
+    default: return null;
+  }
+};
+
+
+const walletProtocols = [
+  PROTOCOLS.TOKENS,
+  PROTOCOLS.BITCOIN,
```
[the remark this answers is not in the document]
poocart: Note sure TBH, @IAmKio added protocol values from Zapper Swagger. However, we also had a discussion earlier today and agreed that wallet balances will be going from Etherspot only because Zapper does not support xDai balances and Etherspot can be more trusted on actual balances when it comes to sending right asset amounts. We leave as it is now and will revert back to Etherspot in separate PR.

src/components/HistoryEventDetails/layouts/WalletEventDetails.js
```diff
@@ -106,7 +106,7 @@ function WalletEventDetails({ event }: Props) {
         <TransactionStatusText status={event.status} color={colors.basic030} variant="medium" />
         <Spacing h={spacing.extraLarge} />
 
-        <FeeLabel value={event.fee.value} symbol={event.fee.symbol} mode="actual" />
+        {!!event?.fee && <FeeLabel value={event.fee.value} symbol={event.fee.symbol} mode="actual" />}
```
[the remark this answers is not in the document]
poocart: IMO the component may become more error prone while there's no strong argument, but I'd rather add fees checks all over than do a no fees no display fallback at the component level.

src/services/zapper.js
```diff
@@ -0,0 +1,228 @@
...
+
+const mapZapperNetworkIdToChain = (network: string): ?Chain => {
+  switch (network) {
+    case 'ethereum': return CHAIN.ETHEREUM;
+    case 'polygon': return CHAIN.POLYGON;
+    case 'binance-smart-chain': return CHAIN.BINANCE;
+    case 'xdai': return CHAIN.XDAI;
+    default: return null;
+  }
+};
```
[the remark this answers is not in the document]
poocart: 👍

src/services/zapper.js
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    return result.data;
+  } catch (error) {
+    reportErrorLog('getZapperProtocolBalanceOnNetwork: API request error', { error });
+    return null;
+  }
+};
+
+type AvailableChainProtocol = {
```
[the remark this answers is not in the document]
poocart: IDK, it's hard to name this result, named to `ZapperNetworkWithProtocols`, hope it makes most sense.

src/actions/transactionEstimateActions.js
```diff
@@ -106,22 +106,32 @@ export const estimateTransactionsAction = (transactionsToEstimate: TransactionTo
...
-    )));
+    let transactions;
+    try {
+      transactions = await Promise.all(transactionsToEstimate.map(({
+        to,
+        data,
+        value,
+        assetData,
+      }) => buildEthereumTransaction(
+        to,
```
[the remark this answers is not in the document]
poocart: Thanks, imo it won't change much in this case, just it would require more code to add in order to construct an object here. Will leave as it is, hope it makes sense 👍

src/models/Home.js
```diff
@@ -34,3 +34,7 @@ export type CategoryBalances = {|
 export type ChainBalances = ChainRecord<BigNumber>;
 
 export type CollectibleCountPerChain = ChainRecord<number>;
+
+export type AccountsTotalBalances = {
```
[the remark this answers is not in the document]
poocart: Changed to `ChainBalancesPerAccount` maybe it would bring more sense as Balances confuses since we have Balances for Assets and Balances for Category (+ Category per Chain).

src/screens/Accounts/Accounts.js
```diff
@@ -164,18 +161,17 @@ const AccountsScreen = ({
...
       const { id, isActive, type } = account;
-      const accountBalances: Balances = balances[id];
       const isActiveWallet = !!isActive && isEthereumActive;
-      let walletBalance;
-      if (accountBalances) {
-        const thisAccountBalance = getTotalBalanceInFiat(accountBalances, rates, fiatCurrency);
-        walletBalance = formatFiat(thisAccountBalance, baseFiatCurrency);
-      }
+
+      const totalCategoryBalances = getTotalCategoryBalances(accountsTotals[id] ?? {});
```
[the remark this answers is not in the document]
poocart: Not really at this piece, here it loops through accounts and gets each balance from all accounts totals.

src/screens/History/HistoryListEtherspot.js
```diff
@@ -36,126 +54,124 @@ function HistoryListEtherspot() {
...
 
-// Return mock data for available cell types
+function parseFee(
+  feeWithGasToken: ?FeeWithGasToken,
+  gasUsed: ?number,
+  gasPrice: ?number,
+): ?TokenValue {
+  if (feeWithGasToken?.feeInWei) {
+    return {
+      value: feeWithGasToken.feeInWei,
```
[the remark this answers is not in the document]
poocart: 👍

src/selectors/balances.js
poocart: 👍

src/reducers/totalsReducer.js
```diff
@@ -0,0 +1,106 @@
...
+} from 'constants/totalsConstants';
+
+// utils
+import { BigNumber } from 'utils/common';
+
+// types
+import type { AccountsTotalBalances } from 'models/Home';
+
+
+export type TotalsReducerState = {
```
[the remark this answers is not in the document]
poocart: Let's move this discussion to Slack.

src/constants/totalsConstants.js
```diff
@@ -0,0 +1,24 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+export const SET_FETCHING_TOTALS = 'SET_FETCHING_TOTALS';
```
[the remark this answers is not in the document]
poocart: Let's move this discussion to Slack.

src/selectors/balances.js
poocart: It's not, but will add a fix.

src/selectors/balances.js
poocart: So it's similar but it's not because it returns total cross chain balance for selected category.

IAmKio (approved):

PR #2643, opened 2021-05-21
PIL-1025: Fix/etherspot ens migrator fixes

poocart:
https://linear.app/pillarproject/issue/PIL-1025/ens-migrator-fix


src/components/Button/Button.js
```diff
@@ -176,8 +176,8 @@ const getLabelTopMargin = (props) => {
   return getButtonFontSize(props) * 0.18;
 };
 
-const getButtonOpacity = ({ disabled, isLoading, theme }) => {
-  if (disabled && !isLoading) {
+const getButtonOpacity = ({ disabled, theme }) => {
```
poocart, line 179: This doesn't belong to the scope, but it confuses every time when submit button is loading and is disabled it's still showing regular opacity. Let me know if you have concerns.

src/services/archanova.js
```diff
@@ -120,6 +120,52 @@ export const formatEstimated = (estimated: ?ParsedEstimate) => {
...
+  }
+
+  reportErrorLog('Smart Wallet service estimateAccountTransaction failed', {
+    errorMessage,
+    ...logExtras,
+  });
+
+  throw new Error(errorMessage);
+};
+
```
poocart, line 167:
```suggestion

```

src/selectors/archanova.js
```diff
@@ -89,7 +89,7 @@ export const isEnsMigrationNeededSelector = createSelector(
       archanovaAccount && (!isProdEnv() || getAccountEnsName(archanovaAccount)?.endsWith(getEnsPrefix()));
 
     const isEnsMigrationTransactionAlreadySent = archanovaAccountHistory.some(
-      ({ extra }) => extra?.isENSMigrationToEtherspot,
+      ({ extra }) => extra?.isEnsMigrationToEtherspot,
```
mdjastrzebski, line 92: 👍🏻

src/locales/en/common.json
```diff
@@ -1653,7 +1653,8 @@
     "walletConnectCallRequestRejectFailed": "Failed to reject Call Request.",
     "walletConnectRequestRejected": "Wallet Connect Call Request rejected.",
     "walletConnectRequestApproved": "Wallet Connect Call Request approved.",
-    "ensMigrationCannotProceed": "Sorry, ENS migration cannot proceed. Please contact support."
+    "ensMigrationCannotProceed": "Sorry, ENS migration cannot proceed. Please contact support.",
```
mdjastrzebski, line 1656: Do we need "Please contanct support" part, there is a Toast.show option `supportLink: true` which displays interactive support link button

src/components/Button/Button.js
```diff
@@ -176,8 +176,8 @@ const getLabelTopMargin = (props) => {
...
+  if (disabled) {
```
mdjastrzebski, line 180:
```suggestion
  if (disabled || isLoading) {
```

src/actions/smartWalletActions.js
```diff
@@ -1757,29 +1756,38 @@ export const estimateEnsMigrationFromArchanovaToEtherspotAction = () => {
...
-    const transactionsToEstimate = migratorTransactions.map(({
-      data,
-      to,
-    }) => ({ to, data, value: EthersBigNumber.from(0) }));
+    dispatch(setEstimatingTransactionAction(true));
+
+    let errorMessage;
+    let estimated;
+    try {
+      estimated = await archanovaService.estimateAccountRawTransactions(rawTransactions);
```
IAmKio, line 1764: Are we sure we need try catch here?

src/actions/smartWalletActions.js
```diff
@@ -1757,29 +1756,38 @@ export const estimateEnsMigrationFromArchanovaToEtherspotAction = () => {
...
+    } catch (error) {
+      errorMessage = error?.message;
```
IAmKio, line 1766: Should we report this message?

src/actions/smartWalletActions.js
```diff
@@ -1793,38 +1801,45 @@ export const migrateEnsFromArchanovaToEtherspotAction = (
...
+    try {
+      hash = await archanovaService.sendRawTransactions(rawTransactions);
+    } catch (error) {
+      reportErrorLog('migrateEnsFromArchanovaToEtherspotAction failed', {
+        error,
+        archanovaAccount,
+        activeAccount,
+        rawTransactions,
+      });
+    }
```
IAmKio, line 1814: Are these methods throwing raw errors, is this why we're try catching?

src/locales/en/common.json
```diff
@@ -1653,7 +1653,8 @@
...
+    "ensMigrationFailed": "Sorry, ENS migration failed. Please contact support."
```
IAmKio, line 1657: Thanks for being polite 🙏

src/services/archanova.js
```diff
@@ -120,6 +120,51 @@ export const formatEstimated = (estimated: ?ParsedEstimate) => {
...
+    /**
+     * the return result from Archanova is problematic,
+     * it contains "ethjs-query" part and rest of message is JSON string
+     * this is known issue on Archanova end, but it's not planned to be fixed
+     * return example:
+     * [ethjs-query] while formatting outputs from RPC
+     * '{"value":{"body":"{\"oneMore\":\"escaped\",\"JSON\":\"object\"}"}}'
+     */
+    if (error.message.includes(errorReplaceString)) {
+      try {
```
IAmKio, line 136: Definitely justified use of try/catch, when JSON.parse explodes - it explodes!

IAmKio (approved):
A few minor points but i'm not being picky after the saga that was ENS migrator! Thanks for the effort!

mdjastrzebski (approved):
Looks good! 💯

src/actions/smartWalletActions.js
poocart: Yes, in case estimation fails. This is isolated from regular estimation flow.

src/actions/smartWalletActions.js
poocart: Yes. I guess these questions arise of me trying to adopt a pattern of await without `catch`,  but use `try/catch` instead. Will see how it goes in long term, this is per @mdjastrzebski inisghts.

src/locales/en/common.json
poocart: 🇨🇦

src/services/archanova.js
poocart: FYI it's previous code ported as method to reuse on other new method.

src/locales/en/common.json
poocart: This was per Kieran suggestion, @IAmKio wdyt?

src/components/Button/Button.js
poocart: Thought about this, but it's not the same as then it doesn't give control over button to show it as enabled while loadin. Hope it makes sense to leave it under control of `disabled` and `isLoading` separately which means `disabled` brings opacity down as it should and `isLoading` has it's purpose to show loading but doesn't have any effect on opacity.

mdjastrzebski: makes sense, you can keep the conditions as you changed it

src/locales/en/common.json
IAmKio: You know i'm just wondering if this was there at a time when we didn't have the support link (pre-toast rework)? Because yeah it would read "Please contact support. Contact Support" 😂 so i'm for removing it.

poocart: 🚀

PR #2647, opened 2021-05-22
Fix/convert zapper usd to app fiat

poocart:
Zapper provides values in USD only, but good guys Zapper also provide cool endpoint to convert from USD, this PR includes it and the conversion part when pulling balances from Zapper.

### DIFF
--- a/src/actions/assetsActions.js
+++ b/src/actions/assetsActions.js
@@ -67,13 +67,12 @@ import archanovaService from 'services/archanova';
 import {
   getZapperAvailableChainProtocols,
   getZapperProtocolBalanceOnNetwork,
-  getZapperUSDBasedFiatRates,
+  getZapperFiatRates,
   mapZapperProtocolIdToBalanceCategory,
 } from 'services/zapper';
 
 // utils
 import {
-  addressesEqual,
   getAssetsAsList,
   getRate,
   transformBalancesToObject,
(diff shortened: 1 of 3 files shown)


src/services/zapper.js
```diff
@@ -61,6 +61,10 @@ type ZapperProtocolBalances = {
   }
 };
 
+type ZapperFiatRates = {
```
mdjastrzebski, line 64: Add comment that the rates are in USD terms

src/services/zapper.js
```diff
@@ -61,6 +61,10 @@ type ZapperProtocolBalances = {
   }
```
mdjastrzebski, line 61: ❗ important
Commenting here, because I cannot comment in the non-related code in this file:

Above when defining ZapperAsset we are using `balance: number` to read asset balances. This will potentially result in subtle rounding errors when user would try to use send max later in the app.  According to zapper swagger api they also include `"balanceRaw": "string",` field which seems to be exact value as string.

src/actions/assetsActions.js
```diff
@@ -490,7 +501,12 @@ export const fetchAllAccountsTotalBalancesAction = () => {
               const protocolBalances = balances?.assets;
               if (isEmpty(protocolBalances)) return categoryBalance;
 
-              const protocolBalanceValues = protocolBalances.map(({ balanceUSD }) => wrapBigNumber(balanceUSD ?? 0));
+              let protocolBalanceValues = protocolBalances.map(({ balanceUSD }) => wrapBigNumber(balanceUSD ?? 0));
```
mdjastrzebski, line 504: you can simply use BigNumber constructor here (can be used without `new`)

```suggestion
              let protocolBalanceValues = protocolBalances.map(({ balanceUSD }) => BigNumber(balanceUSD ?? 0));
```

mdjastrzebski (approved):
Looks good man 👏🏻 

❗ I've added one important comment to strictly related to that PR but rather to previous Zapper code. Seems we were using `balance: number` to get asset balances. This can result in rounding errors later in the app, esp. when using feature that depend on exact token value like send max. In the same entity there exist `balanceRaw: string` which is probably the thing that we should be using.

src/services/zapper.js
poocart: Added above method.

PR #2650, opened 2021-05-23
PIL-1027: pull zapper assets

poocart:
https://linear.app/pillarproject/issue/PIL-1027/wire-up-live-data-with-v2-ui

Part 2.

Includes:
- Pulling service assets balances from Zapper.
- Once again some reasoning name refactor for `Balances`.

Notes:
- I don't like the approach of putting service balances with extra items, but as for this moment this is most pragmatic approach for our internal release.
- Some meta data might be missing such as images for Liquidity Pool tokens (not provided by Zapper) or Deposits APY (need to build extra request flow).

### DIFF
--- a/src/actions/assetsActions.js
+++ b/src/actions/assetsActions.js
@@ -103,7 +103,7 @@ import { sum } from 'utils/bigNumber';
 import { accountAssetsSelector, makeAccountEnabledAssetsSelector } from 'selectors/assets';
 import {
   accountsSelector,
-  balancesSelector,
+  assetsBalancesSelector,
   fiatCurrencySelector,
   ratesSelector,
 } from 'selectors';
@@ -441,6 +441,12 @@ export const fetchAccountWalletBalancesAction = (account: Account) => {
   };
 };
 
+/**
+ * Note: Per current moment of implementation it's not needed to
+ * separate this per single account action because we would even
+ * need to double the requests from Zapper as they allow to query
+ * for multiple addresses and response time does not increase significantly.
+ */
 export const fetchAllAccountsTotalBalancesAction = () => {
   return async (dispatch: Dispatch, getState: GetState) => {
     if (getState().totalBalances.isFetching) return;
@@ -461,12 +467,15 @@ export const fetchAllAccountsTotalBalancesAction = () => {
     }
 
     const currency = fiatCurrencySelector(getState());
-    const zapperUSDBasedRates = await getZapperFiatRates();
-    if (currency !== USD && !zapperUSDBasedRates) {
+
+    const zapperUsdBasedRates = await getZapperFiatRates();
+    if (!zapperUsdBasedRates) {
       dispatch({ type: SET_FETCHING_TOTAL_BALANCES, payload: false });
       reportErrorLog('fetchAllAccountsTotalBalancesAction failed: no zapperUSDBasedRates', { accountsAddresses });
       return;
     }
+
+    const usdRate = zapperUsdBasedRates[currency.toUpperCase()] ?? 1;
 
     try {
       await Promise.all(availableChainProtocols.map(async ({ chain, zapperProtocolIds, zapperNetworkId }) => {
@@ -502,18 +511,48 @@ export const fetchAllAccountsTotalBalancesAction = () => {
             // no need to add anything no matching account found or empty balances
             if (!account || isEmpty(accountProtocolBalances)) return;
 
-            const accountCategoryBalance = accountProtocolBalances.reduce((categoryBalance, balances) => {
-              const protocolBalances = balances?.assets;
-              if (isEmpty(protocolBalances)) return categoryBalance;
+            let categoryAssetsBalances = [];
 
-              let protocolBalanceValues = protocolBalances.map(({ balanceUSD }) => BigNumber(balanceUSD ?? 0));
+            const categoryTotalBalance = accountProtocolBalances.reduce((categoryBalance, balances) => {
+              const protocolAssetsBalances = balances?.assets;
+              if (isEmpty(protocolAssetsBalances)) return categoryBalance;
 
-              if (currency !== USD) {
-                const rate = zapperUSDBasedRates?.[currency.toUpperCase()];
-                protocolBalanceValues = protocolBalanceValues.map((value) => value.times(rate ?? 0));
-              }
+              const assetsBalances = protocolAssetsBalances.map((asset) => {
+                const {
+                  protocol,
+                  symbol,
+                  label,
+                  balanceUSD,
+                  img,
+                  share: tokensShare,
+                  supply: tokensSupply,
+                } = asset;
 
-              return sum([categoryBalance, ...protocolBalanceValues]);
+                const balanceUsdBN = BigNumber(balanceUSD);
+                const value = currency === USD
+                  ? balanceUsdBN
+                  : balanceUsdBN.times(usdRate ?? 0);
+
+                // TODO: do we need to fix exponential for very small values or hide small share at all?
+                const share = tokensShare && tokensSupply
+                  ? BigNumber(tokensShare).dividedBy(tokensSupply).times(100)
+                  : null;
+
+                return {
+                  key: `${protocol}-${symbol}`,
+                  service: balances.label,
+                  title: label,
+                  iconUrl: img ? `https://zapper.fi/images/${img}` : null,
+                  share,
+                  value,
+                };
+              });
+
+              categoryAssetsBalances = [...categoryAssetsBalances, ...assetsBalances];
+
+              const balancesValues = assetsBalances.map(({ value }) => value);
+
+              return sum([categoryBalance, ...balancesValues]);
             }, wrapBigNumber(0));
 
             dispatch({
@@ -522,7 +561,17 @@ export const fetchAllAccountsTotalBalancesAction = () => {
                 accountId: getAccountId(account),
                 chain,
                 category: balanceCategory,
-                balance: accountCategoryBalance,
+                balance: categoryTotalBalance,
+              },
+            });
+
+            dispatch({
+              type: SET_ACCOUNT_ASSETS_BALANCES,
+              payload: {
+                accountId: getAccountId(account),
+                chain,
+                category: balanceCategory,
+                balances: categoryAssetsBalances,
               },
             });
           });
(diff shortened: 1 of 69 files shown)


src/screens/Assets/liquidityPools/selectors.js
```diff
@@ -41,18 +45,7 @@ export function useLiquidityPoolsChainBalances(): TotalBalancesPerChain {
...
-|};
-
-export function useLiquidityPoolAssets(): ChainRecord<LiquidityPoolItem[]> {
-  // TODO: replace once available from Etherspot SDK
-  const liquidityPools = [];
-
-  return { ethereum: liquidityPools };
+export function useLiquidityPoolAssets(): ChainRecord<LiquidityPoolAssetBalance[]> {
+  const assetsBalances = useRootSelector(accountAssetsBalancesSelector);
+  return getChainAssetsBalancesForCategory(assetsBalances, ASSET_CATEGORY.LIQUIDITY_POOLS);
```
IAmKio, line 50:
```suggestion
  return getChainAssetsBalancesForCategory(useRootSelector(accountAssetsBalancesSelector), ASSET_CATEGORY.LIQUIDITY_POOLS);
```
Maybe? Feels like a lot going on for one line, however...

src/services/zapper.js
```diff
@@ -37,9 +37,13 @@ type ZapperAsset = {
   tokenAddress: string,
   label?: string,
   img?: string,
+  noImage?: boolean,
```
IAmKio, line 40: Do you think we need noImage? Could you just `img: ?string` or something similar? However saying this - i'm aware there could be a good reason for the explicit boolean...

IAmKio (approved):
Great work on the refactoring as always! And i can see all my positions - we're living in the future! 🤖

src/services/zapper.js
poocart: 👍

src/screens/Assets/liquidityPools/selectors.js
poocart: 👍

src/screens/Assets/liquidityPools/LiquidityPoolsTab.js
```diff
@@ -92,13 +92,21 @@ function LiquidityPoolsTab() {
...
       return <ServiceListHeader title={headerListItem.key} />;
     }
 
-    const { title, iconUrl, value, change } = headerListItem.item;
-    return <LiquidityPoolListItem title={title} iconUrl={iconUrl} value={value} change={change} />;
+    const { title, iconUrl, value, change, share } = headerListItem.item;
+    return (
+      <LiquidityPoolListItem
+        title={title}
+        subtitle={share ? t('poolShare', { share }) : null}
```
mdjastrzebski, line 104: Pool share could be formatted in LiquidityPoolListItem to match behavior of passing other values as raw values rather than formatted ones.

src/selectors/balances.js
```diff
@@ -38,28 +38,34 @@ import type {
...
   activeAccountIdSelector,
-  (balances, activeAccountId): AssetsBalances => {
+  (balances: AssetBalancesPerAccount, activeAccountId: ?string): CategoryBalancesPerChain | {} => {
     if (!activeAccountId) return {};
-    return balances?.[activeAccountId]?.[CHAIN.ETHEREUM]?.[ASSET_CATEGORY.WALLET] || {};
+    return balances?.[activeAccountId] ?? {};
   },
 );
 
+export const accountEthereumWalletAssetsBalancesSelector = createSelector(
```
mdjastrzebski, line 64: Seems like this selector returning WalletAssetBalances is used as a drop in replacement for AssetBalances type. But the new types stores values a BigNumber while previous one used string. 
1. Pls verify that the affects screens work fine
2. Pls add typing for selector, because createSelectors returns any.
3. 
```suggestion
import type { Selector } from 'reduces/rootReducer';
///.... 
export const accountEthereumWalletAssetsBalancesSelector: Selector<WalletAssetsBalances> = createSelector(
```

src/utils/balances.js
```diff
@@ -38,12 +44,21 @@ export const getTotalBalance = (entries: { [key: string]: BigNumber}): BigNumber
...
-  const balancesOnChains = (Object.values(accountTotals || {}): any);
+  const balancesOnChains = (Object.values(accountTotalBalances || {}): any);
 
   return sum(balancesOnChains.map((chainTotals) => {
     return chainTotals?.[category] || BigNumber(0);
   }));
 };
+
+// TODO: add return AssetsBalancesPerChain later and fix Flow
+export const getChainAssetsBalancesForCategory = (
```
mdjastrzebski, line 58: Having one function to get different categories which are of different types will not work with flow. IMO we need to have 1 function per asset category to get any reasonable type checking by flow.

src/services/zapper.js
```diff
@@ -40,6 +40,9 @@ type ZapperAsset = {
   reserve?: number,
   balance: number,
```
mdjastrzebski, line 41: Just reminding of `balanceRaw: string`

src/screens/Tank/SettleBalanceConfirm/SettleBalanceConfirm.js
```diff
@@ -105,7 +105,7 @@ class SettleBalanceConfirm extends React.Component<Props, State> {
     // add unsettled amounts to balances
     const combinedBalances = Object.keys(balances)
       .reduce((memo, assetName) => {
-        const balanceData: AssetBalance = balances[assetName];
+        const balanceData: WalletAssetBalance = balances[assetName];
         let balance = new BigNumber(balanceData.balance);
```
mdjastrzebski, line 109: Wrapping in BigNumber should no longer be needed, since balance.balance should now be BigNumber itself.

src/actions/assetsActions.js
```diff
@@ -502,18 +511,48 @@ export const fetchAllAccountsTotalBalancesAction = () => {
...
+                  key: `${protocol}-${symbol}`,
+                  service: balances.label,
+                  title: label,
+                  iconUrl: img ? `https://zapper.fi/images/${img}` : null,
+                  share,
+                  value,
+                };
+              });
+
+              categoryAssetsBalances = [...categoryAssetsBalances, ...assetsBalances];
```
mdjastrzebski, line 551: Seems like this reduce is not pure function because it calculating both total & asset array, this can be confusing to discover. Consider having adding 
```
let categoryTotalBalances = BigNumber(0) 
```
next to `let categoryAssetsBalances = [];`

and replacing `reduce` with `forEach`.

src/actions/assetsActions.js
```diff
@@ -461,13 +467,16 @@ export const fetchAllAccountsTotalBalancesAction = () => {
...
-    if (currency !== USD && !zapperUSDBasedRates) {
+
+    const zapperUsdBasedRates = await getZapperFiatRates();
+    if (!zapperUsdBasedRates) {
       dispatch({ type: SET_FETCHING_TOTAL_BALANCES, payload: false });
       reportErrorLog('fetchAllAccountsTotalBalancesAction failed: no zapperUSDBasedRates', { accountsAddresses });
       return;
     }
 
+    const usdRate = zapperUsdBasedRates[currency.toUpperCase()] ?? 1;
```
mdjastrzebski, line 478: Defensive programming:
This check is somewhat unsafe, as it assumes that if currency is not found, then it must be USD, Could we check explicitly for USD and then return 1, so that all unexpected currencies would return undefined?

src/models/Balances.js
```diff
@@ -38,32 +38,58 @@ export type ChainTotalBalancesPerAccount = {
...
+|};
+
+export type DepositAssetBalance = {|
+  ...ServiceAssetBalance,
+  currentApy: ?BigNumber,
+|};
+
+export type WalletAssetBalance = {|
   balance: string,
   symbol: string,
```
mdjastrzebski, line 73: I would suggest migrating that to BigNumber on first possible occasion, but it will probably require changes to the call sites as well.

mdjastrzebski (approved):
Nice job man! 💯 

Is see that the new asset balances are somewhat mix of old and new: WalletAssetBalance (with string value) and services AssetBalances (with BigNumber values). I guess it's fine for now, but would recommend moderniing WalletAssetBalance to use BN as soon as feasible.

src/screens/Tank/SettleBalanceConfirm/SettleBalanceConfirm.js
mdjastrzebski: Eh, just noticed that WalletAssetBalance still keeps balance as string. Pls ignore this comment

src/models/Balances.js
poocart: Totally agree on idea to migrate/refactor this.

src/actions/assetsActions.js
poocart: Thanks, this is also called late night programming :) I thought of not unblocking user from getting balances if rates are not returned, but in fact rate check is done above and actually if there's no rate we shouldn't update at all rather than show wrong balances, will update accordingly, hope it makes sense 👌

src/actions/assetsActions.js
poocart: Yeah I guess either way is fine, but putting totals outside the loop makes more sense 👍

src/models/Balances.js
```diff
@@ -38,32 +38,58 @@ export type ChainTotalBalancesPerAccount = {
...
-  wallet?: AssetsBalances,
-  deposits?: AssetsBalances,
-  investments?: AssetsBalances,
-  liquidityPools?: AssetsBalances,
-  rewards?: AssetsBalances,
+  wallet?: WalletAssetsBalances,
+  deposits?: DepositAssetBalance[],
+  investments?: InvestmentAssetBalance[],
+  liquidityPools?: LiquidityPoolAssetBalance[],
+  rewards?: WalletAssetsBalances,
```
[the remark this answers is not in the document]
poocart: No, will add a basic copy type from other services.

src/screens/Assets/liquidityPools/LiquidityPoolsTab.js
poocart: My idea behind this is to separate it in case we want to add different subtitle for different `LiquidityPool`

src/models/Balances.js
```diff
@@ -38,32 +38,58 @@ export type ChainTotalBalancesPerAccount = {
...
 |};
 
-export type AssetBalance = {
+export type ServiceAssetBalance = {|
```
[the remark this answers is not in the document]
poocart: I guess leave it for now and address it later once we refactor `WalletAssetBalance` to match `ServiceAssetBalance` type.

src/selectors/balances.js
poocart: It does, cannot find where it actually returned BigNumber? It used to be `Balances` type which was always string for `balance` 🤔 screens work fine.

poocart: Ah I see, it expects `BigNumber` on Assets screen for Wallet category.

poocart: For that `useWalletAssets` wraps it into `BigNumber` before it reaches UI.

PR #2651, opened 2021-05-24
Query chains deployment state

mdjastrzebski:
Scope:
- [x] Checking etherspot deployment status for all chains
- [x] Modifying account extra for etherspot account to hold `ChainRecord<EtherspotAccount>` instead of single `EtherspotAccount`
- [x] Received actual deployment status for side chains


src/services/etherspot.js
```diff
@@ -124,8 +118,27 @@ class EtherspotService {
     this.subscription.unsubscribe();
   }
 
-  getAccount(accountAddress: string): Promise<?EtherspotAccount> {
-    return this.sdk.getAccount({ address: accountAddress }).catch((error) => {
+  getSdkForChain(chain: Chain): ?EtherspotSdk {
+    const network = networkNameFromChain(chain);
+    if (!network) {
+      reportErrorLog('EtherspotService getSdkForChain failed', { chain });
```
poocart, line 124:
```suggestion
      reportErrorLog('EtherspotService getSdkForChain failed: no network', { chain });
```

src/services/etherspot.js
```diff
@@ -124,8 +118,27 @@ class EtherspotService {
...
+      return undefined;
```
poocart, line 125: Maybe just return? And for the one below.

```suggestion
      return;
```

src/services/etherspot.js
```diff
@@ -124,8 +118,27 @@ class EtherspotService {
...
+    }
+
+    const sdk = this.instances[network];
+    if (!sdk) {
+      reportErrorLog('EtherspotService getSdkForChain failed', { chain, network });
```
poocart, line 130:
```suggestion
      reportErrorLog('EtherspotService getSdkForChain failed: cannot get SDK instance', { chain, network });
```

src/services/etherspot.js
```diff
@@ -124,8 +118,27 @@ class EtherspotService {
...
+      return undefined;
+    }
+
+    return sdk;
+  }
+
+  getAccount(chain: Chain, accountAddress: string): Promise<?EtherspotAccount> {
+    const sdk = this.getSdkForChain(chain);
+    if (!sdk) return Promise.resolve(null);
```
poocart, line 139: It can be just `null` I guess
```suggestion
    if (!sdk) return null;
```

poocart (approved):
Great work! 🚀  Just a note that app is using Account extra for some other calls such as `getAccountEnsName` which needs to be addressed here as well, DM'd you on this 👌

src/services/etherspot.js
mdjastrzebski: Yes, but had to updated return type to `Promise | null`

PR #2653, opened 2021-05-24
chore/etherspot-1-0-0

IAmKio:
Required to allow ENS migrations to work correctly. Checked with @stanislaw-glogowski - no changes are required within the app.

- Bumping Etherspot Migrator package to 1.0.0
- Stories flag is true by default
- Firebase Analytics user data V3 flag converted to a string as per service limitations


poocart (approved):
🚀

stanislaw-glogowski (approved):

PR #2658, opened 2021-05-25
remove platform core badges fetch

poocart:
### Part of Platform Core remove

This removes new badges fetching and creation calls made from app to platform core service.

Note: it does not completely remove Badges from UI to keep ones for existing users that are stored locally, but after wallet is reimported it will not pull existing badges and they would be lost.


mdjastrzebski (approved):
Does it mean that we no longer have badges in our system? If so we might remove also models, etc

PR #2674, opened 2021-06-02
remove platform core icons

poocart:
Removes icons based on platform core hostname.

Note: there were few hardcoded lists with icons pointing to production platform core, added direct urls to them and will ask back-end to keep those in S3 or we can move those into app later (per discussion with @IAmKio).


mdjastrzebski (approved):
LG

PR #2677, opened 2021-06-02
PIL-1047: fcm to etherspot

poocart:
https://linear.app/pillarproject/issue/PIL-1047/complete-fcmtoken-integration

### DIFF
--- a/src/services/etherspot.js
+++ b/src/services/etherspot.js
@@ -96,18 +96,22 @@ export class EtherspotService {
      * array of instantiated instances
      */
     await Promise.all(this.supportedNetworks.map(async (networkName) => {
-      try {
-        const env = networkName !== NetworkNames.Kovan ? EnvNames.MainNets : EnvNames.TestNets;
-        this.instances[networkName] = new EtherspotSdk(privateKey, { env, networkName });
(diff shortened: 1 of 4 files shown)


src/services/etherspot.js
```diff
@@ -96,19 +95,24 @@ export class EtherspotService {
...
+        }
 
-      // Schedule exection of computeContractAccount's
-      etherspotComputeContractPromises.push(this.instances[networkName].computeContractAccount({ sync: true }));
-    });
+        await this.instances[networkName].computeContractAccount({ sync: true });
+      } catch (error) {
+        reportErrorLog('EtherspotService network init failed', { networkName, error });
+      }
+    }));
```
IAmKio, line 113: I'm seeing this pattern emerge where we have nested `await`s, `try / catch`es, wrapped by another `await`s - i'm just a bit worried that this is starting to get messy, complex and less readable. I think whilst technically it works - we need to have a wider discussion about how we measure cyclomatic complexity within our code after V2 launch.

IAmKio (approved):

src/services/etherspot.js
poocart: I do agree, it needs wider discussion. However, separated 2 try catch's for each `await`.

PR #2684, opened 2021-06-03
fix/remove-branch-json-reference

IAmKio:
Removed branch.json reference in `circleci/config.yml`


poocart (approved):

antonioned (commented):
If we do remove these from android job, should the same be removed from prod_ios as well?

IAmKio:
> If we do remove these from android job, should the same be removed from prod_ios as well?

Done - cold you check again please?

antonioned (approved):

PR #2688, opened 2021-06-05
PIL-1053: Exchange: migrate to Etherspot

mdjastrzebski:
Scope:
- [x] Use app supported assets instead of exchange specific assets
- [x] Fetch exchange offers from Etherspot
- [x] Send exchange transation to Etherspot
- [x] UI changes as per design
- [x] Remove exchange redux state/actions in favor of component state


src/components/AddFundsModal/AddFundsModal.js
```diff
@@ -102,31 +67,15 @@ const AddFundsModal = ({ token, receiveAddress }: Props) => {
       iconName: 'qrDetailed',
       onPress: openReceiveModal,
     },
-    {
+    isExchangeAvailable && {
       key: 'exchange',
       label: t('button.exchange'),
       iconName: 'flip',
       onPress: () => navigation.navigate(EXCHANGE, token && { toAssetCode: token }),
-      hide: hideExchangeOption,
```
poocart, line 110: Can we move this logic outside of array? In case this is falsy options end up having false and then we have to call another method to clean it which for me seems extra load 🤔

src/constants/exchangeConstants.js
```diff
@@ -17,45 +17,18 @@
...
-export const SET_UNISWAP_TOKENS_QUERY_STATUS: 'SET_UNISWAP_TOKENS_QUERY_STATUS' = 'SET_UNISWAP_TOKENS_QUERY_STATUS';
 
-export const PROVIDER_UNISWAP = 'UNISWAPV2-SHIM';
-export const PROVIDER_1INCH = 'ONEINCH-SHIM';
-export const PROVIDER_SYNTHETIX = 'SYNTHETIX';
-
-export const MARK_NOTIFICATION_SEEN = 'MARK_NOTIFICATION_SEEN';
-
-export const EXCHANGE = 'EXCHANGE';
+// Note: these values are matching Etherspot values
```
poocart, line 21: Is it possible maybe to export constants from Etherspot SDK in case it changes?

src/screens/Asset/Asset.js
```diff
@@ -324,7 +307,7 @@ class AssetScreen extends React.Component<Props> {
             <AssetButtons
               onPressReceive={this.openAddFundsModal}
               onPressSend={this.goToSendTokenFlow}
-              onPressExchange={isSupportedByExchange ? () => this.goToExchangeFlow(token) : null}
+              onPressExchange={() => this.goToExchangeFlow(token)}
```
poocart, line 310: Exchange everything! 😂

src/screens/Assets/wallet/WalletTab.js
```diff
@@ -122,13 +123,14 @@ function WalletTab() {
...
   const hasPositiveBalance = totalBalance.value.gt(0);
+
   const buttons = [
     hasPositiveBalance && {
       title: tRoot('button.receive'),
       iconName: 'qrcode',
       onPress: showReceiveModal,
     },
-    hasPositiveBalance && {
+    isExchangeAvailable && hasPositiveBalance && {
```
poocart, line 133: Just as per comment before – can we move inside-array logical operators outside? This results in arrays having falsy values and can easily slip somewhere unfiltered therefore in some scenarios causing crash.

mdjastrzebski: This is tightly scoped to `<FloatingButtons>` with proper type checking, I would keep it as is, because it's much simpler to define it this way then do `let array; if(xxx) array.push(yyy)`. This is established pattern with things likes React Native `style` prop which also accepts array with falsy values.

src/screens/Asset/Asset.js
mdjastrzebski: That's a feature 😂

src/constants/exchangeConstants.js
mdjastrzebski: Resolved by separating these types & creating `parseExchangeProvider` function. Now app should continue working even if etherspot sends new/unknown provider.

src/components/AddFundsModal/AddFundsModal.js
mdjastrzebski: Falsy arrays seem to be quite established concept in React/RN space, e.g. `View.style` prop can accept array of falsy values to allow for more fluent conditional prop passing.

poocart: Yeah, but this is different scope as it would break if filters no applied  👀

src/utils/types/etherspot.js
```diff
@@ -0,0 +1,63 @@
...
+  chainId: number,
+  decimals: number,
+  logoURI: string,
+  name: string,
+  symbol: string,
+};
+
+export type ExchangeOffer = {
+  exchangeRate: number,
+  provider: ExchangeProviders,
```
poocart, line 42: I suddenly expected this to be an array, maybe `ExchangeProviders` can be singular instead of plural, wdyt?

src/components/AddFundsModal/AddFundsModal.js
mdjastrzebski: Yeah, but receiving component has proper typing and applies the necessary filtering. Since this uses strict typing and prop is passed directly to related component the risk of breaking should be very low.

src/utils/types/etherspot.js
```diff
@@ -0,0 +1,63 @@
...
+  receiveAmount: EthersBigNumber,
+  transactions: TransactionData[],
+};
+
+export type ExchangeProviders = 'OneInch' | 'Synthetix' | 'Uniswap';
+
+export type TransactionData = {
```
poocart, line 49: Can you reuse `EthereumTransaction`? I see that we tend to duplicate transaction models.

src/utils/types/etherspot.js
mdjastrzebski: It's etherspot name, not mine

src/utils/types/etherspot.js
```diff
@@ -0,0 +1,63 @@
...
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import { BigNumber as EthersBigNumber } from 'ethers';
+
+export type PaginationResult<T = any> = {
+    items?: T[];
+    currentPage: number;
+    nextPage: number;
```
poocart, line 26:
```suggestion
  items?: T[];
  currentPage: number;
  nextPage: number;
```

src/utils/types/etherspot.js
mdjastrzebski: It's typing for etherspot types, as they use typescript which flow is not aware of. For our stuff we should use our models. Coincidently they might sometimes match.

src/utils/format.js
```diff
@@ -226,3 +226,40 @@ export function formatHexAddress(address: string) {
...
+  if (reverse) {
+    rate = BigNumber(1).dividedBy(rate);
+    return formatExchangeRate(rate, toSymbol, fromSymbol);
+  }
+
+  const rateString = formatExchangeRateWithoutSymbol(rate);
+  return t('exchangeRate', { rate: rateString, fromSymbol, toSymbol });
+}
+
+export function formatExchangeRateWithoutSymbol(rate: ?BigNumber | number) {
```
poocart, line 248: Weird, I think I've seen this util before, but can't find anything 🤔

mdjastrzebski: Maybe there was some older one, but I intend `utils/format` to contain modern replacement for old formatting functions.

poocart (approved):
Great job man!!! Friday YOLO! 🚀

src/screens/Exchange/ExchangeConfirm/ExchangeScheme.js
```diff
@@ -135,88 +151,57 @@ const getLeftArrowPath = (toValueWith: number) => {
...
+  position: absolute;
+  height: 112px;
+  width: ${SCREEN_WIDTH / 2}px;
+  ${({ isLeft }) =>
+    isLeft
+      ? `left: 0;
+      top: 50%;`
+      : `right: 0;
+      bottom: 50%;
+      margin-bottom: -9px;`}
```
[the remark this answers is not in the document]
mdjastrzebski: It's legacy code, I just shuffled it around.

PR #2698, opened 2021-06-11
PIL-1054: Deploy label modification

divyaBacancy:


src/screens/Home/components/ChainListItem.js
```diff
@@ -39,23 +38,19 @@ export type Props = {|
...
-        <DeployContainer onPress={onPressDeploy}>
-          <DeployValue>{t('button.deploy')}</DeployValue>
-          <Icon name="question" width={14} height={14} color={colors.labelTertiary} />
+      {!isDeployed ? (
+        <DeployContainer>
+          <TitleWithIcon onPress={onPress}>{title}</TitleWithIcon>
+          <HazardIcon name="warning" width={16} height={16} color={colors.hazardIconColor} onPress={onPressDeploy} />
         </DeployContainer>
+      ) : (
+        <Title>{title}</Title>
```
poocart, line 51: IMO using ternary for conditional rendering is hard to read when it grows. Can we render here with basic logical operators?

poocart (approved):
LGTM, just small comment.

src/screens/Home/components/ChainListItem.js
```diff
@@ -68,23 +63,28 @@ const Container = styled.TouchableOpacity`
...
+`;
+
 const Title = styled(Text)`
   flex: 1;
   margin: ${spacing.mediumLarge}px 0 ${spacing.mediumLarge}px 36px;
   ${fontStyles.medium};
 `;
 
-const Value = styled(Text)`
+const TitleWithIcon = styled(Text)`
```
mdjastrzebski, line 82: The name is a bit confusing, as there is no icon in this particular component.

src/screens/Home/components/ChainListItem.js
```diff
@@ -39,23 +38,19 @@ export type Props = {|
...
-      )}
-
-      {!isDeployed && (
-        <DeployContainer onPress={onPressDeploy}>
-          <DeployValue>{t('button.deploy')}</DeployValue>
-          <Icon name="question" width={14} height={14} color={colors.labelTertiary} />
+      {!isDeployed ? (
+        <DeployContainer>
+          <TitleWithIcon onPress={onPress}>{title}</TitleWithIcon>
+          <HazardIcon name="warning" width={16} height={16} color={colors.hazardIconColor} onPress={onPressDeploy} />
```
mdjastrzebski, line 48: As the icon is small you probably need to add thing like `hitSlop={hitSlop20}` for user to be able to trigger the action.

src/screens/Home/components/ChainListItem.js
mdjastrzebski: Seem like if you removed `flex: 1` from Title you could share the same style with that.

mdjastrzebski (changes requested):
Generally looks good, some minor tweaks suggested.

mdjastrzebski (approved):
I intended to select approve

src/screens/Home/components/ChainListItem.js
divyaBacancy: Done.

src/screens/Home/components/ChainListItem.js
divyaBacancy: Done.

src/screens/Home/components/ChainListItem.js
divyaBacancy: Done.

PR #2700, opened 2021-06-11
multichain events, multichain tx details

poocart:
Includes:
- Multi chain events listener.
- Fixes for multi chain transaction events listener.
- Added multi chain support for transaction details.

### DIFF
--- a/src/components/HistoryEventDetails/layouts/TokenExchangeEventDetails.js
+++ b/src/components/HistoryEventDetails/layouts/TokenExchangeEventDetails.js
@@ -43,15 +43,17 @@ import { spacing } from 'utils/variables';
 
 // Types
 import type { TokenExchangeEvent } from 'models/History';
+import type { Chain } from 'models/Chain';
 
 // Local
 import BaseEventDetails from './BaseEventDetails';
 
 type Props = {|
   event: TokenExchangeEvent,
+  chain: Chain,
 |};
 
-function TokenExchangeEventDetails({ event }: Props) {
+function TokenExchangeEventDetails({ event, chain }: Props) {
   const { t } = useTranslation();
   const navigation = useNavigation();
   const dispatch = useDispatch();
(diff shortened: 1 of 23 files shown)


src/services/etherspot.js
```diff
@@ -120,15 +120,37 @@ export class EtherspotService {
...
-    this.subscription = this.sdk.notifications$.pipe(map(callback)).subscribe();
+      this.subscriptions[networkName] = sdk.notifications$
+        .pipe(map(((notification) => callback(chain, notification))))
+        .subscribe();
+    });
+  }
+
+  unsubscribeNetworkEvents(network: string) {
+    const subscription = this.subscriptions[network];
+    if (subscription) subscription.unsubscribe();
```
mdjastrzebski, line 147: We probably should clear subscription field 
```
this.subscription[network] = null;
```

src/services/etherspot.js
```diff
@@ -120,15 +120,37 @@ export class EtherspotService {
...
   }
 
   unsubscribe() {
-    if (!this.subscription) return;
-    this.subscription.unsubscribe();
+    this.supportedNetworks.forEach(async (networkName) => {
+      this.unsubscribeNetworkEvents(networkName);
+    });
```
mdjastrzebski, line 153: I would inline `unsubscribeNetworkEvents` probably

src/components/HistoryEventDetails/layouts/TokenExchangeEventDetails.js
```diff
@@ -43,15 +43,17 @@ import { spacing } from 'utils/variables';
...
 // Local
 import BaseEventDetails from './BaseEventDetails';
 
 type Props = {|
   event: TokenExchangeEvent,
+  chain: Chain,
 |};
 
-function TokenExchangeEventDetails({ event }: Props) {
+function TokenExchangeEventDetails({ event, chain }: Props) {
```
mdjastrzebski, line 56: Looking at these repeated changes, it would probably make sense put chain info in the `Event` type directly. wdyt?

mdjastrzebski (approved):
Looks good,

One thing I would consider is putting `chain` prop into `Event` type instead of passing it around for each `EventDetail` component.

PR #2702, opened 2021-06-11
PIL-1061: ENS name is too small and 0x address overflows

divyaBacancy:


package.json
```diff
@@ -112,6 +112,7 @@
     "react-native": "0.63.4",
     "react-native-android-badge": "^0.4.3",
     "react-native-appearance": "^0.3.2",
+    "react-native-auto-scale-text": "^1.0.7",
```
mdjastrzebski, line 115: Not sure if we really need a npm dependency for that. `Text` component has `adjustsFontSizeToFit` prop which allows for nice scaling down of text.

mdjastrzebski (changes requested):
Not sure if we really need a npm dependency for that. Text component has adjustsFontSizeToFit prop which allows for nice scaling down of text.

package.json
divyaBacancy: @mdjastrzebski Thanks for letting me know. Done with the changes.

src/screens/Asset/ReceiveModal.js
```diff
@@ -117,24 +118,35 @@ const ReceiveModal = ({
...
-            <WalletAddress>{address}</WalletAddress>
+            <WalletAddress numberOfLines={1} adjustsFontSizeToFit>
+              {address}
+            </WalletAddress>
           ) : (
             <TextWithCopy
               toastText={t('toast.addressCopiedToClipboard')}
               textToCopy={address}
-              textStyle={{ color: colors.basic030 }}
+              textStyle={{ color: colors.basic030, fontSize: fontSizes.regular }}
```
mdjastrzebski, line 138: regular should be default, so probably no need to pass it here explicitly.

src/screens/Asset/ReceiveModal.js
```diff
@@ -117,24 +118,35 @@ const ReceiveModal = ({
       <ContentWrapper forceInset={{ top: 'never', bottom: 'always' }}>
         <InfoView>
           {!!ensName && (
-            <TextWithCopy textToCopy={ensName} toastText={t('toast.ensNameCopiedToClipboard')} iconColor={colors.link}>
+            <TextWithCopy
+              textToCopy={ensName}
+              toastText={t('toast.ensNameCopiedToClipboard')}
+              iconColor={colors.link}
+              textStyle={{ fontSize: fontSizes.big }}
```
mdjastrzebski, line 125: please extract these styles below as ```const styles = {
  ensName: { ... }
  address: { ... } 
};
```

File: src/screens/Asset/ReceiveModal.js (hunk 2 above)
Comment (divyaBacancy) [reply src/screens/Asset/ReceiveModal.js]: Done.

**Review by poocart — APPROVED** [review]
lgtm 👍

File: src/screens/Asset/ReceiveModal.js [hunk 4]
```diff
@@ -117,24 +118,35 @@ const ReceiveModal = ({
       <ContentWrapper forceInset={{ top: 'never', bottom: 'always' }}>
         <InfoView>
           {!!ensName && (
-            <TextWithCopy textToCopy={ensName} toastText={t('toast.ensNameCopiedToClipboard')} iconColor={colors.link}>
+            <TextWithCopy
+              textToCopy={ensName}
+              toastText={t('toast.ensNameCopiedToClipboard')}
+              iconColor={colors.link}
+              textStyle={styles.ensName}
+            >
```
mdjastrzebski, line 126: You still need to apply `adjustFontSizeToFit` here (and expose it on `TextWithCopy` component), as it breaks without ens name,

src/screens/Asset/ReceiveModal.js
```diff
@@ -186,9 +207,8 @@ const QRCodeWrapper = styled.View`
 `;
 
 const WalletAddress = styled(Text)`
```
mdjastrzebski, line 209: According to design wallet address should use text style `small` as its 12px font. Pls apply it to both ens & non-ens scenario.

mdjastrzebski (changes requested):
Layout for use with ENS name looks nice, but for user without ENS name it still breaks with address & icon line breaking.

src/screens/Asset/ReceiveModal.js
divyaBacancy: Done.

src/screens/Asset/ReceiveModal.js
divyaBacancy: Done.

src/screens/Asset/ReceiveModal.js
```diff
@@ -173,6 +189,16 @@ const combinedMapStateToProps = (state: RootReducerState): $Shape<StateProps> =>
 
 export default (connect(combinedMapStateToProps)(ReceiveModal): AbstractComponent<OwnProps>);
 
+const styles = {
+  ensName: {
+    fontSize: fontSizes.big,
+  },
+  address: {
+    color: getThemeColors().basic030,
```
mdjastrzebski, line 197: getThemeColors it seems that `getThemeColors` expects theme to be passed and you only can pass them from inside the component as you get it from hook/context. Solution here would be to apply `textStyle={[styles.address, color: colors.basic030]}`

Regarding color values, pls use `secondaryText` instead of `basic030` as it's the same hex value for both light & dark but it's easier to read.

mdjastrzebski (changes requested):
The code is almost there, good work.

Couple of points to finish it:
1. I've noticed that the warning "Receive Etherereum Mainnet assets with caution" which leads to deployment interjection is always displayed. IMO it should be displayed only if the ethereum mainnet wallet is not yet deployed. You can check it's status using `useDeploymentStatus` hook & you probably should use `showDeploymentInterjection` from that hook instead of navigating manually to `ETHERSPOT_DEPLOYMENT_INTERJECTION` as the current way does not support archanova smart wallet interjection.
2. Translation names of `paragraph.cautionMessage` & `paragrap.withCaution` are too generic. Please move them over more specific translation keys, e,g. `receiveModal.cautionMessage`, etc. 
3. `paragraph.cautionMessage` probably needs to be split into two messages: 1. `All chains use single wallet address.` which makes sense for etherspot accounts regardless of deployment status, but probably should be hidden for archanova. & 2. `Receive <0>{{chain}}</0> assets` which makes sense only for non-depolyed accounts but both etherspot & archanova

@IAmKio ^ do the above suggestion in point 3 make sense from your POV?

src/screens/Asset/ReceiveModal.js
IAmKio: @divyaBacancy @mdjastrzebski Point 1 definitely makes sense - however we may need to be specific when we say `All chains` because there will always be someone who takes this to mean that they can use their crazy EVM blockchain that we don't support on Etherspot 😂 but yes this message should be hidden altogether from Archanova accounts. Maybe instead of `All chains use single wallet address` we could re-word it to something like `Our supported Ethereum blockchains (mainnet, Matic, ...) use the above address` - just a suggestion.

Point 2 works in both scenarios.

IAmKio: @divyaBacancy @mdjastrzebski on further reflection - i'm going to consult the design and product team on how we should approach this message. It might require design rework. Let's skip this for now.

divyaBacancy: @IAmKio @mdjastrzebski @poocart Done with the changes.

src/screens/Asset/ReceiveModal.js
```diff
@@ -117,34 +119,56 @@ const ReceiveModal = ({
...
+          <QRCodeWrapper>
+            <QRCodeWithTheme value={address} size={104} />
+          </QRCodeWrapper>
+        )}
+        {activeAccount?.type !== ACCOUNT_TYPES.ARCHANOVA_SMART_WALLET && (
+          <WarningText style={{ marginTop: spacing.medium }} center small>
+            {t('receiveModal.message')}
+          </WarningText>
+        )}
+        {!isDeployedOnChain[CHAIN.ETHEREUM] && (
```
mdjastrzebski, line 160:
```suggestion
        {!isDeployedOnChain.ethereum && (
```

src/screens/Asset/ReceiveModal.js
```diff
@@ -117,34 +119,56 @@ const ReceiveModal = ({
...
-          <Text color={colors.link} onPress={() => navigation.navigate(ETHERSPOT_DEPLOYMENT_INTERJECTION)}>
-            {t('paragraph.withCaution')}
-          </Text>
-        </WarningText>
+        {!!address && (
+          <QRCodeWrapper>
+            <QRCodeWithTheme value={address} size={104} />
+          </QRCodeWrapper>
+        )}
+        {activeAccount?.type !== ACCOUNT_TYPES.ARCHANOVA_SMART_WALLET && (
```
mdjastrzebski, line 155:
```suggestion
        {isEtherspotAccount(activeAccount) && (
```

src/screens/Asset/ReceiveModal.js
```diff
@@ -117,34 +119,56 @@ const ReceiveModal = ({
...
+          <WarningText style={{ marginTop: spacing.medium }} center small>
+            {t('receiveModal.message')}
```
mdjastrzebski, line 157: Pls name the translation key in a bit more descriptive way
```suggestion
            {t('receiveModal.etherspotSingleAddressInfo')}
```

src/screens/Asset/ReceiveModal.js
```diff
@@ -117,34 +119,56 @@ const ReceiveModal = ({
...
+          </WarningText>
+        )}
+        {!isDeployedOnChain[CHAIN.ETHEREUM] && (
+          <WarningText center small>
+            {t('receiveModal.cautionMessage', {
```
mdjastrzebski, line 162:
```suggestion
            {t('receiveModal.notDeployedWarning', {
```

src/screens/Asset/ReceiveModal.js
```diff
@@ -117,34 +119,56 @@ const ReceiveModal = ({
...
-            {t('paragraph.withCaution')}
-          </Text>
-        </WarningText>
+        {!!address && (
+          <QRCodeWrapper>
+            <QRCodeWithTheme value={address} size={104} />
+          </QRCodeWrapper>
+        )}
+        {activeAccount?.type !== ACCOUNT_TYPES.ARCHANOVA_SMART_WALLET && (
+          <WarningText style={{ marginTop: spacing.medium }} center small>
```
mdjastrzebski, line 156: 1. modern `Text` does not accept `center` nor `small` props. Pls remove them as they have no effect.

```suggestion
          <WarningText style={{ marginTop: spacing.medium }}>
```

2. Pls extract `{ marginTop: spacing.medium }` as `styles.signleAddressInfo` style.

src/screens/Asset/ReceiveModal.js
```diff
@@ -117,34 +119,56 @@ const ReceiveModal = ({
...
+            {t('receiveModal.message')}
+          </WarningText>
+        )}
+        {!isDeployedOnChain[CHAIN.ETHEREUM] && (
+          <WarningText center small>
```
mdjastrzebski, line 161: Modern `Text` does not accept `small` or `center` props, pls remove these.
```suggestion
          <WarningText>
```

mdjastrzebski (commented):
Code-wise looks good!

Left some stylistic comments.

src/screens/Asset/ReceiveModal.js
divyaBacancy: Done.

src/screens/Asset/ReceiveModal.js
divyaBacancy: Done.

src/screens/Asset/ReceiveModal.js
divyaBacancy: Done.

src/screens/Asset/ReceiveModal.js
divyaBacancy: Done.

src/screens/Asset/ReceiveModal.js
divyaBacancy: Done

src/screens/Asset/ReceiveModal.js
divyaBacancy: Done

PR #2703, opened 2021-06-11
PIL-1057:Top of the screen is cut off

divyaBacancy:
Since the value of "top" used with spacing varies for different devices, hence it is behaving differently with different screens. It is better to put the constant value here as done in other screens. I have tested this value in both platform and various devices. It is working fine.


poocart:
Just a question – have you found out why 60? Is this related to header height size or bottom nav bar side or what?

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferIntro/KeyBasedAssetTransferIntro.js
```diff
@@ -63,7 +61,7 @@ function KeyBasedAssetTransferIntro() {
         backgroundColor={darkTheme.colors.basic090}
       >
         <Content>
-          <Spacing h={top} />
+          <Spacing h={60} />
```
mdjastrzebski, line 64: If I understand the issue correctly the problem is that top of the screen has a different shade of black there. 
Simple solution would be to change `darkTheme.colors.basic090` to `darkTheme.colors.background`.

Proper solution would be to create a new floating back button component that will allow content to flow underneath itself.

Proposed solution of fixed top spacing does not seem to address that.

mdjastrzebski (changes requested):

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferIntro/KeyBasedAssetTransferIntro.js
divyaBacancy: @mdjastrzebski @IAmKio @poocart Done.

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferIntro/KeyBasedAssetTransferIntro.js
```diff
@@ -57,12 +58,12 @@ function KeyBasedAssetTransferIntro() {
...
-        headerProps={{ floating: true }}
-        backgroundColor={darkTheme.colors.basic090}
-      >
-        <Content>
+      <Container>
+        <HeaderBlock
+          navigation={navigation}
+          noPaddingTop
+        />
+        <Content showsVerticalScrollIndicator={false}>
```
mdjastrzebski, line 66: 1. If using `Container` from `components/modern/Layout` you probably would want to use `Content` component from there as well as it handles bottom safe are for iPhone X correctly.

2. Not sure if we would want to hide scroll indicator there. 

3. Additionally in the current state the top safe area is applied to the scroll view, making the top padding of content weird. Using `Content` from `modern/Layout` hopefully should solve it. If not then please do some style override on for that.

image

mdjastrzebski (changes requested):
Suggested some small tweaks as I've noticed some layout issues related to safe area handling

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferIntro/KeyBasedAssetTransferIntro.js
divyaBacancy: @mdjastrzebski Done. Point 2-  scroll indicator was not there previously, so implemented in such a way.

mdjastrzebski: IMO I would keep the scroll indicator in default mode as there seems to be no UI or UX reason to hide it.

mdjastrzebski (changes requested):
Almost there 👍🏻 

Last small thing would be to set background color of the screen as in design https://app.zeplin.io/project/5fa95ca539e6804a10dec0a4/screen/5fb334b092ad8e77a0404aad

Design says its #000000 (dark.basic090) and the current background color is more like dark grey. This normally would not be an issue, but it makes the image stand out:

image

Please apply the color to both `Content` & `HeaderBlock`.

src/screens/KeyBasedAssetTransfer/KeyBasedAssetTransferIntro/KeyBasedAssetTransferIntro.js
mdjastrzebski: @IAmKio are you aware of any reasons why might want to hide scroll indicator on this screen?

mdjastrzebski:
Also noticed that in the designs working on the button is "Upgrade" vs the current implementation "Migrate"

divyaBacancy:
@mdjastrzebski Done. Changed the background color of the screen as in design.

mdjastrzebski (approved):
Looks good 👍🏻

PR #2711, opened 2021-06-14
PIL-1064: Fix/wallet connect transaction fix

poocart:
https://linear.app/pillarproject/issue/PIL-1064/cannot-execute-swap-via-quickswap

This includes:
- Chain native symbol fix for WalletConnect transaction.
- Chain balance check for WalletConnect transaction.
- Fix to wait for actual hash on WalletConnect transaction.
- Rename `isArchanovaWalletActivatedSelector` to `isArchanovaAccountDeployedSelector`;
- Show appropriate WalletConnect error message if other chain than Ethereum and active account Archanova.
- Show active account deployment request only for WalletConnect signature requests on Etherspot account and any request on Archanova.


src/utils/walletConnect.js
```diff
@@ -141,18 +144,19 @@ export const mapCallRequestToTransactionPayload = (
     to = methodToAddress;
   }
 
-  const ethAssetData = getAssetData(accountAssets, supportedAssets, ETH);
+  const { chainId } = callRequest;
+  const chain = chainFromChainId[chainId];
+  const chainNativeSymbol = nativeAssetSymbolPerChain[chain];
+  const chainNativeDecimals = nativeAssetDecimalsPerChain[chain];
+
+  const nativeAssetData = getAssetData(accountAssets, supportedAssets, chainNativeSymbol);
```
mdjastrzebski, line 152: Will it work for BNB token, atm supported assets do not contain BNB.

src/utils/walletConnect.js
```diff
@@ -141,18 +144,19 @@ export const mapCallRequestToTransactionPayload = (
     to = methodToAddress;
   }
 
-  const ethAssetData = getAssetData(accountAssets, supportedAssets, ETH);
+  const { chainId } = callRequest;
+  const chain = chainFromChainId[chainId];
+  const chainNativeSymbol = nativeAssetSymbolPerChain[chain];
+  const chainNativeDecimals = nativeAssetDecimalsPerChain[chain];
```
mdjastrzebski, line 150: Maybe we could get `getNativeAsset(chain)` that would return `Asset` or `AssetData` or `AssetCore` instead of these separated arrays, wdyt?

src/screens/WalletConnect/CallRequest/SignatureRequestContent.js
```diff
@@ -47,10 +54,19 @@ type Props = {|
...
   const { title, iconUrl, chain } = getViewData(request);
   const config = configs[chain];
 
+  /**
+   * Archanova account needs to be deployed regardless of the action,
+   * Etherspot account must to be deployed for signature requests only
+   */
+  const requiresDeployedAccount = (isArchanovaAccount(activeAccount) && !isArchanovaAccountDeployed)
+    || isEtherspotAccountDeployed(activeAccount, chain);
+
```
mdjastrzebski, line 69: Important: second part of the condition seems to be incorrect

```suggestion
  const requiresDeployedAccount = (isArchanovaAccount(activeAccount) && !isArchanovaAccountDeployed)
    || (isEtherspotAccount(activeAccount) && !isEtherspotAccountDeployed(activeAccount, chain);
```

But probably you should use `isDeployedOnChainSelector` which contains the same logic in a nice encapsulation.

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
```diff
@@ -158,18 +166,31 @@ const useTransactionFee = (request: WalletConnectCallRequest) => {
...
+  const isArchanovaAccountActive = isArchanovaAccount(activeAccount);
 
   let errorMessage = null;
   const chain = chainFromChainId[request.chainId];
   if (!chain) {
     errorMessage = t('error.walletConnect.cannotDetermineEthereumChain');
   }
 
-  const isDeployedOnChain = useRootSelector(isDeployedOnChainSelector);
+  if (isArchanovaAccountActive && chain === CHAIN.ETHEREUM) {
```
mdjastrzebski, line 180: Important seems like second condition need to be negated!

```suggestion
  if (isArchanovaAccountActive && chain !== CHAIN.ETHEREUM) {
```

Better yet would use `useSupportedChains` instead of manually figuring out archanova status.

mdjastrzebski (changes requested):
Pls review comments, esp. those about logical conditions as I believe these could be incorrect.

src/screens/WalletConnect/CallRequest/SignatureRequestContent.js
poocart: No, because it returns hardcoded values for deployed states which won't work in case it's not actually deployed. Will add generic selector to make it more clear.

src/utils/walletConnect.js
poocart: It won't. It will work once we add support to it.

src/utils/walletConnect.js
poocart: Not sure since we have `AssetData` as type and this method might confuse in further development.

src/utils/walletConnect.js
```diff
@@ -141,18 +144,19 @@ export const mapCallRequestToTransactionPayload = (
...
+
+  const nativeAssetData = getAssetData(accountAssets, supportedAssets, chainNativeSymbol);
 
   const {
-    symbol = ETH,
-    address: contractAddress = ethAssetData.address,
-    decimals = 18,
+    symbol = chainNativeSymbol,
+    address: contractAddress = nativeAssetData.address,
```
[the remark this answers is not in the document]
poocart: There won't be any balance for `BNB` therefore it will be blocked even if estimate goes through.

src/actions/walletConnectActions.js
```diff
@@ -163,20 +161,6 @@ export const approveWalletConnectConnectorRequestAction = (peerId: string, chain
       return;
     }
 
-    const chain = chainFromChainId[chainId];
-    const requiresSmartWalletDeployment = !isDeployedOnChainSelector(getState())[chain];
```
mdjastrzebski, line 167: Should we keep some sort of check like this in the action as a defensive measure, wdyt?

src/screens/WalletConnect/CallRequest/SignatureRequestContent.js
```diff
@@ -47,10 +54,19 @@ type Props = {|
...
 
   const { title, iconUrl, chain } = getViewData(request);
   const config = configs[chain];
 
+  /**
+   * Archanova account needs to be deployed regardless of the action,
+   * Etherspot account must to be deployed for signature requests only
+   */
+  const requiresDeployedAccount = (isArchanovaAccount(activeAccount) && !isArchanovaAccountDeployed)
+    || isEtherspotAccountDeployed(activeAccount, chain);
```
mdjastrzebski, line 68: This condition still seems wrong for etherspot: 
* if etherspot account is deployed it will be true meaning it needs deployment
* if etherspot account is not deployed it will be false meaning it does not need deployment

* for archanova deployed/not-deployes it seems ok, but a bit hackish as it assumes that `isEtherspotAccountDeployed` will return false when account is not etherspot, which is not very clear.

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
```diff
@@ -158,18 +166,31 @@ const useTransactionFee = (request: WalletConnectCallRequest) => {
...
-  const isDeployedOnChain = useRootSelector(isDeployedOnChainSelector);
+  if (isArchanovaAccountActive && chain !== CHAIN.ETHEREUM) {
+    errorMessage = t('error.walletConnect.activeAccountDoesNotSupportSelectedChain');
+  }
+
   const estimationErrorMessage = useRootSelector((root) => root.transactionEstimate.errorMessage);
 
+  /**
+   * Archanova account needs to be deployed regardless of the action,
+   * Etherspot account must to be deployed for signature requests only
```
mdjastrzebski, line 188:
```suggestion
   * Archanova account needs to be deployed for all call requests.
   * Etherspot account doesn't need to be deployed for transaction requests.
```

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
```diff
@@ -132,22 +138,24 @@ const useTransactionFee = (request: WalletConnectCallRequest) => {
...
 
-  const balances = useRootSelector(accountEthereumWalletAssetsBalancesSelector);
-  const { amount, symbol, decimals } = useTransactionPayload(request);
-
   const chain = chainFromChainId[request.chainId];
   if (!chain && !estimationErrorMessage) {
     estimationErrorMessage = t('error.walletConnect.cannotDetermineEthereumChain');
   }
 
+  const accountAssetsBalances = useRootSelector(accountAssetsBalancesSelector);
```
mdjastrzebski, line 146: Not related to this particular line. I thing we should put `TransactionDeploymentWarning` here to signal that accepting this transaction will deploy your wallet.

mdjastrzebski (changes requested):
Looks like there are still some issues with logical conditions atm.

src/screens/WalletConnect/CallRequest/SignatureRequestContent.js
poocart: Whoops 👀

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
poocart: This is signature request so I would leave signature keyword here:

```
 /**
   * Archanova account needs to be deployed for all types call requests.
   * Etherspot account needs to be deployed for signature type call requests only.
   */
```

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
poocart: Additionally added Archanova account check to `TransactionDeploymentWarning` since we shouldn't show this warning for Archanova accounts.

src/actions/walletConnectActions.js
poocart: IMO at least in this case it's unnecessary since the button is disabled and writing it in action encourages others to do so in first place while we should show this in UI before user does any action and block him from going further. If in similar flows it reaches this kind of point – we should fix that user is blocked before he comes to action.

mdjastrzebski (changes requested):
`requiresDeployedAccount` still seem to need some tweaking, pls check comment.

src/screens/WalletConnect/CallRequest/SignatureRequestContent.js
```diff
@@ -47,10 +54,19 @@ type Props = {|
...
 
   const { title, iconUrl, chain } = getViewData(request);
   const config = configs[chain];
 
+  /**
+   * Archanova account needs to be deployed for all types call requests.
+   * Etherspot account needs to be deployed for signature type call requests only.
+   */
+  const requiresDeployedAccount = (isArchanovaAccount(activeAccount) && !isArchanovaAccountDeployed)
+    || !isEtherspotAccountDeployed(activeAccount, chain);
```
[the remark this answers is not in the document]
poocart: Not really, this method has `if (!isEtherspotAccount(account)) return false;`

mdjastrzebski (approved):
Looks good 🎸

PR #2717, opened 2021-06-15
PIL-1070: Local fiat currency changing doesn't work

mdjastrzebski:
Scope:
- [x] Support fiat currencies for total balances (by memoized selectors)
- [x] Support fiat currencies for asset balances (by in component handling)
- [x] Some minor refactor regarding to totalBalances type naming
- [x] Strongly typed Currency usage

### DIFF
--- a/src/models/TotalBalances.js
+++ b/src/models/TotalBalances.js
@@ -20,41 +20,46 @@
 
 import { BigNumber } from 'bignumber.js';
 
-// types
+// Types
 import type { AccountRecord } from 'models/Account';
+import type { AssetCategoryRecord } from 'models/AssetCategory';
 import type { ChainRecord } from 'models/Chain';
 
 /**
  * Total balances are represented in top-down nesting:
  *  - level 1. Account
- *  - level 2. Category
+ *  - level 2. Asset Category
  *  - level 3. Chain
  *
  * This hierarchy represents the order in which they are most frequently used.
  */
 
-// Category balances are calculated using selectors in redux
-export type CategoryRecord<T> = {
-  wallet?: T,
-  deposits?: T,
-  investments?: T,
-  liquidityPools?: T,
-  rewards?: T,
-};
+/** Generic helper type for Account -> Category -> Chain nesting. */
+export type AccountCategoryChainRecord<T> = AccountRecord<AssetCategoryRecord<ChainRecord<T>>>;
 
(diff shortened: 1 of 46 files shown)


src/models/TotalBalances.js
```diff
@@ -20,41 +20,46 @@
...
-  rewards?: T,
-};
-
-export type StoreTotalBalances = AccountRecord<?StoreAccountTotalBalances>;
-export type StoreAccountTotalBalances = StoreCategoryRecord<ChainRecord<BigNumber>>;
-
-// Wallet balances are calculated using selectors in redux
-export type WalletTotalBalances = AccountRecord<ChainRecord<BigNumber>>;
+/** Generic helper type for Account -> Category -> Chain nesting. */
+export type AccountCategoryChainRecord<T> = AccountRecord<AssetCategoryRecord<ChainRecord<T>>>;
```
poocart, line 38: 😱

mdjastrzebski: Yup, this is our actual data structure, only renamed it for clarity

src/utils/rates.js
```diff
@@ -0,0 +1,42 @@
...
+  const ethRates = rates[ETH];
+  if (!ethRates || !ethRates[currency] || !ethRates[USD]) {
+    return null;
+  }
+
+  return ethRates[currency] / ethRates[USD];
+}
+
+export function getFiatValueFromUsd(valueInUsd: ?BigNumber | string, usdToFiatRate: ?number): ?BigNumber {
+  return valueInUsd && usdToFiatRate != null ? BigNumber(valueInUsd).times(usdToFiatRate) : null;
```
poocart, line 41: Isn't it BigNumber already?

```suggestion
  return valueInUsd && usdToFiatRate != null ? valueInUsd.times(usdToFiatRate) : null;
```

poocart (approved):
LGTM 🚀

src/utils/rates.js
mdjastrzebski: When coming back from redux restore it can be string

poocart: Nevermind, maybe having it wrapped here makes sense, but maybe we can use `wrapBigNumber` instead, wdyt?

mdjastrzebski: Applied

PR #2729, opened 2021-06-18
multi chain asset rates wip

poocart:
Includes:
- Added multi chain asset price rates.
- Removed Cryptocompare per discussion with @IAmKio.


src/models/RatesByAssetSymbol.js
```diff
@@ -0,0 +1,36 @@
...
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+// constants
+import { ETH } from 'constants/assetsConstants';
+
+// types
+import type { ChainRecord } from 'models/Chain';
+import type { Currency } from 'models/Asset';
```
mdjastrzebski, line 26: Consider moving Currency to this file, as it seems related more to rates than assets.

```suggestion
import type { Currency } from 'models/Asset';
```

src/models/RatesByAssetSymbol.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+/*
+    Pillar Wallet: the personal data locker
```
mdjastrzebski, line 3: Consider renaming file to `Rates.js` as we will probably put different rate-related models here.

src/models/RatesByAssetSymbol.js
```diff
@@ -0,0 +1,36 @@
...
+import { ETH } from 'constants/assetsConstants';
+
+// types
+import type { ChainRecord } from 'models/Chain';
+import type { Currency } from 'models/Asset';
+
+
+export type RateByCurrencySymbol = { [currencySymbol: string]: number };
+
+export type RatesByAssetSymbol = { [assetSymbol: string]: RateByCurrencySymbol };
```
mdjastrzebski, line 31: Consider simplifying name, as `symbol` in our project normally means asset symbol.

```suggestion
export type RatesBySymbol = { [assetSymbol: string]: RateByCurrencySymbol };
```

src/models/RatesByAssetSymbol.js
```diff
@@ -0,0 +1,36 @@
...
+
+// constants
+import { ETH } from 'constants/assetsConstants';
+
+// types
+import type { ChainRecord } from 'models/Chain';
+import type { Currency } from 'models/Asset';
+
+
+export type RateByCurrencySymbol = { [currencySymbol: string]: number };
```
mdjastrzebski, line 29: Consider changing key name, as `currencySymbol` is not exact, because we have USD, EUR, GBP & ETH, so last one is not exactly currency symbol.

```suggestion
// TODO: refactor rateKey to be of RateKey type.
export type Rates = { [rateKey: string]: number };
```

mdjastrzebski: Consider moving `export type RateKey = Currency | typeof ETH;` next to this type because they are linked.

src/reducers/ratesReducer.js
```diff
@@ -17,17 +17,17 @@
...
+
 
 export type RatesReducerState = {
-  data: Rates,
-  isFetched: boolean,
+  data: RatesPerChain,
+  isFetching: boolean,
 };
 
 export type RatesReducerAction = {
```
mdjastrzebski, line 33: Pls provide typed actions & `RatesReducerAction` as a union type

src/reducers/ratesReducer.js
```diff
@@ -36,17 +36,29 @@ export type RatesReducerAction = {
...
 };
 
 const ratesReducer = (
   state: RatesReducerState = initialState,
   action: RatesReducerAction,
 ): RatesReducerState => {
   switch (action.type) {
-    case UPDATE_RATES:
-      return { ...state, data: action.payload, isFetched: true };
+    case SET_RATES:
```
mdjastrzebski, line 48: Seems there is no case for `SET_FETCHING_RATES`.

src/actions/ratesActions.js
```diff
@@ -35,54 +38,74 @@ import { reportErrorLog } from 'utils/common';
...
-      reportErrorLog('fetchAssetsRatesAction failed: allAccountsCrossChainAssets is empty');
-      return;
-    }
+        await dispatch(setRatesAction(chain, rates));
+      }));
+    }));
 
-    const rates = await getExchangeRates(allAccountsCrossChainAssets);
-    dispatch(setRatesAction(rates));
+    dispatch(setIsFetchingRatesAction(false));
```
mdjastrzebski, line 108: Would we need some `try-catch` to make sure we reset isFetching flag?

src/actions/ratesActions.js
```diff
@@ -91,24 +114,33 @@ export const fetchSingleChainAssetRatesAction = (
...
 
-    const rates = await getExchangeRates({ [asset.symbol]: asset });
+    const rates = await getExchangeRates(chain, { [asset.symbol]: asset });
 
     if (isEmpty(rates)) return;
 
-    const { rates: { data: currentRates } } = getState();
-    const updatedRates = { ...currentRates, ...rates };
-    dispatch(setRatesAction(updatedRates));
+    await dispatch(setRatesAction(chain, rates));
```
mdjastrzebski, line 142: This looks like it would override record (for given chain) with various rates, with a new record with only one rate. Consider adding ```const updatedRates = { ...currentChainRates, ...rates };``` logic to ratesReducer.

mdjastrzebski (commented):
Added some wip comments for models/reducer/actions.

src/models/RatesByAssetSymbol.js
poocart: Yeah, the rename was picked due refactor.

src/actions/ratesActions.js
poocart: Actually I think this shouldn't be spread at all as we're setting, not updating, fixed.

src/models/Rates.js
```diff
@@ -0,0 +1,42 @@
...
+  EUR,
+  GBP,
+  USD,
+} from 'constants/assetsConstants';
+
+// types
+import type { ChainRecord } from 'models/Chain';
+
+
+export type RateBySymbol = { [symbol: string]: number };
```
mdjastrzebski, line 33: 1. `RateBySymbol` & `RatesBySymbol` seems to be too closely named. 
2. Pls move RateKey next to `RateBySymbol`, as RateKey is intended to be key for rates entry (after fixing type issues with codebase).
 
```suggestion
export type RateKey = Currency | typeof ETH;

// TODO: migrate to strongly typed keys: USD, EUR, GBP, ETH.
export type RateEntry = { [key: string]: number };
```

src/screens/Assets/wallet/WalletListItem.js
```diff
@@ -29,30 +29,42 @@ import FiatChangeView from 'components/modern/FiatChangeView';
...
+  value,
+  change,
+  symbol,
+  onPress,
+  chain,
+}: Props) {
   const colors = useThemeColors();
 
-  const rates = useRates();
+  const chainRates = useChainRates(chain);
```
mdjastrzebski, line 64: Consider creating simplier hook
```suggestion
  const rate = useTokenRate(chain, symbol); // expressed in user currency.
```

src/actions/ratesActions.js
```diff
@@ -35,54 +38,72 @@ import { reportErrorLog } from 'utils/common';
...
+export const setRatesAction = (
+  chain: Chain,
+  rates: RatesBySymbol,
+) => {
+  return async (dispatch: Dispatch, getState: GetState) => {
+    if (isEmpty(rates)) return;
+
+    dispatch({ type: SET_CHAIN_RATES, payload: { chain, rates } });
+
+    const updatedRatesPerChain = ratesPerChainSelector(getState());
```
mdjastrzebski, line 70: Pls use raw redux selection here, as using selector could cause serialization problems if selector is changed in the future, which will work for components but not for saveDb

```suggestion
    const updatedRatesPerChain = getState().rates....
```

src/actions/ratesActions.js
```diff
@@ -35,54 +38,72 @@ import { reportErrorLog } from 'utils/common';
...
 
-      return { ...combinedAssets, ...accountAssets };
-    }, {});
+    dispatch(setIsFetchingRatesAction(true));
 
-    if (isEmpty(allAccountsCrossChainAssets)) {
-      reportErrorLog('fetchAssetsRatesAction failed: allAccountsCrossChainAssets is empty');
-      return;
-    }
+    const assetsBalances = assetsBalancesSelector(getState());
```
mdjastrzebski, line 86:
```suggestion
    const assetsBalancesPerAccount = assetsBalancesSelector(getState());
```

src/actions/ratesActions.js
```diff
@@ -35,54 +38,72 @@ import { reportErrorLog } from 'utils/common';
...
+    const supportedAssets = supportedAssetsPerChainSelector(getState());
```
mdjastrzebski, line 87:
```suggestion
    const supportedAssetsPerChain = supportedAssetsPerChainSelector(getState());
```

src/actions/ratesActions.js
```diff
@@ -35,54 +38,72 @@ import { reportErrorLog } from 'utils/common';
...
+      const accountAssetsBalances = assetsBalances[accountId] ?? {};
+      await Promise.all(Object.keys(accountAssetsBalances).map(async (chain) => {
+        try {
+          const chainSupportedAssets = supportedAssets[chain] ?? [];
+          const walletAssets = accountAssetsBalances[chain]?.wallet ?? {};
+
+          const accountAssets = mapWalletAssetsBalancesIntoAssetsBySymbol(walletAssets, chainSupportedAssets);
+          const rates = await getExchangeRates(chain, accountAssets);
+
+          await dispatch(setRatesAction(chain, rates));
```
mdjastrzebski, line 99: ❗ Seems like we're replacing previous chain rates for user assets. This will cause problems if user has 2 accounts: 1st with assets A, B, C, 2nd with assets A, D, E. The last of promises to finish will set its chain rates resulting either in A, B, C rates or A, D, E while redux should contain all A, B, C, D, E.

Consider improving the logic:
```
for each chain:
  get all assets for given chain across accounts
  query exchange rates for that chain & all accounts assets
  set redux state

This should result in fewer network calls. and all needed rates in redux store.

```

src/actions/ratesActions.js
mdjastrzebski: ❗ 
This will overwrite all rates for given chain, with only one rate. Eg. store had rates for A, B, C but here we fetch rates for A and using current version of `setRates` will result in store no longer having rates for B, C.

src/reducers/ratesReducer.js
```diff
@@ -17,36 +17,77 @@
...
+        data: action.payload,
+      };
+
+    case SET_CHAIN_RATES:
+      const { chain, rates } = action.payload;
+      return {
+        ...state,
+        data: {
+          ...state.data,
+          [chain]: rates,
```
mdjastrzebski, line 81: The way in other code uses `SET_CHAIN_RATES` suggests that this action should apply new rates over the previous ones.  Also consider renaming to `UPDATE_CHAIN_RATES` to reflact that fact.

```suggestion
          [chain]: { ...state.data?.[chain], ...rates},
```

src/services/coinGecko.js
```diff
@@ -53,10 +67,26 @@ const requestConfig = {
 
 const currenciesParam = rateKeys.map(key => key.toLocaleString()).join(',');
 
+/* eslint-disable i18next/no-literal-string */
+export const nativeAssetSymbolToCoinGeckoCoinId = {
```
mdjastrzebski, line 71: Consider keying it by chain, instead of asset symbol.

src/services/assets.js
```diff
@@ -301,85 +300,39 @@ export function fetchRinkebyETHBalance(walletAddress: Address): Promise<string>
...
-      reportErrorLog('getLegacyExchangeRates cryptocompare.cryptocompare failed', { assets });
-      return null;
-    });
-}
-
-export async function getExchangeRates(assets: AssetsBySymbol): Promise<?Object> {
+export async function getExchangeRates(
+  chain: Chain,
+  assets: AssetsBySymbol,
+): Promise<?Object> {
```
mdjastrzebski, line 306: Pls return proper promise type

mdjastrzebski (commented):
Overall a great refactor.

Found some issues with SET_CHAIN_RATES when new value overwrites the previous ones, and potentially resulting in removal of recent & valid rate entries.

Also suggested better naming between `RatesBySymbol` & `RateBySymbol` as the current naming is too similar.

src/screens/Assets/wallet/WalletListItem.js
poocart: Mate, tbh given the time we have I would consider this as future improvement as there are more important issues piling up 👀  hope it makes sense

src/actions/ratesActions.js
poocart: This has changed by replacing rates from setting to updating flow.

poocart: Combined requested assets tho.

src/models/Rates.js
poocart: Moving currency inside causes more flow typing issues which takes time to solve, leaving as `string` for now, hope it makes sense.

mdjastrzebski (approved):
Ship it mate! 🚢

PR #2736, opened 2021-06-22
PIL-1069: fix double counting of service balaces

mdjastrzebski:
Scope:
- [x] Exclude asset from wallet total balance if it already has service balance
- [x] Naming cleanup around findAsset & balance types


poocart (approved):
LGTM

src/actions/assetsActions.js
```diff
@@ -449,6 +450,7 @@ export const fetchAllAccountsTotalBalancesAction = () => {
 
                 return {
                   key: `${protocol}-${symbol}`,
+                  address,
```
poocart, line 453: Will need to ask to add this to recent work for Etherspot totals.

PR #2738, opened 2021-06-23
PIL-1094: replaced zapper to etherspot totals

poocart:
https://linear.app/pillarproject/issue/PIL-1094/wire-up-account-totals

Replaces Zapper totals integration to Etherspot.

Notes:
- Need to add `address` per https://github.com/pillarwallet/pillarwallet/pull/2736 changes.
- SDK to be updated to latest version before merging (PR is built on expected SDK method result on new SDK version).


mdjastrzebski:
Blocked until etherspot exposes `address` field for protocol balances.

src/actions/assetsActions.js
```diff
@@ -383,112 +376,62 @@ export const fetchAllAccountsTotalBalancesAction = () => {
...
+      if (!accountTotalBalances) return;
+
+      accountTotalBalances.forEach(({
+        balances,
+        category: balancesCategory,
+        chainId,
+        totalBalance,
+      }) => {
+        const chain = chainFromChainId[chainId];
+        const assetsCategory = assetsCategoryFromEtherspotBalancesCategory[balancesCategory];
```
mdjastrzebski, line 396: Pls check for unexpected balanacesCategory that would return undefined assetsCategory.

mdjastrzebski (commented):
Looks promising :-) added some initial comments until actual method implemented.

src/actions/assetsActions.js
poocart: There won't be as it's directly mapped, please check Etherspot `AccountDashboardProtocols` schema (https://docs.etherspot.dev/enums/accountdashboardprotocols.html).

Will still add a check tho just in case.

src/actions/assetsActions.js
```diff
@@ -383,112 +376,62 @@ export const fetchAllAccountsTotalBalancesAction = () => {
...
-            zapperProtocolId,
-            zapperNetworkId,
-          );
-
-          const balanceCategory = mapZapperProtocolIdToBalanceCategory(zapperProtocolId);
-          if (!protocolBalancesByAccounts || !balanceCategory) return null;
-
-          return { category: balanceCategory, balances: protocolBalancesByAccounts };
+      // we're fetching and storing values in USD and converting rates by app selected currency later
+      const accountTotalBalances = await etherspotService.getAccountTotalBalances(accountAddress, USD);
```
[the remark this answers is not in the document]
poocart: EtherspotAccountTotalBalancesItem[]` is already defined as return type `getAccountTotalBalances` method. Do you want me to extract it separately into `utils/types/etherspot.js`? Doesn't make sense 🤔

src/actions/assetsActions.js
mdjastrzebski: We're not mapping rewards, so if backend update causes it to return rewards we will get `undefined` here. Same if for some reason they will add any other category in the future, which could affect existing clients.

IAmKio (approved):
LGTM

PR #2739, opened 2021-06-23
do not wait for history fetch during onboarding

poocart:
When we onboard users we wait until history is fetched which sometimes increases onboarding time by 30-40s (I think it has cold boot / throttle on the 3rd party API we’re using on Etherspot or else, will check with Etehrspot team and provide you an update).

Based on these changes we still dispatch history fetch action, but do not wait for result and let user go through onboarding, history will still appear once it’s fetched.

### DIFF
--- a/src/actions/__tests__/onboardingActions.test.js
+++ b/src/actions/__tests__/onboardingActions.test.js
@@ -250,8 +250,8 @@ describe('Onboarding actions', () => {
       { payload: false, type: SET_FETCHING_TOTAL_BALANCES },
 
       { type: SET_FETCHING_HISTORY, payload: true },
-      { type: SET_FETCHING_HISTORY, payload: false },
       { type: SET_FETCHING_RATES, payload: true },
+      { type: SET_FETCHING_HISTORY, payload: false },
 
       { type: UPDATE_CHAIN_RATES, payload: { chain: CHAIN.ETHEREUM, rates: mockExchangeRates } },
       { type: SET_FETCHING_RATES, payload: false },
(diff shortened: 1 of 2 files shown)


src/actions/__tests__/onboardingActions.test.js
```diff
@@ -303,9 +303,8 @@ describe('Onboarding actions', () => {
       { payload: false, type: SET_FETCHING_TOTAL_BALANCES },
 
       { type: SET_FETCHING_HISTORY, payload: true },
-      { type: SET_HISTORY, payload: { [mockArchanovaAccount.id]: { ethereum: [] } } },
-      { type: SET_FETCHING_HISTORY, payload: false },
       { type: SET_FETCHING_RATES, payload: true },
+      { type: SET_HISTORY, payload: { [mockArchanovaAccount.id]: { ethereum: [] } } },
```
mdjastrzebski, line 307: Why is there no matching `SET_FETCHING_HISTORY false`?

poocart: Because there is a promise inside an action and action is not waited therefore race condition happens on test itself and it's never updated. We have similar cases cross all tests.

mdjastrzebski: Ok, just wanted to bring your attention to it.

mdjastrzebski (approved):
Looks good.

Can we merge it directly to `develop` instead of `fix/zapper-to-etherspot-totals`?

IAmKio (approved):

src/actions/__tests__/onboardingActions.test.js
poocart: I'm aware about this, thanks.

PR #2741, opened 2021-06-23
remove some zapper leftovers

poocart:
Removes some Zapper relations that left un-removed with previous PR.


mdjastrzebski (approved):
LGTM

PR #2745, opened 2021-06-24
PIL-1089: onboarding ens name fixes

poocart:
https://linear.app/pillarproject/issue/PIL-1089/user-tried-importing-choose-a-username-screen-appears

Includes:
- Added new Etherspot SDK method `validateENSName` that checks provided username for being invalid/blacklisted/taken.
- Added missing reserved ENS name check.
- Fixed previously unnoticed change of incorrect Etherspot SDK method rename.


src/services/etherspot.js
```diff
@@ -292,7 +292,7 @@ export class EtherspotService {
 
   reserveEnsName(username: string): Promise<?ENSNode> {
     const fullEnsName = getEnsName(username);
-    return this.sdk.getAccountTokenListTokens({ name: fullEnsName }).catch((error) => {
+    return this.sdk.reserveENSName({ name: fullEnsName }).catch((error) => {
```
poocart, line 295: This is silly :)

IAmKio: Copy pasta?

mdjastrzebski: 😂

src/services/onboarding.js
```diff
@@ -81,6 +99,10 @@ export const isUsernameTaken = async (username: string): Promise<boolean> => {
 
   try {
     const ensName = getEnsName(username);
+
+    const isValid = await etherspotService.isValidEnsName(ensName);
+    if (!isValid) return true; // invalid/blacklisted/taken
```
mdjastrzebski, line 104:
```suggestion
    if (!isValid) return true; // invalid/blacklisted/taken/error
```

IAmKio (approved):
LGTM

src/services/etherspot.js
```diff
@@ -305,6 +305,14 @@ export class EtherspotService {
     });
   }
 
+  isValidEnsName(name: string): Promise<boolean> | boolean {
+    try {
+      return this.sdk.validateENSName({ name });
+    } catch (error) {
+      return false; // no need to report
```
mdjastrzebski, line 312: I would consider throwing here, as failed network call will be reported as failure.

mdjastrzebski (approved):
Looks goods,

I would consider indicating error when isValidEnsName fails rather than silently treating it the same as reserved.

src/services/etherspot.js
poocart: I noticed that during last week this is second time this happens 👀  hot weather? rest? 🌴

PR #2750, opened 2021-06-24
PIL-1099: replace Collectible(s) text with NFT(s)

mdjastrzebski:


src/models/Collectible.js
```diff
@@ -57,9 +57,6 @@ export type CollectibleTransaction = {
   _id: string,
 };
 
-
-export type Collectibles = Collectible[];
-
 export type CollectiblesStore = {
```
poocart, line 60: Maybe we can proceed with general renaming to `NFT` including types and etc.? As if someone else steps in they might continue on `collectible` narrative naturally. Wdyt?

IAmKio (approved):

PR #2754, opened 2021-06-29
PIL-1102: restore lost wallet connect connectors

poocart:
https://linear.app/pillarproject/issue/PIL-1102/walletconnect-connection-initiation


IAmKio (approved):
Great work mate 🎉

PR #2756, opened 2021-06-30
PIL-1104: Fix/nft fixes

poocart:
https://linear.app/pillarproject/issue/PIL-1104/uniswap-nfts-do-not-display-properly

Additionally this fixes some issues on NFT send flow.

### DIFF
--- a/src/components/ReviewSummary/CollectibleReviewSummary.js
+++ b/src/components/ReviewSummary/CollectibleReviewSummary.js
@@ -43,7 +43,7 @@ const CollectibleReviewSummary = ({ collectible, text }: Props) => {
   return (
     <Container>
       <CollectibleImageWrapper>
-        <CollectibleImage source={{ uri: collectible.imageUrl }} width={192} heigh={192} />
+        <CollectibleImage source={{ uri: collectible.imageUrl }} width={192} height={192} />
       </CollectibleImageWrapper>
       <Spacing h={16} />
       <BaseText regular>{text}</BaseText>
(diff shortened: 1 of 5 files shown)


src/components/ReviewSummary/CollectibleReviewSummary.js
```diff
@@ -34,15 +34,17 @@ const Container = styled.View`
...
   width: 192px;
   height: 192px;
 `;
 
 const CollectibleReviewSummary = ({ collectible, text }: Props) => {
   return (
     <Container>
-      <StyledCollectibleImage source={{ uri: collectible.imageUrl }} />
+      <CollectibleImageWrapper>
+        <CollectibleImage source={{ uri: collectible.imageUrl }} width={192} height={192} />
```
IAmKio, line 46: If we're defining a  `CollectibleImageWrapper` here - but then defining `width={192} height={192}` - would `CollectibleImage` not expand? Just seems off that we have two elements both with the same height / width definitions...

IAmKio (approved):
Just a few questions but LGTM!

src/components/ReviewSummary/CollectibleReviewSummary.js
poocart: The problem was that SVG images were taking whole screen since it didn't have any wrapper around and we're not setting fixed dimensions on them. With this we have a fixed wrapper and just passing same dimensions to image that might be SVG or not – in case it's not SVG then it takes the passed dimensions to match wrapper. Similar implementation is on Send screen as well.

IAmKio: Got it - cheers!

PR #2757, opened 2021-06-30
added mutli chain nfts structure

poocart:
Added support for multi chain NFTs. No service hooked yet for other chains than Ethereum mainnet.


src/actions/assetsActions.js
```diff
@@ -125,11 +124,11 @@ export const sendAssetAction = (
     const accountId = getAccountId(activeAccount);
     const accountAddress = getAccountAddress(activeAccount);
 
-    const accountCollectibles = collectibles[accountId] || [];
+    const accountCollectiblesOnChain = accountCollectiblesSelector(getState())?.[chain] ?? [];
```
IAmKio, line 127:
```suggestion
    const accountCollectiblesOnChain = accountCollectiblesSelector(getState())?.[chain] ?? [];
```

Just curious here - are we evaluating the output of getState() as a whole to be truthy or falsey here?

src/reducers/collectiblesReducer.js
```diff
@@ -82,12 +86,24 @@ const collectiblesReducer = (
...
+
   switch (action.type) {
-    case UPDATE_COLLECTIBLES:
+    case SET_COLLECTIBLES:
       return {
         ...state,
         data: action.payload || {},
       };
+    case SET_ACCOUNT_COLLECTIBLES:
+      ({ accountId } = action.payload);
```
IAmKio, line 98: Do we need the parenthesis here?

src/reducers/collectiblesReducer.js
```diff
@@ -100,8 +116,8 @@ const collectiblesReducer = (
         updatingTransaction: action.payload,
       };
     case ADD_COLLECTIBLE_HISTORY_TRANSACTION:
-      const { accountId } = action.payload;
-      const accountCollectibles = state.data[accountId] ?? [];
+      ({ accountId } = action.payload);
```
IAmKio, line 119: Ahh i can see what you're achieving here

src/services/dataMigration/collectibles.js
```diff
@@ -1,9 +1,48 @@
...
+
+// actions
+import { saveDbAction } from 'actions/dbActions';
+
+// types
+import type { Dispatch } from 'reducers/rootReducer';
+
+
+export default function (storageData: Object, dispatch: Dispatch) {
+  const collectibles = storageData?.collectibles?.collectibles ?? {}; // not a mistype
```
IAmKio, line 29: 😂 NFT inception

IAmKio (approved):
Just a few questions but all good!

src/reducers/collectiblesReducer.js
poocart: Yeah, either we need to kill that eslint rule or keep applying like this 👀

src/actions/assetsActions.js
poocart: In case the output of `accountCollectiblesSelector` is falsey then it falls back to array

IAmKio: Got it - cheers!

PR #2759, opened 2021-06-30
added poap nfts on xdai

poocart:
Per title.

### DIFF
--- a/src/services/poap.js
+++ b/src/services/poap.js
@@ -25,6 +25,7 @@ import { callSubgraph } from 'services/theGraph';
 
 // utils
 import httpRequest from 'utils/httpRequest';
+import { reportErrorLog } from 'utils/common';
 
 // abis
 import ERC721_CONTRACT_ABI from 'abi/erc721.json';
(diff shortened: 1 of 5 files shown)


src/services/poap.js
```diff
@@ -0,0 +1,103 @@
...
+
+// types
+import type { Collectible } from 'models/Collectible';
+
+
+/* eslint-disable i18next/no-literal-string */
+const xDaiRpcUrl = 'https://dai.poa.network/';
+const poapContractAddress = '0x22c1f6050e56d2876009903609a2cc3fef83b415';
+const poapSubgraphName = 'poap-xyz/poap-xdai';
+/* eslint-enable i18next/no-literal-string */
```
IAmKio, line 44: I know this is temporary before we move to the backend so i'm happy to keep this here for now

src/services/poap.js
```diff
@@ -0,0 +1,103 @@
...
+
+  const tokensWithMetadata = await Promise.all(tokenIds.map(async (tokenId) => {
+    const tokenMetadataUri = await collectibleContract.tokenURI(tokenId).catch(() => null);
+    if (!tokenMetadataUri) return null;
+
+    try {
+      const { data } = await httpRequest.get(tokenMetadataUri);
+      return { ...data, tokenId };
+    } catch (error) {
+      //
```
IAmKio, line 79: Shall we log and report something here?

IAmKio (changes requested):
Just a small point re: error logging but apart from that - all good!

src/services/poap.js
poocart: Exactly ❤️  didn't want to pollute configs/constants

src/services/poap.js
poocart: Added

IAmKio (approved):

PR #2763, opened 2021-07-02
claim etherspot ens name tx

poocart:
Puts Etherspot ENS claim transaction in any transaction batch that is estimated/sent.

Before put it's checked if the ENS is needed to claim.

~Few notes that needed to be fixed on Etherspot SDK/back-end before this gets merged:~
1. ~`validateENSName` should return invalid (throw exception) for reserved names (i.e. ardd.pillar.eth)~
2. ~`reserveENSName` should not let reserve on occupied ENS names~
3. ~after ENS name is claimed getENSNode returns updated ENS name state (from Reserved to Claimed) only after SDK is reinitiated~

Notes above are fixed. 3rd issue tends to be just a longer delay.

### DIFF
--- a/src/actions/ensRegistryActions.js
+++ b/src/actions/ensRegistryActions.js
@@ -36,7 +36,7 @@ import {
   reportErrorLog,
   resolveEnsName,
 } from 'utils/common';
-import { findFirstArchanovaAccount } from 'utils/accounts';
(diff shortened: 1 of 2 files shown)


poocart:
Etherspot SDK notes were fixed. 3rd issue tends to be just a longer delay.

IAmKio (approved):
LGTM!

PR #2766, opened 2021-07-07
add native asset on supported empty list

poocart:
Per title.

Returning `null` causes empty supported assets list which affects further flows such as fetching assets balances.


divya-svmx (approved):
lgtm

IAmKio (approved):

PR #2770, opened 2021-07-09
PIL-1101: Fix/assets by contract address

poocart:
https://linear.app/pillarproject/issue/PIL-1101/exchange-rate-reported-on-app-is-incorrect

Includes:
- [x] Asset by asset contract address.
- [x] Wallet asset balance by contract address.
- [x] Asset rates by asset contract address (https://github.com/pillarwallet/pillarwallet/pull/2772).
- [x] Asset rates and asset balances by lowercase asset contract address (rates by asset contract address added assets and rates by lowercase contract address #2773).


mdjastrzebski (commented):
Added some initial comments

src/actions/etherspotActions.js
```diff
@@ -328,10 +328,10 @@ const handleGatewayBatchUpdatedNotification = async (
   const supportedAssetsPerChain = supportedAssetsPerChainSelector(getState());
   const chainSupportedAssets = supportedAssetsPerChain[chain] ?? [];
 
-  const assetData = getAssetData(
+  const assetData = findAsset(
```
mdjastrzebski, line 331: passing chainAccountAssets seems redundant as they are taken from supportedAssets anyway

src/actions/transactionEstimateActions.js
```diff
@@ -146,8 +149,12 @@ export const estimateTransactionsAction = (
     const supportedAssetsPerChain = supportedAssetsPerChainSelector(getState());
     const chainSupportedAssets = supportedAssetsPerChain[chain] ?? [];
 
+    const gasTokenAddress = preferredGasTokenSelector(getState()) === PLR
```
mdjastrzebski, line 152: Seems like all non PLR values lead to using native asset. Is that intened?

mdjastrzebski: Q2: is PLR availalbe on all chains? maybe we need to default to nativeAsset in case there is no PLR for given side chain

src/components/ActivityFeed/ActivityFeedItem.js
```diff
@@ -361,7 +362,9 @@ export class ActivityFeedItem extends React.Component<Props> {
     }
     const relevantAddress = this.getRelevantAddress(event);
 
-    const assetSymbol = event ? event.asset : null;
+    // this component is Archanova (Ethereum) only
+    const { assetSymbol = nativeAssetPerChain.ethereum.symbol } = event ?? {};
```
mdjastrzebski, line 366: Consider
```suggestion
    const assetSymbol = event?.assetSymbol ?? nativeAssetPerChain.ethereum.symbol;
```

src/components/EventDetails/EventDetails.js
```diff
@@ -431,14 +431,15 @@ export class EventDetail extends React.Component<Props> {
     return isReceived ? event.from : event.to;
   };
 
-  getFormattedGasFee = (formattedFee: number, token: string) => {
+  getFormattedGasFee = (formattedFee: number) => {
```
mdjastrzebski, line 434: why is number argument named formattedFee? Probably should be fee.

src/components/EventDetails/EventDetails.js
```diff
@@ -431,14 +431,15 @@ export class EventDetail extends React.Component<Props> {
...
     const { baseFiatCurrency, ratesPerChain } = this.props;
     const ethereumRates = ratesPerChain[CHAIN.ETHEREUM] ?? {};
     const fiatCurrency = baseFiatCurrency || defaultFiatCurrency;
-    const rate = getRate(ethereumRates, token, fiatCurrency);
+    const { symbol, address } = nativeAssetPerChain.ethereum;
+    const rate = getAssetRateInFiat(ethereumRates, address, fiatCurrency);
```
mdjastrzebski, line 439: Potentially rate could be nil

src/components/FeeLabelToggle/FeeLabelToggle.js
```diff
@@ -94,8 +94,8 @@ const FeeLabelToggle = ({
   const feeValue = getFormattedTransactionFeeValue(chain, txFeeInWei, gasToken);
   const currencySymbol = getCurrencySymbol(fiatCurrency);
 
-  const gasSymbol = getGasSymbol(chain, gasToken);
-  const feeInFiat = parseFloat(feeValue) * getRate(chainRates, gasSymbol, fiatCurrency);
+  const gasAddress = gasToken?.address || nativeAssetPerChain[chain].address;
```
mdjastrzebski, line 97: Maybe we could create some helper `function getGasAddress(gasToken, chain)` that would centralise it

src/components/FeeLabelToggle/FeeLabelToggle.js
```diff
@@ -94,8 +94,8 @@ const FeeLabelToggle = ({
...
+  const feeInFiat = parseFloat(feeValue) * getAssetRateInFiat(chainRates, gasAddress, fiatCurrency);
```
mdjastrzebski, line 98: Similar suggestion here, pls create a helper `function getFiatValue(balance, rate)` that will handle edge cases like NaN and nil rates.

src/components/HistoryEventDetails/layouts/EnsNameEventDetails.js
```diff
@@ -65,7 +65,15 @@ function EnsNameEventDetails({ event, chain }: Props) {
       <Text variant="large">{t('label.registered')}</Text>
       <Spacing h={spacing.extraLarge} />
 
-      {!!event?.fee && <FeeLabel value={event.fee.value} symbol={event.fee.symbol} mode="actual" chain={chain} />}
+      {!!event?.fee && (
+        <FeeLabel
+          value={event.fee.value}
+          assetSymbol={event.fee.symbol}
```
mdjastrzebski, line 71: IMO would drop assetSymbol prop, and ask FeeLabel do find it from supported assets. That would make FeeLabel easier to use.

src/components/Table/Table.js
```diff
@@ -130,10 +130,11 @@ export const TableUser = ({ ensName, address }: TableUserProps) => {
 
 export const TableFee = ({ txFeeInWei, gasToken, chain }: TableFeeProps) => {
   const decimals = gasToken?.decimals || nativeAssetPerChain[chain].decimals;
+  const gasTokenAddress = gasToken?.address || nativeAssetPerChain[chain].address;
   const formattedFee = txFeeInWei ? formatUnits(txFeeInWei.toString(), decimals) : '0';
   const gasSymbol = getGasSymbol(chain, gasToken);
 
-  return <TableAmount amount={formattedFee} token={gasSymbol} chain={chain} />;
+  return <TableAmount amount={formattedFee} assetSymbol={gasSymbol} assetAddress={gasTokenAddress} chain={chain} />;
```
mdjastrzebski, line 137: Consider dropping assetSymbol prop in favor of TableAmount quiering supported assets for it in order to simplify TableAmount usage.

src/components/modern/AssetListItem/AssetListItem.js
```diff
@@ -64,10 +66,11 @@ function AssetListItem({
       <Icon url={iconUrl} />
       <Name>{name}</Name>
 
-      {!!symbol && balance != null && (
+      {!!symbol && !!address && balance != null && (
         <Balance
           chain={chain}
-          symbol={symbol}
+          assetSymbol={symbol}
```
mdjastrzebski, line 72: Similar to other componens, consider dropping assetSymbol to simplify component usage.

src/configs/assetsConfig.js
```diff
@@ -35,3 +45,16 @@ export default {
 };
 
 export const PPN_TOKEN = PLR;
+
+/* eslint-disable i18next/no-literal-string */
+export const getPlrAddressForChain = (chain: Chain): string => {
+  if (chain === CHAIN.BINANCE) return '0x790cfdc6ab2e0ee45a433aac5434f183be1f6a20';
```
mdjastrzebski, line 51: Consider moving exact address to assetConstants.

src/screens/Assets/rewards/RewardListItem.js
```diff
@@ -44,18 +44,19 @@ type Props = {|
...
 
-function RewardListItem({ title, subtitle, iconUrl, value, symbol, onPress, chain }: Props) {
+function RewardListItem({ title, subtitle, iconUrl, value, assetSymbol, assetAddress, onPress, chain }: Props) {
   const colors = useThemeColors();
 
   const chainRates = useChainRates(chain);
   const currency = useFiatCurrency();
 
-  const fiatValue = value.times(getRate(chainRates, symbol, currency));
+  const fiatValue = value.times(getAssetRateInFiat(chainRates, assetAddress, currency));
```
mdjastrzebski, line 59: Consider adding rates util `function getAssetBalanceInFiat(balance, rates, address, currency)` that would handle edge cases like missing asset rate.

src/utils/archanova.js
```diff
@@ -258,10 +270,10 @@ export const parseArchanovaTransactions = (
     if (transactionType === AccountTransactionTypes.Settlement) {
       // get and process all transactions with the same hash
       const extra = sameHashTransactions.map(tx => {
-        const txAsset = getAssetSymbolByAddress(assets, supportedAssets, tx.tokenAddress) || ETH;
+        const txAsset = findAsset(assets, supportedAssets, tx.tokenAddress);
         const txValue = tx.tokenValue.toString();
         return {
-          symbol: txAsset,
+          symbol: txAsset ? txAsset.symbol : ETH, // Archanova supports ETH only
```
mdjastrzebski, line 276:
```suggestion
          symbol: txAsset?.symbol ?? ETH, // Archanova supports ETH only
```

src/utils/rates.js
```diff
@@ -20,28 +20,49 @@
...
 import { wrapBigNumber } from 'utils/bigNumber';
+import { valueForAddress } from 'utils/common';
 
 // Types
-import type { RatesBySymbol, Currency } from 'models/Rates';
+import type { Currency, RatesByAssetAddress } from 'models/Rates';
 
-export function getUsdToFiatRate(rates: RatesBySymbol, currency: Currency): ?number {
+
+export const getUsdToFiatRate = (rates: RatesByAssetAddress, currency: Currency): ?number => {
```
mdjastrzebski, line 33: Looks somewhat weird that we would pass rates from different chain here ;-)

mdjastrzebski (approved):
Wow that was a heroic undertaking ⚔️ Well done 🥇 

I've found that we have some weird looking code after the changes, added related comments e.g. about `findAsset`

src/models/Collectible.js
```diff
@@ -37,7 +37,8 @@ export type Collectible = {
 };
 
 export type CollectibleTransaction = {
-  asset: string,
+  assetSymbol: string,
```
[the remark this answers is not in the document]
poocart: You will be amazed to hear that deep down the answer is yes :D however in this case we were using `asset` from any history item in both Archanova and Etherspot history components to map into title of event, but now since it has changed to `assetSymbol` it had to be changed here as well. Completely removing it is kinda yet one more refactor on whole history in the app. TBH `asset` never made sense here (and on `Transaction`) either and it confused a lot with `Asset` type.

src/models/Transaction.js
```diff
@@ -212,7 +213,8 @@ export type TransactionEthers = {
   value: string | Object,
```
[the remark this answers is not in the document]
poocart: So I looked at the past for this :) https://github.com/pillarwallet/pillarwallet/commit/bfc16072ef011e82bad31eb23e5e44025bcc46a5

Yes, it seems it had history of transactions sent through `ethers` and response being mapped into our type.

src/utils/assets.js
```diff
@@ -267,58 +190,16 @@ export const addressesInclude = (addresses: string[], addressToFind: ?string): b
...
-
-  // NOTE: ZERO_ADDRESS usually means it's ETH transaction
-  if (address === ZERO_ADDRESS) return ETH;
-
-  const { symbol } = getAssetDataByAddress(assets, supportedAssets, address);
-  if (symbol) assetSymbol = symbol;
-
-  return assetSymbol;
-};
+): ?Asset => userAssets.find(({ address }) => addressesEqual(address, assetAddress))
```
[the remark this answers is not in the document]
poocart: Makes sense as there are quite few calls with empty array as one or another argument. Will add a fix to accept only single assets array and for the calls that use 2 asset arrays I will spread them as single `findAsset` param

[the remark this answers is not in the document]
poocart: Or actually just call all of them with supported assets, spreading doesn't make sense. Wdyt?

src/actions/transactionEstimateActions.js
poocart: Yes, that is intended at this moment in time. PLR is available on all chains except xDai. If there's no or empty PLR address returned by `getPlrAddressForChain` then `findAssetByAddress` would return no asset therefore gas token will be native asset.

Actually now thinking that this should be replaced into:

```
const gasTokenAddress = preferredGasTokenSelector(getState()) === PLR
      ? getPlrAddressForChain(chain)
      : null
const gasToken = useGasToken && gasTokenAddress
      ? findAssetByAddress(chainSupportedAssets, gasTokenAddress)
      : null;
```

src/components/EventDetails/EventDetails.js
poocart:
```
export const getAssetRateInFiat = (
  rates: RatesByAssetAddress,
  assetAddress: string,
  fiatCurrency: Currency,
): number => valueForAddress(rates, assetAddress)?.[fiatCurrency] ?? 0;
```

So might be just 0, not nil, unless I miss something? 🤔

src/screens/Assets/rewards/RewardListItem.js
poocart: `getAssetRateInFiat` returns `0` in case there's no rate so I guess it's fine unless I miss something here? 🤔

src/utils/rates.js
poocart: Wdym?

src/components/FeeLabelToggle/FeeLabelToggle.js
poocart: Not sure I follow, is this about `getAssetRateInFiat` or `feeInFiat` result?

src/components/EventDetails/EventDetails.js
mdjastrzebski: I would differentiate between zero value and missing value for the assets

src/screens/Assets/rewards/RewardListItem.js
mdjastrzebski: y, now it will probably coalesce null to zero. IMO we should differentiate between zero value and missing value. wdyt?

src/components/Table/Table.js
poocart: Per our internal discussion added this as TODO: : get assetSymbol from matching asset once assets can be queried by assetAddress as key instead of performing expensive search on whole assets array.

src/components/modern/AssetListItem/AssetListItem.js
poocart: Per our internal discussion added this as to do: get assetSymbol from matching asset once assets can be queried by assetAddress as key instead of performing expensive search on whole assets array.

src/components/HistoryEventDetails/layouts/EnsNameEventDetails.js
poocart: Per our internal discussion added this as to do: get assetSymbol from matching asset once assets can be queried by assetAddress as key instead of performing expensive search on whole assets array.

src/utils/rates.js
mdjastrzebski: We are doing EUR/USD or GBP/USD and we base the calculation on different native token prices depending on which chain rates we use. Hopefully all chains will result it the same EUR/USD rate, but probably mainnet would be most reliable one.

Seems like this function is not used a lot, maybe we could inline it into `usdToFiatRateSelector` to make sure it's always based on mainnet, wdyt?

src/components/FeeLabelToggle/FeeLabelToggle.js
mdjastrzebski: Ignore the original comment. What I meant is to support cases where getAssetRateInFiat return null. Probably the best way would be to have `getAssetValueInFiat(value, address, rates, currency)` to do the multiplication and edge case handling. wdyt?

src/screens/Assets/rewards/RewardListItem.js
poocart: I do agree, but this involves either showing some guidance to user that cannot get rate he expects and/or hide the value, not sure 🤔

src/components/FeeLabelToggle/FeeLabelToggle.js
poocart: `getAssetValueInFiat` is not null and has fallback of `0`, but I guess per our last discussion let's keep as is.

src/utils/rates.js
poocart: Will see what we can do here.

PR #2771, opened 2021-07-09
 PIL-870- Maintenance: Remove Key Wallet notifications

divyaBacancy:


src/actions/notificationsActions.js
```diff
@@ -103,14 +101,6 @@ const onFirebaseMessageAction = (message: FirebaseMessage) => {
     if (type === FCM_DATA_TYPE.COLLECTIBLE) {
       dispatch(fetchAllCollectiblesDataAction());
     }
-
-    if (message.notification) {
-      const { wallet: { data: wallet } } = getState();
-      const notification = message.data && getToastNotification(message.data, wallet?.address);
```
poocart, line 109: What if we replace this piece to get current account address and pass it to `getToastNotification` instead of key based wallet? This way we can have same functionality for Smart wallets when it will start working and the `getToastNotification`can remain, wdyt?

divyaBacancy: @poocart Done with the changes.

src/actions/notificationsActions.js
```diff
@@ -105,8 +106,11 @@ const onFirebaseMessageAction = (message: FirebaseMessage) => {
     }
 
     if (message.notification) {
-      const { wallet: { data: wallet } } = getState();
-      const notification = message.data && getToastNotification(message.data, wallet?.address);
+      const {
+        accounts: { data: accounts },
+      } = getState();
+      const walletAddress = getActiveAccountAddress(accounts);
```
poocart, line 112:
```suggestion
      const walletAddress = activeAccountAddressSelector(getState());
```

poocart (approved):
Small note, but LGTM 👍

src/actions/notificationsActions.js
divyaBacancy: @poocart Done with the changes.

PR #2772, opened 2021-07-13
rates by asset contract address

poocart:
Includes:
- [x] Adds Rates by contract address
- [x] Fixes other missing pieces for Assets by contract address
- [x] Asset rates and asset balances by lowercase asset contract address (rates by asset contract address #2773).

### DIFF
--- a/src/actions/ratesActions.js
+++ b/src/actions/ratesActions.js
@@ -85,8 +85,8 @@ export const fetchAssetsRatesAction = () => {
     const supportedAssetsPerChain = supportedAssetsPerChainSelector(getState());
 
     // combine assets balances from all accounts
-    const assetsAddressPerChain = Object.keys(CHAIN).reduce((
-      combinedAssetsBySymbolPerChain,
+    const assetsByAddressPerChain = Object.keys(CHAIN).reduce((
+      combinedAssetsByAddressPerChain,
       chainKey,
     ) => {
       const chain = CHAIN[chainKey];
@@ -94,7 +94,7 @@ export const fetchAssetsRatesAction = () => {
       const chainSupportedAssets = supportedAssetsPerChain[chain] ?? [];
 
       const chainAccountsAssetsByAddress = Object.keys(assetsBalancesPerAccount).reduce((
-        combinedAssetsBySymbol,
+        combinedAssetsByAddress,
         accountId,
       ) => {
         const accountAssetsBalances = assetsBalancesPerAccount[accountId] ?? {};
@@ -106,15 +106,15 @@ export const fetchAssetsRatesAction = () => {
         );
 
         // $FlowFixMe
-        return { ...combinedAssetsBySymbol, ...assetsByAddress };
+        return { ...combinedAssetsByAddress, ...assetsByAddress };
       }, {});
 
       // $FlowFixMe
-      return { ...combinedAssetsBySymbolPerChain, [chain]: chainAccountsAssetsByAddress };
+      return { ...combinedAssetsByAddressPerChain, [chain]: chainAccountsAssetsByAddress };
     }, {});
 
-    await Promise.all(Object.keys(assetsAddressPerChain).map(async (chain) => {
-      const chainAssetsByAddress = assetsAddressPerChain[chain] ?? {};
+    await Promise.all(Object.keys(assetsByAddressPerChain).map(async (chain) => {
+      const chainAssetsByAddress = assetsByAddressPerChain[chain] ?? {};
 
       try {
         const rates = await getExchangeRates(chain, getAssetsAsList(chainAssetsByAddress));
--- a/src/actions/ratesActions.js
+++ b/src/actions/ratesActions.js
@@ -45,7 +45,7 @@ import { assetsBalancesPerAccountSelector } from 'selectors/balances';
 // models, types
 import type { Dispatch, GetState } from 'reducers/rootReducer';
 import type { Chain } from 'models/Chain';
-import type { RatesBySymbol } from 'models/Rates';
+import type { RatesByAssetAddress } from 'models/Rates';
 
 // actions
 import { saveDbAction } from './dbActions';
@@ -58,7 +58,7 @@ export const setIsFetchingRatesAction = (isFetching: boolean) => ({
 
 export const updateRatesAction = (
   chain: string,
-  rates: RatesBySymbol,
+  rates: RatesByAssetAddress,
 ) => {
   return async (dispatch: Dispatch, getState: GetState) => {
     if (isEmpty(rates)) return;
@@ -85,7 +85,7 @@ export const fetchAssetsRatesAction = () => {
     const supportedAssetsPerChain = supportedAssetsPerChainSelector(getState());
 
     // combine assets balances from all accounts
-    const assetsBySymbolPerChain = Object.keys(CHAIN).reduce((
+    const assetsAddressPerChain = Object.keys(CHAIN).reduce((
       combinedAssetsBySymbolPerChain,
       chainKey,
     ) => {
@@ -93,34 +93,34 @@ export const fetchAssetsRatesAction = () => {
 
       const chainSupportedAssets = supportedAssetsPerChain[chain] ?? [];
 
-      const chainAccountsAssetsBySymbol = Object.keys(assetsBalancesPerAccount).reduce((
+      const chainAccountsAssetsByAddress = Object.keys(assetsBalancesPerAccount).reduce((
         combinedAssetsBySymbol,
         accountId,
       ) => {
         const accountAssetsBalances = assetsBalancesPerAccount[accountId] ?? {};
         const accountWalletAssetsBalances = accountAssetsBalances[chain]?.wallet ?? {};
 
-        const assetsBySymbol = mapWalletAssetsBalancesIntoAssetsByAddress(
+        const assetsByAddress = mapWalletAssetsBalancesIntoAssetsByAddress(
           accountWalletAssetsBalances,
           chainSupportedAssets,
         );
 
         // $FlowFixMe
-        return { ...combinedAssetsBySymbol, ...assetsBySymbol };
+        return { ...combinedAssetsBySymbol, ...assetsByAddress };
       }, {});
 
       // $FlowFixMe
-      return { ...combinedAssetsBySymbolPerChain, [chain]: chainAccountsAssetsBySymbol };
+      return { ...combinedAssetsBySymbolPerChain, [chain]: chainAccountsAssetsByAddress };
     }, {});
 
-    await Promise.all(Object.keys(assetsBySymbolPerChain).map(async (chain) => {
-      const chainAssetsBySymbol = assetsBySymbolPerChain[chain] ?? {};
+    await Promise.all(Object.keys(assetsAddressPerChain).map(async (chain) => {
+      const chainAssetsByAddress = assetsAddressPerChain[chain] ?? {};
 
       try {
-        const rates = await getExchangeRates(chain, chainAssetsBySymbol);
+        const rates = await getExchangeRates(chain, getAssetsAsList(chainAssetsByAddress));
         await dispatch(updateRatesAction(chain, rates));
       } catch (error) {
-        reportErrorLog('fetchAssetsRatesAction failed', { error, chain, chainAssetsBySymbol });
+        reportErrorLog('fetchAssetsRatesAction failed', { error, chain, chainAssetsByAddress });
       }
     }));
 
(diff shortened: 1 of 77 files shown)


src/models/Rates.js
```diff
@@ -34,6 +34,6 @@ export type Rates = {
   ETH?: number,
 }
 
-export type RatesBySymbol = { [symbol: string]: Rates };
+export type RatesByAssetAddress = { [assetAddress: string]: Rates };
```
mdjastrzebski, line 37:
```suggestion
export type RatesByAddress = { [assetAddress: string]: Rates };
```

Simplier name is descriptive enough, since these are rates for assets.

src/models/Value.js
```diff
@@ -23,6 +23,7 @@ import { BigNumber } from 'bignumber.js';
 export type TokenValue = {|
```
mdjastrzebski, line 23: This type got a bit confusing in the multichain world.

Probably should have only following fields: `chain, address, value`, as a minimal asset balance representation. Without `chain` it feels incomplete. Wdyt?

src/actions/ratesActions.js
```diff
@@ -85,42 +85,42 @@ export const fetchAssetsRatesAction = () => {
...
 
-        const assetsBySymbol = mapWalletAssetsBalancesIntoAssetsByAddress(
+        const assetsByAddress = mapWalletAssetsBalancesIntoAssetsByAddress(
           accountWalletAssetsBalances,
           chainSupportedAssets,
         );
 
         // $FlowFixMe
-        return { ...combinedAssetsBySymbol, ...assetsBySymbol };
+        return { ...combinedAssetsBySymbol, ...assetsByAddress };
```
mdjastrzebski, line 109:
```suggestion
        return { ...combinedAssetsByAddress, ...assetsByAddress };
```

src/actions/ratesActions.js
```diff
@@ -85,42 +85,42 @@ export const fetchAssetsRatesAction = () => {
...
       }, {});
 
       // $FlowFixMe
-      return { ...combinedAssetsBySymbolPerChain, [chain]: chainAccountsAssetsBySymbol };
+      return { ...combinedAssetsBySymbolPerChain, [chain]: chainAccountsAssetsByAddress };
```
mdjastrzebski, line 113:
```suggestion
      return { ...combinedAssetsByAddressPerChain, [chain]: chainAccountsAssetsByAddress };
```

src/actions/smartWalletActions.js
```diff
@@ -1208,14 +1208,16 @@ export const fetchAvailableTxToSettleAction = () => {
     const activeAccountAddress = activeAccountAddressSelector(getState());
     const accountEthereumHistory = archanovaAccountEthereumHistorySelector(getState());
     const accountAssets = archanovaAccountEthereumAssetsSelector(getState());
+    const accountAssetsList = getAssetsAsList(accountAssets);
```
mdjastrzebski, line 1211: The longer I think about our wallet assets balances and supportes assets, I think they should be otherway around:

`walletAssetBalances` should be array as a simple list of user assets
`supportedAssets` should be record by address to ease up lookup.

That's just my feeling, I know that changing this would be huge undertaking ;-)

src/actions/smartWalletActions.js
```diff
@@ -1229,7 +1231,7 @@ export const fetchAvailableTxToSettleAction = () => {
         return {
           token: item.token,
           hash: item.hash,
-          value: new BigNumber(formatUnits(item.value, decimals)),
+          value: new BigNumber(formatUnits(item.value, asset?.decimals ?? 18)),
```
mdjastrzebski, line 1234: pls use `fromEthersBigNumber` from 'utils/bignumber'

src/utils/rates.js
```diff
@@ -18,30 +18,63 @@
...
 
-export function getFiatValueFromUsd(valueInUsd: ?BigNumber | string, usdToFiatRate: ?number): ?BigNumber {
+export const getFiatValueFromUsd = (valueInUsd: ?BigNumber | string, usdToFiatRate: ?number): ?BigNumber => {
   if (!valueInUsd || usdToFiatRate == null) return null;
 
   return wrapBigNumber(valueInUsd)?.times(usdToFiatRate);
-}
+};
+
+export const convertValueInUsdToFiat = (
```
mdjastrzebski, line 53: Seems like duplication of functionality of `getFiatValueFromUsd`, consider re-using existing function which encourages single calculation of usd to fiat rate.

src/utils/rates.js
```diff
@@ -18,30 +18,63 @@
...
+  fiatCurrency: Currency,
+): number => {
+  const usdToFiatRate = getUsdToFiatRate(rates, fiatCurrency);
+
+  if (!usdToFiatRate) return 0;
+
+  return value * usdToFiatRate;
+};
+
+export const getRate = (
```
mdjastrzebski, line 65:
```suggestion
export const getAssetRateInFiat = (
```

src/utils/rates.js
```diff
@@ -18,30 +18,63 @@
...
+  rates: RatesByAssetAddress,
+  assetAddress: string,
+  fiatCurrency: Currency,
+): number => rates?.[assetAddress]?.[fiatCurrency] ?? 0;
+
+export const getFormattedRate = (
```
mdjastrzebski, line 71:
```suggestion
export const formatAssetBalanceInFiat = (
```
Existing name suggested that we returned rate where as in fact we returned formatted balance in fiat.

src/utils/assets.js
```diff
@@ -99,54 +104,6 @@ export const getBalance = (balances: ?WalletAssetsBalances, assetAddress: string
   return +formatAmount(number.toString());
 };
 
-const baseRate = (rates: RatesBySymbol, asset: string, fiatCurrency: Currency): number => {
```
mdjastrzebski, line 102: great to see that 🪓 🔥

src/services/coinGecko.js
```diff
@@ -79,26 +78,23 @@ const chainToCoinGeckoNetwork = {
...
-  assets: AssetByAddress,
-): Promise<?RatesBySymbol> => {
-  const assetsList = getAssetsAsList(assets);
-  const nativeAssetSymbol = nativeAssetPerChain[chain].symbol;
-
+  assets: Asset[],
+): Promise<?RatesByAssetAddress> => {
   // native asset not always fit into token price endpoint, it is fetched with other API call
-  const assetsListWithoutNativeAsset = assetsList.filter(({ symbol }) => symbol !== nativeAssetSymbol);
+  const assetsWithoutNativeAsset = assets.filter(({ address }) => address !== nativeAssetPerChain[chain].address);
```
mdjastrzebski, line 95: `addressEqual`?

src/selectors/totalBalances.js
```diff
@@ -154,18 +153,21 @@ export const accountRewardsBalancePerChainSelector = (root: RootReducerState) =>
...
-
-    const rate = getRate(rates, symbol, currency);
-    return BigNumber(balance).times(rate);
-  });
+  const assetBalancesInFiat = map(
+    categoryAssetBalances.wallet ?? {},
+    ({ symbol, address, balance }: WalletAssetBalance) => {
+      if (!balance) return BigNumber(0);
+
+      const hasMatchingServiceAsset = hasServiceAssetBalanceForSymbol(categoryAssetBalances, supportedAssets, symbol);
```
mdjastrzebski, line 164: We should do matching by address here as well

src/screens/Tank/TankWithdrawal/TankWithdrawal.js
```diff
@@ -241,7 +254,7 @@ class TankWithdrawal extends React.Component<Props, State> {
     );
     const formFields = getAmountFormFields({
       icon: iconUrl,
-      currency: token,
+      currency: symbol,
```
mdjastrzebski, line 257: mixing currency and symbols? something does not feel right here.

src/screens/Assets/wallet/selectors.js
```diff
@@ -81,10 +81,10 @@ const buildWalletItem = (
   const { name: title, iconUrl } = asset;
 
   return {
-    key: `${chain}-${symbol}`,
+    key: `${chain}-${assetSymbol}`,
```
mdjastrzebski, line 84:
```suggestion
    key: `${chain}-${address}`,
```

src/models/Rates.js
poocart: Yes, but the reason I tried to be explicit here (and it other places) is that when it comes to address it can be either Wallet address, Asset address and etc. address. IMO it's better to have explicit which is easier to newcomers to read.

src/screens/Tank/TankWithdrawal/TankWithdrawal.js
poocart: No idea mate, nothing changed here except naming. Seems like `getAmountFormFields` is used only on PPN flows and all of the calls have `currency`. Will check deeper, but I'm leaning to not touch this at all (except the rename part).

src/utils/common.js
```diff
@@ -595,3 +596,13 @@ export const formatBigFiatAmount = (value: Value, fiatCurrency: string) => {
...
+
+export const addressAsKey = (address: string): string => {
+  if (!address) return '';
+  return address.toLowerCase();
+};
+
+export const valueForAddress = <RecordValue>(
+  record: Record<RecordValue>,
+  address: string,
+): ?RecordValue => record[addressAsKey(address)];
```
mdjastrzebski, line 608: Consider allowing for nil record to avoid need for `?? {}` in various call sites.

```suggestion
export const valueForAddress = <Value>(
  record: ?Record<Value>,
  address: string,
): ?Value => record?.[addressAsKey(address)];
```

mdjastrzebski (approved):

src/models/Value.js
poocart: I don't think it needs `chain` just because the value is within history that comes for a selected chain and at higher level we already know which chain it is. Same would happen when it comes to sending if we would reuse this type.

Mixed feeling about `symbol` tho. It surely helps when it comes to rendering as we no longer need to lookup for asset just for symbol, but it's not attached to `TokenValue` besides that.

I would keep as is without removing `symbol` as removing symbol would require us to add asset lookups in quite few components.

wdyt?

src/actions/smartWalletActions.js
poocart: I does makes sense. But as you stated it's one more huge refactor so for this task let's leave as it is and keep the improvement in mind. Sounds good?

src/utils/rates.js
poocart: Not quite the same as `convertValueInUsdToFiat` additionally gets the `usdToFiat` rate.

src/models/Value.js
mdjastrzebski: This was intended as general purpose token value structure (not only related to history functionality). So it contained minimal data that identified token type and amount.

IMO I would drop `symbol` as this can be derrived from `address`, but would consider adding `chain` (even if it's duplicating `chain` in history events), because address without chain is not clearly describing asset anymore in multichain app.

src/utils/rates.js
poocart: Also the value returned using `convertValueInUsdToFiat` is `number` which is expected on the screens where the method is used. Changing this method would require refactoring the screens which is out of this scope.

src/utils/rates.js
poocart: It might be any value so I would go for `formatAssetValueInFiat`, wdyt?

src/utils/common.js
poocart: `Value` is reserved for imported type in the above, added changes for nullish.

src/screens/Tank/TankWithdrawal/TankWithdrawal.js
poocart: I don't think `currency` is used at all, it's passed as `config`, but never destructured from it, only `icon` from these calls.

src/models/Value.js
poocart: Nothing really changed since it was the same before with `symbol` since `symbol` can exist in any chain and without `chain` it's the same as it is without `address`. Not sure about applying this fix into this PR 🤔

poocart: I get your thinking, but if we go this direction it's not just `TokenValue` type and related components.

mdjastrzebski: Yeah, this seems more to be technical debt related to mutlichain than something about this PR. maybe just add a TODO comment about debt here.

src/utils/rates.js
poocart: Removed.

PR #2773, opened 2021-07-13
added assets and rates by lowercase contract address

poocart:
Adds asset rates and asset balances by lowercase asset contract address.

### DIFF
--- a/src/components/ValueInput/ValueInput.js
+++ b/src/components/ValueInput/ValueInput.js
@@ -46,7 +46,7 @@ import {
   hasTooMuchDecimals,
   noop,
   formatFiat,
-  valueAsKey,
+  valueForAddress,
 } from 'utils/common';
 import { getThemeColors } from 'utils/themes';
 import { images } from 'utils/images';
@@ -169,7 +169,7 @@ const ValueInputComponent = ({
   const assetSymbol = assetData?.symbol || '';
   const chain = assetData?.chain || CHAIN.ETHEREUM;
   const walletBalances = accountAssetsBalances?.[chain]?.wallet ?? {};
-  const assetBalance = (customBalances || walletBalances)?.[valueAsKey(assetAddress)]?.balance || '0';
+  const assetBalance = valueForAddress(customBalances ?? walletBalances, assetAddress)?.balance || '0';
   const balanceAvailable = calculateMaxAmount(assetSymbol, assetBalance);
 
   const collectibles = collectiblesPerChain[chain] ?? [];
--- a/src/components/ValueInput/ValueInput.js
+++ b/src/components/ValueInput/ValueInput.js
@@ -46,6 +46,7 @@ import {
   hasTooMuchDecimals,
   noop,
   formatFiat,
+  valueAsKey,
 } from 'utils/common';
 import { getThemeColors } from 'utils/themes';
 import { images } from 'utils/images';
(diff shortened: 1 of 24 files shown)


src/utils/common.js
```diff
@@ -595,3 +595,8 @@ export const formatBigFiatAmount = (value: Value, fiatCurrency: string) => {
 export const getEnsName = (username: string) => `${username}${getEnsPrefix()}`;
 
 export const extractUsernameFromEnsName = (ensName: string) => ensName.replace(getEnsPrefix(), '');
+
+export const valueAsKey = (value: string | number): string => {
```
mdjastrzebski, line 599:
```suggestion
export const addressAsKey = (value: string): string => {
```

Other potential names to consider `toAddressKey`, `keyFromAddress`, etc

Naming should reflect that this is intended for address keys, I think that we do not want to assume that each record value needs to be lowercase. wdyt?

src/utils/common.js
```diff
@@ -595,3 +595,8 @@ export const formatBigFiatAmount = (value: Value, fiatCurrency: string) => {
...
+  if (!value) return '';
+  return value.toString().toLowerCase();
+};
```
mdjastrzebski, line 602: I would also add 
```
export const valueForAddress = <Value>(record: Record<Value>, address: ?string): ?Value => {
   return record[valueAsKey(address)];
}
```

Which would slightly improve call site by reducing parenthesis nesting & correctly indicate that the key might return nil value.

```
if (!balances[valueAsKey(gasTokenAddress)]) return false;
```
 =>
```
if (!valueForAddress(balances, gasTokenAddress)) return false;
```

src/utils/rates.js
```diff
@@ -36,7 +36,7 @@ export const getUsdToFiatRate = (rates: RatesByAssetAddress, currency: Currency)
   if (currency === USD) return 1;
 
   // will select native asset rates to calculate between
-  const nativeAssetRates = rates[ethers.constants.AddressZero];
+  const nativeAssetRates = rates[valueAsKey(ethers.constants.AddressZero)];
```
mdjastrzebski, line 39: Can we have zero address re-exported from e.g. assetConstants. Importing ethers every time is a bit tedious.

src/components/ValueInput/ValueInput.js
```diff
@@ -168,7 +169,7 @@ const ValueInputComponent = ({
   const assetSymbol = assetData?.symbol || '';
   const chain = assetData?.chain || CHAIN.ETHEREUM;
   const walletBalances = accountAssetsBalances?.[chain]?.wallet ?? {};
-  const assetBalance = (customBalances || walletBalances)?.[assetAddress]?.balance || '0';
+  const assetBalance = (customBalances || walletBalances)?.[valueAsKey(assetAddress)]?.balance || '0';
```
mdjastrzebski, line 172:
```suggestion
  const assetBalance = (customBalances ?? walletBalances)?.[valueAsKey(assetAddress)]?.balance || '0';
```

mdjastrzebski (approved):
Looks great!

Suggested some minor improvements that could help with readability.

src/utils/common.js
poocart: Honestly I would prefer this for more generic name, but we use it for address only as for now so I guess it's ok, replaced

PR #2779, opened 2021-07-15
value input asset address fix, single rate fetch by address fix

poocart:
Includes:
- Quick fix to get either `address` or `contractAddress` on `ValueInput` component.
- Small fix that found for fetching single asset rates.


mdjastrzebski (approved):
LG

We should do global rename `contractAddress` to `address` ;-)

PR #2784, opened 2021-07-20
PIL-1117: Archanova -> Etherspot Migration Tool

mdjastrzebski:
Scope:
- [x] Archanova asset migration flow: intro -> select assets -> review -> pin confirm -> tokens sent
- [x] Edit amount when pressing on token balance in select assets
- [x] Status screen when there is a pending migration transaction
- [x] Support send max ETH by reducing the ETH transfer to ETH balance - fee if needed
- [x] Use Archanova migrator contract  
- [x] Detailed breadcrumbs and error reporting for transaction estimation/submission part


mdjastrzebski:
This PR contains some code that could be shared with Deimantas' ENS migration but since we developed these functionalities kinda separately, and both are tested separately I would recommend to keep the code as is in this regard.

src/locales/en/common.json
```diff
@@ -693,6 +701,55 @@
     }
   },
 
+  "walletMigrationArchanova": {
+    "banner": {
+      "title": "Migrate assets from Pillar v1 to Etherspot (Pillar v2)"
```
poocart, line 706: just a note to check re `v2` I think it's marketed as `V2`

mdjastrzebski: I've got that Dmitry's designs (e.g. https://app.zeplin.io/project/5fa95ca539e6804a10dec0a4/screen/608befab4cf4eb02d439bb11). Will check with him

src/models/Collectible.js
```diff
@@ -21,6 +21,12 @@
 // Types
 import type { Chain, ChainRecord } from 'models/Chain';
 
+/** Colectible is uniquly identified by both address & id */
+export type CollectibleId = {|
```
poocart, line 25: Not sure what, but I have mixed feelings about this. Nothing major tho, but for me `id` sounds like single value identifier as either number or string. Checked throughout the code and I think there are 3 options to improve this:
1. Create a method which we can reuse in other screens/components/flows too called `getCollectibleId` and construct single string value `ID` using `tokenId` and `contractAddress` of `Collectible`.
2. Rename `CollectibleId` to `CollectibleContractMeta` or similar, but without `id` or `key`?
3. Just pass `Collectible` everywhere instead of `CollectibleId` and get needed values.

src/reducers/walletMigrationArchanovaReducer.js
```diff
@@ -0,0 +1,108 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+import produce from 'immer';
```
poocart, line 21: Not sure about this, extra library just for this piece is extra technical debt to maintain, can't we just spreads like in other reducers? 👀

src/screens/Menu/Menu.js
```diff
@@ -254,17 +255,19 @@ const Menu = ({
...
-            <MigrateEnsBanner style={{ marginTop: 15 }} />
-            {showMigrateWallet && (
-              <FooterBanner>
-                <MigrateWalletBanner onPress={navigateToKeyBasedAssetMigration} />
-              </FooterBanner>
-            )}
+            <BannersWrapper>
+              <MigrateEnsBanner style={styles.footerBanner} />
+
+              <WalletMigrationArchanovaBanner style={styles.footerBanner} />
```
poocart, line 261: Do you think we can add feature flag for Archanova migration too?

src/models/Collectible.js
mdjastrzebski: It's a tricky subject. So according to ERC721 spec, an NFT is uniquely identified by (contract) address and (token) id. Hence I created this type, which I guess is quite useful in general, and I believe is more natural to use this form than string concat of address & id.

Naming is hard here, as we want a type to uniquely identify NFT, but part of that also called `id`. So maybe we can rename type to `CollectibleUniqueIdentifier`/`CollectibleUniqueId`?

Re option 2: the name is more confusing as contract meta seems very general term, and without the `id` part we are no longer uniquely identifying an NFT. Which defeats the purpose.

Re option 3: passing collectible around is an option, but imo that kinda repeating yourself, as you should have 1 source of truth for collectible defintion and then you should be able to reference the collectible without need to duplicate the whole thing. (compare to having asset definition in `supportedAsset` and referencing it with using token `address`)

src/reducers/walletMigrationArchanovaReducer.js
mdjastrzebski: In my experience from past projects, this allows more readable code than spreading deeply nested structures. This dep is small and has a clear purpose which it already achieves, so there is little risk of building tech debt because of it.

src/screens/Menu/Menu.js
mdjastrzebski: @IAmKio ^

src/locales/en/common.json
mdjastrzebski: Dmitry confirmed it's `v2` and not `V2`.

src/reducers/walletMigrationArchanovaReducer.js
mdjastrzebski: As an argument "for" using Immer, it is included in the [Redux Toolkit](https://redux-toolkit.js.org/usage/immer-reducers), which is officially toolkit from Redux team.

src/screens/WalletMigrationArchanova/Review/MigrationSummary.js
```diff
@@ -0,0 +1,144 @@
...
+        <Table.RowValue fontVariant="tabular-nums">{formatTokenValue(balance, token.symbol)}</Table.RowValue>
+
+        {!!balanceInFiat && (
+          <Table.RowValue variant="secondary" fontVariant="tabular-nums">
+            {formatFiatValue(balanceInFiat, currency)}
+          </Table.RowValue>
+        )}
+      </Table.RowContainer>
+    );
+  };
```
poocart, line 85:
```suggestion
  const renderTokenItem = ({ token, balance, balanceInFiat }: TokenItem, index: number) => (
    <Table.RowContainer key={token.address} separator={index !== 0}>
      <Table.RowTitle>{token.name}</Table.RowTitle>
      <Table.RowValue fontVariant="tabular-nums">{formatTokenValue(balance, token.symbol)}</Table.RowValue>

      {!!balanceInFiat && (
        <Table.RowValue variant="secondary" fontVariant="tabular-nums">
          {formatFiatValue(balanceInFiat, currency)}
        </Table.RowValue>
      )}
    </Table.RowContainer>
  );
```

src/screens/WalletMigrationArchanova/Review/MigrationSummary.js
```diff
@@ -0,0 +1,144 @@
...
+  const renderCollectibleItem = (collectible: Collectible, index: number) => {
+    return (
+      <Table.Row
+        key={collectible.contractAddress}
+        title={collectible.name}
+        value={tRoot('label.collectible')}
+        separator={index !== 0}
+      />
+    );
+  };
```
poocart, line 96:
```suggestion
  const renderCollectibleItem = (collectible: Collectible, index: number) => (
    <Table.Row
      key={collectible.contractAddress}
      title={collectible.name}
      value={tRoot('label.collectible')}
      separator={index !== 0}
    />
  );
```

src/screens/WalletMigrationArchanova/Review/WalletMigrationArchanovaReview.js
```diff
@@ -0,0 +1,137 @@
...
+
+  const handlePinValid = async (pin: string, wallet: any) => {
+    setShowCheckAuth(false);
+    setIsEstimating(true);
+
+    try {
+      const feeInEth = await estimateMigrationTransactions(wallet, accounts, tokensToMigrate, collectiblesToMigrate);
+      setFee(feeInEth);
+    } catch (error) {
+      reportErrorLog('WalletMigrationArchanovaReview: failed to estimate transaction', {
```
poocart, line 76: Any Toast for user or some information?

mdjastrzebski: `estimateArchanovaRawTransactions` will show toast if estimation fails. For success scenario, setting `fee` will show the fee table on that screen.

src/locales/en/common.json
poocart: All articles, vids, banners, emails use `V2` so I guess it's `V2` 🤔

src/models/Collectible.js
poocart: Yes, what makes NFTs unique is contract address and token ID.

I do agree this one is tricky, however I don't feel like putting both address and token ID into object and instead just use single value from `getCollectibleId` (option 1), wdyt?

Re option 2 I do agree that meta makes think of that there's more which leads to just option 3 which is the same as passing current object, but instead we don't add extra type to out codebase which you're aware has waaaaaay to many types that describe same thing which is another concern I have about this new `CollectibleId ` type.

src/screens/WalletMigrationArchanova/Review/WalletMigrationArchanovaReview.js
poocart: 👍

src/reducers/walletMigrationArchanovaReducer.js
poocart: `there is little risk of building tech debt because of it` – last famous words 😂 

but putting jokes aside lets go with it then even and if it goes well we can adapt it to other reducers

src/models/Collectible.js
mdjastrzebski: How about following solution:


```
const buildCollectibleKey = (contractAddress: string, tokenId: string) => `${valueAsKey(contractAddress)}-${tokenId}`
```

and additionally

```
const  getCollectibleKey = (collectible: Collectible) => buildCollectibleKey(collectible.contractAddress, collectible.id)
```

src/screens/WalletMigrationArchanova/SelectAssets/WalletMigrationArchanovaSelectAssets.js
```diff
@@ -0,0 +1,136 @@
...
+
+  const handleTokenBalancePress = (address: string, balance: BigNumber) => {
+    navigation.navigate(WALLET_MIGRATION_ARCHANOVA_SET_AMOUNT, { address, balance: balance.toFixed() });
+  };
+
+  const navigateToReview = () => {
+    navigation.navigate(WALLET_MIGRATION_ARCHANOVA_REVIEW);
+  };
+
+  const walletAddress = archanovaAccount?.id ?? '';
```
poocart, line 110: Would be better to use `getAccountAddress` util

src/selectors/accounts.js
```diff
@@ -0,0 +1,41 @@
...
+// Utils
+import { findFirstEtherspotAccount, findFirstArchanovaAccount } from 'utils/accounts';
+
+// Types
+import type { RootReducerState, Selector } from 'reducers/rootReducer';
+import type { Account } from 'models/Account';
+
+export const etherspotAccountSelector: Selector<?Account> = createSelector(accountsSelector, (accounts: Account[]) =>
+  findFirstEtherspotAccount(accounts),
+);
```
poocart, line 35:
```suggestion
export const etherspotAccountSelector: Selector<?Account> = createSelector(
  accountsSelector,
  (accounts: Account[]) => findFirstEtherspotAccount(accounts),
);
```

src/selectors/accounts.js
```diff
@@ -0,0 +1,41 @@
...
+
+export const archanovaAccountSelector: Selector<?Account> = createSelector(accountsSelector, (accounts: Account[]) =>
+  findFirstArchanovaAccount(accounts),
+);
```
poocart, line 39:
```suggestion
export const archanovaAccountSelector: Selector<?Account> = createSelector(
  accountsSelector,
  (accounts: Account[]) => findFirstArchanovaAccount(accounts),
);
```

src/selectors/selectors.js
```diff
@@ -43,9 +44,9 @@ export const useRootSelector = <T>(selector: (state: RootReducerState) => T): T
 // Most commonly used selectors
 export const useFiatCurrency = () => useRootSelector(fiatCurrencySelector);
 export const useRatesPerChain = () => useRootSelector(ratesPerChainSelector);
-export const useChainRates = (chain: Chain) => useRatesPerChain()[chain] ?? {};
+export const useChainRates = (chain: Chain) => useRatesPerChain()[chain] ?? EMPTY_OBJECT;
```
poocart, line 47: Uhhh, not sure on `EMPTY_OBJECT` and `EMPTY_ARRAY`, looks too much hassle for nothing 👀

src/selectors/walletMigrationArchanova.js
```diff
@@ -0,0 +1,84 @@
...
+  },
+);
+
+// Checks for non-negligible balances
+export const showWalletMigrationArchanovaSelector: Selector<boolean> = createSelector(
+  archanovaWalletAssetsBalancesSelector,
+  archanovaCollectiblesSelector,
+  ethereumRatesSelector,
+  (walletBalances: WalletAssetsBalances, collectibles: Collectible[], rates: RatesByAssetAddress): boolean => {
+    return hasNonNegligileWalletBalances(walletBalances, rates) || !!collectibles.length;
```
poocart, line 62: Do you think we can replace `hasNonNegligileWalletBalances` into something easier to type and read like `hasEnoughWalletBalancesToMigrate`?

src/services/archanova.js
```diff
@@ -246,6 +250,10 @@ export class ArchanovaService {
       .catch(this.handleError);
   }
 
+  getConnectedAccountDevice(deviceAddress: string): Promise<?ArchanovaAccountDevice> {
+    return this.getSdk().getConnectedAccountDevice(deviceAddress);
```
poocart, line 254: Please add exception handler 🙏

src/utils/archanova.js
```diff
@@ -339,11 +344,11 @@ export const parseArchanovaTransactions = (
...
 
     const migratorContractAddress = getEnv().ARCHANOVA_MIGRATOR_CONTRACT_ADDRESS;
-    if (addressesEqual(migratorContractAddress, transaction.to)
-      && inputData?.includes(ARCHANOVA_ENS_TRANSFER_METHOD_HASH)) {
+    if (addressesEqual(migratorContractAddress, transaction.to)) {
+      const isEnsMigration = inputData?.includes(ARCHANOVA_ENS_TRANSFER_METHOD_HASH);
       transaction = {
         ...transaction,
-        tag: ARCHANOVA_WALLET_ENS_MIGRATION,
+        tag: isEnsMigration ? ARCHANOVA_WALLET_ENS_MIGRATION : ARCHANOVA_WALLET_ASSET_MIGRATION,
```
poocart, line 351: 🚀

src/utils/archanova.js
```diff
@@ -502,6 +507,29 @@ export const buildEnsMigrationRawTransactions = async (accounts: Account[], wall
     });
 };
 
+export async function estimateArchanovaRawTransactions(rawTransactions: string[]): Promise<BigNumber> {
+  if (!rawTransactions?.length) {
+    reportErrorLog('estimateArchanovaRawTransactions: no transactions to estimate');
+    throw new Error(t('error.noTransactionsToEstimate'));
```
poocart, line 513: This one seems to not be handled gently

src/utils/object.js
```diff
@@ -42,10 +42,21 @@ export function buildRecord<Element, Value>(
...
-export function recordValues<Value>(record: Record<Value>): Value[] {
+export function recordValues<Value>(record: ?Record<Value>): Value[] {
+  if (!record) return [];
   return Object.keys(record).map((key) => record[key]);
 }
 
+/**
+ * Create a shallow copy of given record with removed key.
+ */
+export function recordWithRemovedKey<Value>(record: Record<Value>, key: string): Record<Value> {
```
poocart, line 53: Since it involves an action maybe `removeKeyFromRecord`?

src/utils/transactions.js
```diff
@@ -183,14 +181,39 @@ const mapTransactionToTransactionPayload = (transaction: EthereumTransaction): T
...
+
+  return {
+    fee: sumBy(feeInfos, info => info.fee),
+    gasToken: feeInfos[0].gasToken,
+  };
 };
 
-export const getGasAddress = (chain: Chain, gasToken: ?GasToken) => {
-  return gasToken?.address ?? nativeAssetPerChain[chain].address;
+// TODO: remove
```
poocart, line 193: 👀  remove `TransactionFeeSummary`?

src/utils/walletMigrationArchanova.js
```diff
@@ -0,0 +1,299 @@
...
+    collectiblesToMigrate,
+  });
+  const transactionsToEstimate = await buildAssetMigrationRawTransactions(
+    wallet,
+    accounts,
+    tokensToMigrate,
+    collectiblesToMigrate,
+  );
+  if (!transactionsToEstimate?.length) {
+    reportErrorLog('submitMigrationTransactions: transactionsToEstimate is empty');
```
poocart, line 124:
```suggestion
    reportErrorLog('estimateMigrationTransactions: transactionsToEstimate is empty');
```

src/utils/walletMigrationArchanova.js
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  logBreadcrumb('walletMigrationArchanova', 'estimate archanova raw transaction');
+  const fee = await estimateArchanovaRawTransactions(transactionsToEstimate);
+
+  logBreadcrumb('walletMigrationArchanova', 'estimated archanova raw transaction', { fee: fee.toFixed() });
+  return fee;
+}
+
+export function hasNonNegligileWalletBalances(balances: ?WalletAssetsBalances, rates: RatesByAssetAddress) {
```
poocart, line 135: Maybe `hasEnoughWalletBalancesToMigrate`?

src/utils/walletMigrationArchanova.js
```diff
@@ -0,0 +1,299 @@
...
+  if (!balances) return false;
+
+  return Object.keys(balances).some((symbol) => isNonNegligibleBalance(balances[symbol], rates));
+}
+
+const NON_NEGLIGIBLE_BALANCE_THRESHOLD_IN_TOKEN = '0.25';
+const NON_NEGLIGIBLE_BALANCE_THRESHOLD_IN_USD = 1;
+
+export function isNonNegligibleBalance(balance: WalletAssetBalance, rates: RatesByAssetAddress) {
```
poocart, line 144: Maybe `isEnoughBalanceToMigrate`?

src/utils/walletMigrationArchanova.js
```diff
@@ -0,0 +1,299 @@
...
+      }
+    }
+
+    return data;
+  });
+};
+
+async function applyAddMigratorDeviceTransactionIfNeeded(migrator: Migrator): Migrator {
+  const migratorDevice = await archanovaService.getConnectedAccountDevice(migrator.migratorAddress);
+  const isMigratorDeviceDeployed = migratorDevice?.state === 'Deployed';
```
poocart, line 248: Can use `isArchanovaDeviceDeployed`

poocart (approved):
Well done man! 💪 🚀 🚀 🚀 

Please let's resolve all comments before merge! 🙏

src/selectors/selectors.js
mdjastrzebski: Original code would return new object instance (`{}`) each time the `useRatesPerChain()[chain]` returned null/undefined. Hence if any `useEffect`/`useMemo`/etc relied on that value then, it would be triggered on each render because the identity of the object changed. This would not be a problem if `useRatesPerChain` used `useMemo`, but I believe that returning the same instance of `EMPTY_OBJECT` is simpler solution.

src/selectors/walletMigrationArchanova.js
mdjastrzebski: "Enough" would suggest we are talking about having enough ETH to pay for the tx fee. Here we are checking that user funds are potentially worth migrating (i.e. are non-negligible). Feel free to suggest another name that more succinctly describes that concept.

src/utils/object.js
mdjastrzebski: I used this during dev, but replaced it with `immer`. Removing that method.

src/utils/transactions.js
mdjastrzebski: Removed

src/utils/walletMigrationArchanova.js
mdjastrzebski: Pls check comments in this reply: https://github.com/pillarwallet/pillarwallet/pull/2784#discussion_r679360435

src/utils/walletMigrationArchanova.js
mdjastrzebski: `isArchanovaDeviceDeployed` not only checks `state === Deployed` but also returns true when `nextState === Deployed`. Not sure what is the meaning of `nextState` in`ArchanovaAccountDevice`. wdyt?

src/utils/walletMigrationArchanova.js
mdjastrzebski: Please check [this comment](https://github.com/pillarwallet/pillarwallet/pull/2784#discussion_r679360435).

src/utils/walletMigrationArchanova.js
poocart: `nextState` means that account device is being deployed.

src/selectors/walletMigrationArchanova.js
poocart: Not necessarily, we don't say anything about ETH and tx so it would have no relations to that, however, if my suggestion still confuses you and I cannot come with any other options we can keep, but I think there's a typo on `negligile` which should be `negligible`.

src/selectors/selectors.js
poocart: Maybe we can then rename `EMPTY_OBJECT` to `LOCKED_ EMPTY_OBJECT` or `FROZEN_ EMPTY_OBJECT`, wdyt?

src/screens/Menu/Menu.js
mdjastrzebski: DOne

PR #2794, opened 2021-07-30
PIL-1130:- Change Max Ethereum Fee text

divyaBacancy:
FeeLabel


IAmKio (approved):

poocart (changes requested):
Please also check string `allowancePlusEthFee` and there are more cases for `ethFee` (i.e. `DetailsTable` which is used in Exchange flow supported by Etherspot).

divyaBacancy:
@poocart Though etherspot is used, but it only uses ethereum chain here. 
Still do we need to handle?

divyaBacancy:
@poocart Done with the changes.

poocart (approved):

PR #2795, opened 2021-07-30
check if address starts with 0x before validating

poocart:
Turns out `isValidAddress` from `ethereumjs-util` doesn't work with any values that doesn't start with `0x`.


src/services/etherspot.js
```diff
@@ -297,7 +297,7 @@ export class EtherspotService {
 
   getEnsNode(nameOrHashOrAddress: string): Promise<?ENSNode> {
     // if it's address – getENSNode accepts only checksum addresses
-    const nameOrHashOrChecksumAddress = isValidAddress(nameOrHashOrAddress)
+    const nameOrHashOrChecksumAddress = nameOrHashOrAddress.startsWith('0x') && isValidAddress(nameOrHashOrAddress)
```
IAmKio, line 300: Wowser

IAmKio (approved):

PR #2797, opened 2021-08-03
PIL-1129: added few improvements to fix wc estimate loop

poocart:
https://linear.app/pillarproject/issue/PIL-1129/check-continuous-estimation-triggered-on-walletconnect-transaction

This can only be replicated on production build and I was unable to do this locally. By inspecting code this is what I've found and improved:
1. I noticed that upon WC transaction reject it is still estimated, that happens because request that is passed from component is still there while at hook level for updated `callRequests` it's already removed. Adde additional check to follow this.
2. React `useEffect` return is a function that is executed on update, removed estimate action dispatch from that.
3. Added memoized return for `useTransactionFee`.
4. Added fallbacks to gas token address and decimals (we already had default symbol falling back to ETH).

### DIFF
--- a/src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
+++ b/src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
@@ -170,11 +170,31 @@ const useTransactionFee = (request: WalletConnectCallRequest) => {
   };
   const hasNotEnoughGas = !isEnoughBalanceForTransactionFee(walletBalances, balanceCheckTransaction, chain);
 
-  const { estimateCallRequestTransaction } = useWalletConnect();
+  const { estimateCallRequestTransaction, callRequests } = useWalletConnect();
 
-  React.useEffect(() => estimateCallRequestTransaction(request), [request, estimateCallRequestTransaction]);
+  React.useEffect(() => {
+    // perform additional check to avoid estimate on request dismiss
+    const requestExists = callRequests.some(({ callId }) => +callId === +request.callId);
+    if (!requestExists) return;
 
-  return { fee, gasSymbol, gasAddress, hasNotEnoughGas, isEstimating, estimationErrorMessage };
+    estimateCallRequestTransaction(request);
+  }, [request, estimateCallRequestTransaction, callRequests]);
+
+  return React.useMemo(() => ({
+    fee,
+    gasSymbol,
+    gasAddress,
+    hasNotEnoughGas,
+    isEstimating,
+    estimationErrorMessage,
+  }), [
+    fee,
+    gasSymbol,
+    gasAddress,
+    hasNotEnoughGas,
+    isEstimating,
+    estimationErrorMessage,
+  ]);
 };
 
 const useViewData = (request: WalletConnectCallRequest) => {
--- a/src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
+++ b/src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
@@ -174,7 +174,7 @@ const useTransactionFee = (request: WalletConnectCallRequest) => {
 
   React.useEffect(() => {
     // perform additional check to avoid estimate on request dismiss
-    const requestExists = callRequests.some(({ callId }) => +callId === +request.callId);
+    const requestExists = callRequests.some(({ callId }) => callId === request.callId);
     if (!requestExists) return;
 
     estimateCallRequestTransaction(request);
--- a/src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
+++ b/src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
@@ -180,21 +180,14 @@ const useTransactionFee = (request: WalletConnectCallRequest) => {
     estimateCallRequestTransaction(request);
   }, [request, estimateCallRequestTransaction, callRequests]);
 
-  return React.useMemo(() => ({
+  return {
     fee,
     gasSymbol,
     gasAddress,
     hasNotEnoughGas,
     isEstimating,
     estimationErrorMessage,
-  }), [
-    fee,
-    gasSymbol,
-    gasAddress,
-    hasNotEnoughGas,
-    isEstimating,
-    estimationErrorMessage,
-  ]);
+  };
 };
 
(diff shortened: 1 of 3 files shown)


IAmKio (approved):
Defensive FTW 👍 LGTM

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
```diff
@@ -170,11 +170,31 @@ const useTransactionFee = (request: WalletConnectCallRequest) => {
   };
   const hasNotEnoughGas = !isEnoughBalanceForTransactionFee(walletBalances, balanceCheckTransaction, chain);
 
-  const { estimateCallRequestTransaction } = useWalletConnect();
+  const { estimateCallRequestTransaction, callRequests } = useWalletConnect();
 
-  React.useEffect(() => estimateCallRequestTransaction(request), [request, estimateCallRequestTransaction]);
+  React.useEffect(() => {
+    // perform additional check to avoid estimate on request dismiss
+    const requestExists = callRequests.some(({ callId }) => +callId === +request.callId);
```
mdjastrzebski, line 177: Why are we doing `+string` casting to `number` here? Could we go with regular string comparison?

```suggestion
    const requestExists = callRequests.some(({ callId }) => +callId === +request.callId);
```

poocart: Not sure, definitely doesn't make sense. And it's actually a number heh 👀  changed

src/utils/transactions.js
```diff
@@ -182,11 +186,11 @@ const mapTransactionToTransactionPayload = (transaction: EthereumTransaction): T
 };
 
 export const getGasDecimals = (chain: Chain, gasToken: ?GasToken) => {
-  return gasToken?.decimals ?? nativeAssetPerChain[chain].decimals;
+  return gasToken?.decimals ?? nativeAssetPerChain[chain].decimals ?? 18;
```
mdjastrzebski, line 189: Seems like this change could actually mask the real problem, which is passing `null/undefined` chain object, which should be handled at call site(s). wdyt?

poocart: Hm, would make sense, but then I would suggest expand these gas calls with a check if `nativeAssetPerChain[chain]` exists for `chain` and log it if it doesn't, wdyt?

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
```diff
@@ -170,11 +170,31 @@ const useTransactionFee = (request: WalletConnectCallRequest) => {
...
+  React.useEffect(() => {
+    // perform additional check to avoid estimate on request dismiss
+    const requestExists = callRequests.some(({ callId }) => callId === request.callId);
+    if (!requestExists) return;
 
-  return { fee, gasSymbol, gasAddress, hasNotEnoughGas, isEstimating, estimationErrorMessage };
+    estimateCallRequestTransaction(request);
+  }, [request, estimateCallRequestTransaction, callRequests]);
+
+  return React.useMemo(() => ({
```
mdjastrzebski, line 183: There is no point in using `React.useMemo` here, because the only current use-case applies destructuring to the call and does not use the whole returned object.

src/utils/transactions.js
mdjastrzebski: yeah, we could add an error log here

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
```diff
@@ -170,11 +170,31 @@ const useTransactionFee = (request: WalletConnectCallRequest) => {
...
 
-  React.useEffect(() => estimateCallRequestTransaction(request), [request, estimateCallRequestTransaction]);
+  React.useEffect(() => {
+    // perform additional check to avoid estimate on request dismiss
+    const requestExists = callRequests.some(({ callId }) => callId === request.callId);
+    if (!requestExists) return;
 
-  return { fee, gasSymbol, gasAddress, hasNotEnoughGas, isEstimating, estimationErrorMessage };
+    estimateCallRequestTransaction(request);
+  }, [request, estimateCallRequestTransaction, callRequests]);
```
mdjastrzebski, line 181: Looks like we could improve the things a bit by applying `useCallback` to `estimateCallRequestTransaction`. While the whole `useWalletConnect` result is memoized, the particular callback should stay the same since it's only calling action method. Same for `approveCallRequest` and `rejectCallRequest`

src/utils/transactions.js
poocart: Added.

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
poocart: Added for all calls.

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
poocart: On one hand since this `useTransactionFee` is performing lots of operations maybe it's good to memoize it, but on the other hand I realized since it has `useEffect` in it then it won't perform estimate as return is memoized based on other values. For now removed, but something to think about it here as well.

poocart: Ok, removing for now, but I think this needs rework.

src/utils/transactions.js
```diff
@@ -181,14 +186,38 @@ const mapTransactionToTransactionPayload = (transaction: EthereumTransaction): T
...
 };
 
-export const getGasSymbol = (chain: Chain, gasToken: ?GasToken) => {
-  return gasToken?.symbol ?? nativeAssetPerChain[chain].symbol ?? ETH;
+export const getGasSymbol = (chain: Chain, gasToken: ?GasToken): string => {
+  if (gasToken?.symbol) return gasToken.symbol;
+
+  const chainNativeAsset = nativeAssetPerChain[chain];
+  if (!chainNativeAsset) {
+    reportErrorLog('getGasSymbol failed: no native asset for chain', { chain });
```
mdjastrzebski, line 218: I think we should also transfor `chainFromChainId` record to a function which would also call `reportErrorLog`. There we would also have info about incorrect `chainId`.

PR #2801, opened 2021-08-05
PIL-1136: Legacy NFT support

mdjastrzebski:
Scope:
- [x] Handle legacy NFTs in send collectible flow
- [x] Improve type checking in related code


src/components/SendAsset/SendAsset.js
```diff
@@ -175,19 +175,14 @@ const SendAsset = ({
...
       to: selectedContact.ethAddress,
-      // $FlowFixMe: flow update to 0.122
       receiverEnsName: selectedContact.ensName,
       amount: truncateAmount(currentValue.toString(), assetData.decimals),
-      // $FlowFixMe: bignumber.js types
       txFeeInWei: feeInfo.fee,
       // $FlowFixMe: flow update to 0.122
-      symbol: assetData.token,
-      // $FlowFixMe: flow update to 0.122
+      symbol: assetData.symbol || assetData.token,
```
poocart, line 184: Side noter: I think we really really need to get rid of any remaining `token` references :( or even replace `AssetData` usage with just `Asset`

src/models/Transaction.js
```diff
@@ -174,11 +177,11 @@ export type Transaction = {
 export type TransactionPayload = {
   gasLimit?: number,
   to: string,
-  receiverEnsName?: string,
+  receiverEnsName?: ?string,
```
poocart, line 180: How about either present as valid or not present at all?

src/models/Collectible.js
```diff
@@ -18,27 +18,35 @@
...
-  symbol?: void,
-  decimals?: void,
-  chain: Chain,
 
   // Legacy NFTs are minted pre ERC-721 (e.g. cryptokitties) and require using `transfer` method instead of one
   // inferred from ABI.
   isLegacy: boolean,
+
+  // Cooperation with AssetData/TokenData model
+  token?: void,
```
poocart, line 45: Do we really need that? Is there any way we can get rid of `token`? Having it has many down sides that might end up in wrong mappings where `symbol` can be missing because of `token` and we only look for `symbol`.

src/utils/common.js
```diff
@@ -121,7 +121,7 @@ export const logBreadcrumb = (
     else printLog(`${level} - ${category}: ${message}`);
   }
   Sentry.addBreadcrumb({
-    category, message, level, data: { extra },
+    category, message, level, data: extra,
```
poocart, line 124: My only concern would be here if `extra` is passed not as object and flow fails to read that what would happen then?

src/components/SendAsset/SendAsset.js
mdjastrzebski: I definitely agree, Asset vs AssetData vs AssetOption vs Collectible is a bit confusing atm. Let's plan that for the future. Also would suggest clarify the naming, e.g. `Asset = (Fungible)Token | NonFungibleToken` or similar.

src/models/Transaction.js
mdjastrzebski: I've updated the type to reflect true potential values with the current code.

src/models/Collectible.js
mdjastrzebski: This is only to clarify the typing in the components we already use. Cleaning that up is a matter of large tech debt PR.

src/utils/common.js
mdjastrzebski: The reason I've put it here, was that breadcrumbs have limited depth in Sentry, and adding another level of nesting actually limited displayed data in sentry. Will add some safety checks here.

src/components/SendAsset/SendAsset.js
poocart: 👍

src/utils/common.js
poocart: 👍

poocart (approved):

PR #2802, opened 2021-08-05
PIL-1112: Fix/wallet event improvements

poocart:
https://linear.app/pillarproject/issue/PIL-1112/wallets-deployment-ui-guidance

Includes
- [x] Refactored `userEvents` to multi chain supported `walletEvents` that share `WalletEvent` type from updated history,
- [x] Removed "Key wallet" stories from `ActivityFeedItem` component.
- [x] Added Wallet Deployment events from Etherspot transactions history
- [x] Added a fix to recover from mutated date strings into actual `Date`.
- [x] Added Wallet Deployment tooltip in transaction confirm screen.
- [x] Added extra Wallet Event UI improvements per ticket.
- [x] Added check for missing Wallet Create event.
- [x] Added ENS Registration to Etherspot history for Etherspot accounts that does not have it migrated.

Known:
- No fees are returned from Etherspot transactions history yet.
- Wallet Backup is not shown on Etherspot history.


src/actions/walletEventsActions.js
```diff
@@ -0,0 +1,151 @@
...
+        });
+      }
+    }
+
+    const updatedWalletEvents = getState().walletEvents.data;
+    dispatch(saveDbAction('walletEvents', { walletEvents: updatedWalletEvents }, true));
+  };
+};
+
+export const getWalletsCreationEventsAction = () => {
```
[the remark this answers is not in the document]
mdjastrzebski: What is the difference between this and `addMissingWalletEventsIfNeededAction`? Looks like we should have only one of these, probalby `addMissingWalletEventsIfNeededAction`.

src/utils/common.js
```diff
@@ -621,3 +621,6 @@ export const valueForAddress = <V>(
 export const setValueForAddress = <V>(record: Record<V>, address: string, value: V) => {
   record[addressAsKey(address)] = value;
 };
+
+
+export const parseDate = (date: Date | string | number): Date => new Date(date);
```
mdjastrzebski, line 626: I'm wondering if we need date parsing function if all it does is `new Date(date)`?

src/utils/common.js
```diff
@@ -471,7 +470,8 @@ export const groupSectionsByDate = (
   const sections: { [string]: SectionData } = {};
 
   orderBy(data, [dateField], [sortDirection]).forEach((item) => {
-    const date = new Date(item[dateField] * timestampMultiplier);
+    const safeTimestamp = +parseDate(item[dateField]);
```
mdjastrzebski, line 473: Using unary `+` seems a bit hard to read. Would you conside using `parseDate(...).getTime()`? Alternatively could we add `getTimestamp(Date|string|number)`?

src/utils/common.js
mdjastrzebski: Maybe we should return `null` for cases where input is null or when the parsed date is invalid (e.g. incorrect string value). Pls compare: https://stackoverflow.com/a/1353711/484499
Also for undefined input we will return the current date, which might not be the expected value.
Also consider using parsing function from `date-fns` which we are already including.

src/utils/common.js
mdjastrzebski: With current implementation, safeTimestamp is a not adequate name, as the parsed date might be invalid and hence timestamp might be `NaN`.

src/utils/history.js
```diff
@@ -305,9 +309,28 @@ export const getHistoryEventsFromTransactions = (
     historyEvent = { ...historyEvent, ensName: extra.ensName };
   }
 
+  historyEvent = { ...historyEvent, type: eventType };
+
+  if (addressesEqual(fromAddress, activeAccountAddress)) {
```
mdjastrzebski, line 314: Aren't we going to add this deployment event before every tx originating from active account?

src/screens/SendToken/SendTokenConfirm.js
```diff
@@ -70,6 +74,12 @@ const SendTokenConfirm = () => {
     gasToken,
   } = transactionPayload;
 
+  const isDeployedOnChain = useRootSelector(isDeployedOnChainSelector)?.[chain];
+  const feeTooltip = isEtherspotAccount(activeAccount)
+    && chain
+    && !isDeployedOnChain
+    && t('tooltip.includesDeploymentFee');
```
mdjastrzebski, line 81:
```suggestion
  const feeTooltip = isEtherspotAccount(activeAccount)
    && chain
    && !isDeployedOnChain
    ? t('tooltip.includesDeploymentFee') : undefined;
```

mdjastrzebski: BTW do we need this `chain` check here?

src/screens/History/MultiChainHistoryEtherspot.js
```diff
@@ -140,3 +157,63 @@ function useHistoryEvents(chain: Chain): Event[] {
...
+  const walletEvents = useRootSelector((root) => root.walletEvents.data);
+  const activeAccount = useActiveAccount();
+
+  if (!activeAccount) return [];
+
+  const activeAccountId = getAccountId(activeAccount);
+  const accountChainWalletEvents = walletEvents?.[activeAccountId]?.[chain] ?? [];
+
+  // $FlowFixMe: does not cover WalletActivated and flow fails
+  return accountChainWalletEvents.map((event) => ({
```
mdjastrzebski, line 171: Consider using `useMemo` here.

mdjastrzebski: Or maybe this should be a `reselect` memoized selector?

src/screens/History/MultiChainHistoryEtherspot.js
```diff
@@ -140,3 +157,63 @@ function useHistoryEvents(chain: Chain): Event[] {
...
+    ...event,
+    date: parseDate(event.date),
+  }));
+}
+
+function useEnsRegisteredEvent(chain: Chain): ?EnsNameRegisteredEvent {
```
mdjastrzebski, line 177: Consider using extracting this as memoized selector to avoid expensive computation on each render.

src/screens/History/HistoryListArchanova.js
```diff
@@ -76,7 +78,17 @@ export default HistoryListArchanova;
...
+  const activeAccount = useActiveAccount();
+
+  const archanovaEthereumWalletEvents = activeAccount
+    ? walletEvents?.[getAccountId(activeAccount)]?.ethereum ?? []
+    : [];
+
+  const mappedArchanovaEthereumWalletEvents = archanovaEthereumWalletEvents.map((walletEvent) => ({
+    ...walletEvent,
+    createdAt: new Date(+parseDate(walletEvent.date) / 1000),
+  }));
```
mdjastrzebski, line 91: Consider using `useMemo` or extracting as memoized selector to avoid computation on every render.

src/reducers/walletEventsReducer.js
```diff
@@ -0,0 +1,74 @@
...
+  const accountChainWalletEvents = accountWalletEvents?.[chain] ?? [];
+
+  return {
+    ...walletEvents,
+    [accountId]: {
+      ...accountWalletEvents,
+      [chain]: accountChainWalletEvents.filter(({ type }) => type !== newWalletEvent.type).concat(newWalletEvent),
+    },
+  };
+};
```
mdjastrzebski, line 59: Consider using `immer`'s produce function to simplify the new state creation.

src/components/EventDetails/EventDetails.js
```diff
@@ -1138,10 +1127,10 @@ export class EventDetail extends React.Component<Props> {
       default:
         eventData = null;
     }
-    if (eventData) {
+    if (eventData && event.createdAt) {
       eventData = {
         ...eventData,
-        date: event.createdAt,
+        date: +parseDate(event.createdAt),
```
mdjastrzebski, line 1133: Consider creating `parseTimestamp` function, to avoid unary `+`.
```suggestion
        date: parseTimestamp(event.createdAt),
```

mdjastrzebski (approved):
Overall looks good!

Added some comments to improve code readability and structure.

src/actions/walletEventsActions.js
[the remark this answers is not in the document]
poocart: Removed completely.

src/screens/SendToken/SendTokenConfirm.js
poocart: Removed `chain`.

src/utils/common.js
poocart: Replaced this with `parseTimestamp`.

src/utils/common.js
poocart: Removed `parseDate` and replaced either with `new Date` or newly added `parseTimestamp` where needed.

src/utils/history.js
poocart: Updated this with check if successful or pending event is added. It will keep adding failed outgoing until adds successful or pending.

src/reducers/walletEventsReducer.js
poocart: Not yet familiar on using it, will consider adding it with next reducer created.

PR #2804, opened 2021-08-09
PIL-1131: added multi chain support for exchange

poocart:
https://linear.app/pillarproject/issue/PIL-1131/multichain-exchange-support

Adds multi chain swaps support.

Notes:
- Exchange chain changes based on selected pair asset.
- ~xDai not supported~.


src/utils/transactions.js
```diff
@@ -165,25 +161,34 @@ export const mapToEthereumTransactions = async (
...
+const mapTransactionToTransactionPayload = (
+  chain: Chain,
+  transaction: EthereumTransaction,
+): TransactionPayload => {
+  const { symbol } = nativeAssetPerChain[chain];
   const { to, value, data } = transaction;
   const amount = fromEthersBigNumber(value, 18).toFixed();
 
-  return { to, amount, symbol: ETH, data, decimals: 18 };
+  return { to, amount, symbol, data, decimals: 18 };
```
mdjastrzebski, line 191: Pls take `decimals` from `nativeAssetPerChain`, and apply it also to `fromEthersBigNumber` send arg.

mdjastrzebski: Pls also add `chain` prop here, instead of line 179

src/services/etherspot.js
```diff
@@ -683,7 +683,7 @@ export class EtherspotService {
         fromAmount: fromAmountEthers,
       });
 
-      return offers.map((offer) => buildExchangeOffer(fromAsset, toAsset, fromAmount, offer));
+      return offers.map((offer) => buildExchangeOffer(fromAsset, toAsset, fromAmount, offer, chain));
```
mdjastrzebski, line 686: Consider putting `chain` as first argument, in line with other functions.

src/screens/Exchange/Exchange/Exchange.js
```diff
@@ -225,16 +263,23 @@ function Exchange() {
 
 export default Exchange;
 
+const supportedExchangeChains = [
```
mdjastrzebski, line 266:
```suggestion
const exchangeSupportedChains = [
```

src/screens/Exchange/Exchange/Exchange.js
```diff
@@ -225,16 +263,23 @@ function Exchange() {
...
+  CHAIN.ETHEREUM,
+  CHAIN.BINANCE,
+  CHAIN.POLYGON,
+];
+
 function useOffersQuery(
   fromAsset: ?AssetOption,
   toAsset: ?AssetOption,
   fromAmount: string,
+  chain: Chain,
```
mdjastrzebski, line 276: Pls put chain as first argument in order to match other functions

src/screens/Exchange/Exchange/Exchange.js
```diff
@@ -73,46 +79,72 @@ function Exchange() {
...
+      const chainOptions = getExchangeToAssetOptions(
+        supportedAssetsPerChain,
+        walletBalancesPerChain,
+        fiatCurrency,
+        ratesPerChain,
+        supportedChain,
+      );
+      return [...multiChainOptions, ...chainOptions];
+    }, []),
+    [supportedAssetsPerChain, walletBalancesPerChain, fiatCurrency, ratesPerChain],
```
mdjastrzebski, line 129: AFAIR the original logic, picking the from assets restricted the available to asset options, so you may want to check that with Dmitry.

src/screens/Exchange/Exchange/Exchange.js
```diff
@@ -183,7 +218,10 @@ function Exchange() {
             disabled
             value={formattedToAmount}
             assetData={toAsset || fallbackAsset}
-            onAssetDataChange={(asset) => setToSymbol(asset.symbol)}
+            onAssetDataChange={(asset) => {
+              if (asset.chain !== chain) setChain(asset.chain);
```
mdjastrzebski, line 222: ❗ 
This looks like a potential source of trouble, because if `to` asset changes `chain`, then `from` asset address from different chain no longer makes sense.

I see two options to solve it:
1) Keep `chain` state but restrict to options to that particular chain (which I believe was in the original design)
2) Keep `fromChain` and `toChain` separate states

mdjastrzebski (commented):
Pls review the comment marked with exclamation.

Otherwise everything's fine with some readability comments.

src/screens/Exchange/Exchange/Exchange.js
mdjastrzebski: 3. Reset `fromAsset` if `toAsset` changes state and vice versa.

src/screens/Exchange/Exchange/Exchange.js
poocart: Per what I discussed with Dmitry before we allow to select either `from` or `to` asset on any chain and chain on the screen changed accordingly, but still allowing to choose assets from any chain.

src/screens/Exchange/Exchange/Exchange.js
poocart: In case this happens (either `to` or `from` doesn't match) then `from` or `to` asset changes to native asset now –`fallbackAsset`.

Based on updated `ValueInput` component UI found in Zeplin and per what I discussed with Dmitry earlier in this case it should fallback not to `fallbackAsset` but to `Select asset`, but current `ValueInput` does not support that. 

The only way is either leave as is (falling back to `fallbackAsset`) or improve `ValueInput` so it would show `Select asset` label in case there's no asset provided (we can then remove `fallbackAsset` completely).

src/utils/transactions.js
```diff
@@ -166,25 +166,34 @@ export const mapToEthereumTransactions = async (
...
+  chain: Chain,
+  transaction: EthereumTransaction,
+): TransactionPayload => {
+  const { symbol, decimals } = nativeAssetPerChain[chain];
   const { to, value, data } = transaction;
-  const amount = fromEthersBigNumber(value, 18).toFixed();
+  const amount = fromEthersBigNumber(value, decimals).toFixed();
 
-  return { to, amount, symbol: ETH, data, decimals: 18 };
+  return { to, amount, symbol, data, decimals };
```
mdjastrzebski, line 196: Do we need to set `contractAddress` (to `nativeAssetPerChain` `address`) here as well?

src/screens/Exchange/Exchange/utils.js
```diff
@@ -74,7 +90,8 @@ export const shouldTriggerSearch = (
     !!toAsset &&
     !!+fromAmount &&
     fromAmount[fromAmount.length - 1] !== '.' &&
-    fromAsset.symbol !== toAsset.symbol &&
+    fromAsset.address !== toAsset.address &&
```
mdjastrzebski, line 93: Pls use `addressesEqual` for address comparison.

mdjastrzebski (approved):
LGTM

PR #2806, opened 2021-08-12
PIL-1135 New V2 App Events

divyaBacancy:


src/actions/onboardingActions.js
```diff
@@ -547,7 +551,7 @@ export const checkUsernameAvailabilityAction = (username: string) => {
       'checkUsernameAvailabilityAction: done, dispatching SET_ONBOARDING_USER',
       username,
     );
-
+    isLogV2AppEvents() && dispatch(logEventAction('v2_ens_claimed'));
```
IAmKio, line 554: I'm not sure that this is the right place - I don't actually know if we receive an event for this specifically, @poocart do you know? If we don't get an event we may need to rename this to something more relevant...

src/actions/smartWalletActions.js
```diff
@@ -823,6 +825,7 @@ export const onSmartWalletSdkEventAction = (event: Object) => {
         && [ARCHANOVA_PPN_PAYMENT_COMPLETED, ARCHANOVA_PPN_PAYMENT_PROCESSED].includes(txStatus)) {
         const paymentInfo = `${formatMoney(txAmountFormatted.toString(), 4)} ${txToken}`;
         if (txStatus === ARCHANOVA_PPN_PAYMENT_COMPLETED) {
+          isLogV2AppEvents() && dispatch(logEventAction('v2_transaction_received'));
```
IAmKio, line 828: Wrong place, this is a transaction for the Archanova service which is now obsolete. You will need to add this where we receive transaction events from the Etherspot SDK.

src/actions/walletConnectActions.js
```diff
@@ -105,6 +106,7 @@ export const connectToWalletConnectConnectorAction = (uri: string) => {
     }
 
     dispatch(logEventAction('walletconnect_connector_requested'));
+    isLogV2AppEvents() && dispatch(logEventAction('v2_wallet_connect_dapp_connect'));
```
IAmKio, line 109: Is this line not supposed to go _inside_ the `connector.on` event listener below?

src/constants/chainConstants.js
```diff
@@ -25,6 +25,20 @@ export const CHAIN = {
...
+  BINANCE: 'binance',
+  XDAI: 'xdai',
+};
+
+export const CHAIN_SHORT = {
+  ETHEREUM: 'Mainnet',
+  POLYGON: 'BSC',
+  BINANCE: 'xDai',
+  XDAI: 'Polygon',
+};
```
IAmKio, line 40:
```suggestion
export const CHAIN_SHORT = {
  ETHEREUM: 'Mainnet',
  POLYGON: 'Polygon',
  BINANCE: 'BSC',
  XDAI: 'xDai',
};
```

src/screens/SendToken/SendTokenPinConfirmScreen.js
```diff
@@ -93,6 +95,7 @@ class SendTokenPinConfirmScreen extends React.Component<Props, State> {
       isChecking: true,
     }, () => {
       logEvent('transaction_sent', { source: this.source });
+      isLogV2AppEvents() && logEvent('v2_transaction_sent', { source: this.source });
```
IAmKio, line 98: Source looks like it might not contain any useful data... could we add the from / to symbols here please?

src/actions/etherspotActions.js
```diff
@@ -191,12 +199,44 @@ export const refreshEtherspotAccountsAction = () => {
...
+    case CHAIN_NAMES.BINANCE:
+      return CHAIN_SHORT.BINANCE.toLowerCase();
+    case CHAIN_NAMES.XDAI:
+      return CHAIN_SHORT.XDAI.toLowerCase();
+    case CHAIN_NAMES.POLYGON:
+      return CHAIN_SHORT.POLYGON.toLowerCase();
+    default:
+      return '';
+  }
+};
```
IAmKio, line 222: After seeing the definitions of `CHAIN_NAMES` and `CHAIN_SHORT`, it feels like we may have gone the long way around here. Where we have `const chains = Object.keys(etherspotAccountExtra)`, how does the data look? Can we not `toLowerCase()` this string in some way?

IAmKio (changes requested):
Great start, just a few things need changing and correcting 👍

divyaBacancy:
@IAmKio Thanks for reviewing. Done with the changes.

src/screens/SendToken/SendTokenPinConfirmScreen.js
```diff
@@ -88,11 +92,14 @@ class SendTokenPinConfirmScreen extends React.Component<Props, State> {
...
       return;
     }
+    const senderAccountAddress = getActiveAccountAddress(accounts);
+    const { to: recipient } = transactionPayload;
 
     this.setState({
       isChecking: true,
     }, () => {
       logEvent('transaction_sent', { source: this.source });
+      isLogV2AppEvents() && logEvent('v2_transaction_sent', { to: recipient, from: senderAccountAddress });
```
IAmKio, line 102: Sorry, i don't think i made this clear, but here the `from` and `to` need to be the `from` and `to` asset symbol (like PLR)

IAmKio (changes requested):
Just one minor thing but we're nearly there!

src/screens/SendToken/SendTokenPinConfirmScreen.js
divyaBacancy: @IAmKio Done

IAmKio (approved):
LGTM

PR #2807, opened 2021-08-12
PIL-1143: fixed no asset handling fr value input

poocart:
https://linear.app/pillarproject/issue/PIL-1143/improve-valueselector-component-to-allow-no-asset-state

Per title.

Side note: noticed that `fallbackSource` prop used for `FastImage` does not exist and likely is technical debt that needs to be fixed separately.


mdjastrzebski (approved):
LG

PR #2808, opened 2021-08-12
allow to select from asset if only one option available

poocart:
Covers edge case where `fromOptions` length is 1 and it does not contain ETH.

### DIFF
--- a/src/components/ValueInput/ValueInput.js
+++ b/src/components/ValueInput/ValueInput.js
@@ -296,7 +296,7 @@ const ValueInputComponent = ({
         onAssetPress={openAssetSelector}
         labelText={labelText}
         onLabelPress={() => !disabled ? handleUsePercent(100) : undefined}
-        disableAssetSelection={disableAssetChange || assetsOptions.length <= 1}
+        disableAssetSelection={disableAssetChange !== undefined ? disableAssetChange : assetsOptions.length <= 1}
       />
     );
   };
--- a/src/screens/Exchange/Exchange/Exchange.js
+++ b/src/screens/Exchange/Exchange/Exchange.js
@@ -207,6 +207,8 @@ function Exchange() {
               fromInputRef.current = ref;
             }}
             onBlur={() => fromInputRef.current?.blur()}
+            disableAssetChange={false}
+            hideMaxSend={!fromAsset}
           />
 
           <TouchableSwapIcon onPress={handleSwapAssets} disabled={!allowSwap}>


src/screens/Exchange/Exchange/Exchange.js
```diff
@@ -207,6 +207,8 @@ function Exchange() {
               fromInputRef.current = ref;
             }}
             onBlur={() => fromInputRef.current?.blur()}
+            disableAssetChange={false}
```
mdjastrzebski, line 210: It feels a bit wierd that disable has to be explicitly put to false. Normally you expect boolean value to be either true or false by default rather than default state being inactive.

src/components/ValueInput/ValueInput.js
```diff
@@ -296,7 +296,7 @@ const ValueInputComponent = ({
         onAssetPress={openAssetSelector}
         labelText={labelText}
         onLabelPress={() => !disabled ? handleUsePercent(100) : undefined}
-        disableAssetSelection={disableAssetChange || assetsOptions.length <= 1}
+        disableAssetSelection={disableAssetChange !== undefined ? disableAssetChange : assetsOptions.length <= 1}
```
mdjastrzebski, line 299: IMO i would rather modify the logic to autoselect the only asset if there is only one available. Wdyt?

src/screens/Exchange/Exchange/Exchange.js
poocart: Well the other approach is to add one more prop `forceAssetChange` which basically overrides this and also looks weird, idk :)

src/components/ValueInput/ValueInput.js
poocart: Maybe, but then this has to be done on `Exchange` screen and based on any first native asset found it should change initial value accordingly.

mdjastrzebski: Hmmm, how about following logic:
1) If there are zero available then display "No assets to select" or similar
2) If there is exactly one asset then automatically select it, trigger onAssetDataChange (as an useEffect)
3) If there are 2+ assets then follow the please select logic.

I guess that this would be optimal from UX perspective, as doing "Please select" when there is only one asset available is an unnecessary step.

wdyt?

poocart: Thanks for suggestion, but being pragmatic I feel that it's too much of logic added for something that will be refactored soon.

IAmKio (approved):

PR #2810, opened 2021-08-13
PIL-1138: Fix/narrow infura usage

poocart:
https://linear.app/pillarproject/issue/PIL-1138/narrow-infura-use


src/actions/historyActions.js
```diff
@@ -217,9 +214,7 @@ export const fetchTransactionsHistoryAction = () => {
         relayerExtensionDevice?.address,
       );
 
-      const aaveHistory = await mapTransactionsHistoryWithAave(accountAddress, archanovaTransactionsHistory);
-      const poolTogetherHistory = await mapTransactionsPoolTogether(accountAddress, aaveHistory);
-      const sablierHistory = await mapTransactionsHistoryWithSablier(accountAddress, poolTogetherHistory);
+      const sablierHistory = await mapTransactionsHistoryWithSablier(accountAddress, archanovaTransactionsHistory);
       const rariHistory = await mapTransactionsHistoryWithRari(accountAddress, sablierHistory, ethereumSupportedAssets);
```
poocart, line 218: I think we can gradually remove these mappings too.

mdjastrzebski: I recall that I removed all rari & sablier screens some time ago, maybe it's time to remove history entries as well?

poocart: Yes, we still had UI mappings for Archanova history items, but I think the time has come for them too :)

poocart: Will add to backlog.

src/utils/etherspot.js
```diff
@@ -70,6 +70,13 @@ const ETHERSPOT_TRANSACTION_HISTORY_STATUS = {
   REVERTED: 'Reverted',
 };
 
+export const parseEtherspotTransactionStatus = (transactionStatus: string) => {
+  if (transactionStatus === ETHERSPOT_TRANSACTION_HISTORY_STATUS.COMPLETED) return TRANSACTION_STATUS.CONFIRMED;
+  if (transactionStatus === ETHERSPOT_TRANSACTION_HISTORY_STATUS.REVERTED) return TRANSACTION_STATUS.FAILED;
+
+  return TRANSACTION_STATUS.PENDING;
```
mdjastrzebski, line 77: Do we somehow support timeout status? If etherspot does not return one, could we remove `TRANSACTION_STATUS.TIMEDOUT` on our end?

src/utils/liquidityPools.js
```diff
@@ -126,10 +144,25 @@ export const getAddLiquidityTransactions = async (
...
+    }
+
+    let approvedAmountBN;
+    try {
+      approvedAmountBN = await erc20Contract.callAllowance(sender, UNISWAP_ROUTER_ADDRESS);
+    } catch (error) {
+      reportErrorLog('getAddLiquidityTransactions failed: callAllowance failed', { tokenAsset, error });
+    }
 
     if (!approvedAmountBN || tokenAmountBN.gt(approvedAmountBN)) {
```
mdjastrzebski, line 167: Seems like return type is now ethers big number. Does it still work correctly?

src/utils/liquidityPools.js
```diff
@@ -269,10 +302,23 @@ export const getRemoveLiquidityTransactions = async (
...
+  }
+
+  let approvedAmountBN;
+  try {
+    approvedAmountBN = await erc20Contract.callAllowance(sender, UNISWAP_ROUTER_ADDRESS);
+  } catch (error) {
+    reportErrorLog('getRemoveLiquidityTransactions failed: callAllowance failed', { poolToken, error });
+  }
 
   if (!approvedAmountBN || poolTokenAmountBN.gt(approvedAmountBN)) {
```
mdjastrzebski, line 323: Seems like `callAllowance` return type is now ethers.js BigNumber. Does it still work correctly with regular BigNumber?

mdjastrzebski: Please also verify other places

mdjastrzebski (commented):
Generally looks good. 

I have concern about BigNumber type returned from Contract calls (`callAllowance` etc) being ethers.js BigNumber instead of BigNumber.js's one. Not sure we are aren't now using the result in a mixed ethers.js + BigNumber.js operations without proper mapping. 

Since BigNumber.js seems to coalesce to string this might work in some cases but not all.

Also we lack typing for ethers.js BigNumber, making it `any` type.

src/utils/liquidityPools.js
poocart: It was ethers BigNumber before as well since it was using ethers Contract interface directly and now the return is wrapper through Etherspot SDK.

src/utils/liquidityPools.js
poocart: It was ethers BigNumber before as well since it was using ethers Contract interface directly and now the return is wrapper through Etherspot SDK.

src/utils/etherspot.js
poocart: On Etherspot itself it doesn't exist, but there are existing mappings for Archanova so would keep as long as we support it.

src/utils/liquidityPools.js
mdjastrzebski: My suggestion would be to transform it to `bignumber.js` e.g. using `fromEthersBigNumber`. Otherwise were working on `any` type and mixing two not really compatible types.

src/utils/etherspot.js
mdjastrzebski: Makes sense

src/services/etherspot.js
```diff
@@ -698,6 +698,35 @@ export class EtherspotService {
       return [];
     }
   }
+
+  getContract<T>(
```
mdjastrzebski, line 702: Should we call this `registerContract` instead?

mdjastrzebski (approved):
LG

src/utils/liquidityPools.js
poocart: Not sure if we're on same track mate. Two things:
1. `fromEthersBigNumber` is used to convert wei values in that are in ethers BigNumber to other BigNumber lib.
2. ethers BigNumber was used in these calls before and replacing each usage to other BigNumber lib is out of scope for this PR as each use has different approaches.

src/services/etherspot.js
poocart: I think no because the name of Etherspot's `registerContract` is used to register and return while we the returned contract and we don't care about the register part. The flow is to get contract instance, not to register it, however, `registerContract` is the only way to get the contract instance.

PR #2814, opened 2021-08-16
PIL-1144: added subsidized deployment fix

poocart:
https://linear.app/pillarproject/issue/PIL-1144/fix-subsidised-etherspot-smart-wallet-deployments-on-first-transaction

Adds manual deployment for every transaction which is the only way to get subsidised deployments on Etherspot back-end.

Side note: `batchDeployAccount` on back-end additionally checks if account is deployed regardless of our state check and either skips or adds deployment transaction.

### DIFF
--- a/src/actions/transactionEstimateActions.js
+++ b/src/actions/transactionEstimateActions.js
@@ -69,7 +69,7 @@ export const setEstimatingTransactionAction = (isEstimating: boolean) => ({
 
 export const setTransactionsEstimateFeeAction = (feeInfo: ?TransactionFeeInfo) => {
   return (dispatch: Dispatch) => {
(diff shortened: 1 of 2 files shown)


src/actions/transactionEstimateActions.js
```diff
@@ -69,7 +69,7 @@ export const setEstimatingTransactionAction = (isEstimating: boolean) => ({
 
 export const setTransactionsEstimateFeeAction = (feeInfo: ?TransactionFeeInfo) => {
   return (dispatch: Dispatch) => {
-    if (!feeInfo || (feeInfo.fee && !feeInfo.fee.gt(0))) {
+    if (!feeInfo || (feeInfo.fee && !feeInfo.fee.gte(0))) {
```
IAmKio, line 72: Good catch

IAmKio (approved):
LGTM

PR #2817, opened 2021-08-16
target assets only matching source asset chain

poocart:
Includes:
1. After recent discussion we decided to allow target exchange asset to be only selected from assets matching source asset chain.
2. Sushiswap Exchange provider added.


IAmKio (approved):
LGTM

IAmKio:
Linear: https://linear.app/pillarproject/issue/PIL-1149/exchange-send-flow-chain-matching

PR #2825, opened 2021-08-19
TokenIcon & FiatIcon components

mdjastrzebski:
Scope:
- [x] FiatIcon & its application in the existing code
- [x] TokenIcon refactor & its application in the existing code

Both of these controls will be used for upcoming Send Flow update, but this PR is also useful as-is for existing code.


src/components/Icons/FiatIcon.js
```diff
@@ -19,34 +19,34 @@
...
-  style?: ViewStyleProp
+  currency: Currency,
+  size?: number,
+  style?: ImageStyleProp,
 |};
 
-function TokenIcon({ url, diameter = 48, style }: Props) {
-  const { genericToken } = useThemedImages();
+function FiatIcon({ currency, size = 48, style }: Props) {
+  const source = iconSoureForCurrency[currency];
```
poocart, line 40: Seems like `FiatIcon` would no longer have `genericToken` fallback, is this intentional?

poocart (approved):
LGTM except `FiatIcon` fallback 👀

src/components/Icons/FiatIcon.js
mdjastrzebski: I'll add fallback for fiat as well

PR #2831, opened 2021-08-20
PIL-1160 "Disconnect" text missing in WalletConnect

divyaBacancy:
lightheme
blackTheme


mdjastrzebski (changes requested):
I've checked all the current designs, and it seems that we're no longer using "primary destructive" button layout. Instead these rare cases that did, now seem to use text-destructive variant. 
image

Since there are only 2 use-cases for this style, could we change them to text-destructive & get rid of primary destructive?

mdjastrzebski (approved):
Looks good! 🚀

PR #2837, opened 2021-08-24
do not fetch onboarding tutorial on login if its disabled

poocart:
Includes:
- Some chore around Tutorial flow.
- Added feature flag check to Tutorial data fetch to not try fetching if it's disabled.

Note: In the code we're still waiting for this call to complete, but while `loginAction` is yet to be refactored to not wait for it to complete before showing Dashboard.


src/actions/onboardingActions.js
```diff
@@ -340,11 +336,11 @@ export const finishOnboardingAction = (retry?: boolean) => {
     );
     dispatch(initialDeeplinkExecutedAction());
 
-    logBreadcrumb('onboarding', 'finishOnboardingAction: checking for FEATURE_ONBOARDING flag for enable onboarding');
-    const enableOnboarding = firebaseRemoteConfig.getString(REMOTE_CONFIG.FEATURE_ONBOARDING);
+    logBreadcrumb('onboarding', 'finishOnboardingAction: dispatching getTutorialDataAction');
```
mdjastrzebski, line 339:
```suggestion
    logBreadcrumb('onboarding', 'finishOnboardingAction: dispatching fetchTutorialDataIfNeededAction');
```

src/actions/onboardingActions.js
```diff
@@ -340,11 +336,11 @@ export const finishOnboardingAction = (retry?: boolean) => {
...
+    await dispatch(fetchTutorialDataIfNeededAction());
 
     const { onboarding: { tutorialData } } = getState();
-    const routeName = tutorialData && enableOnboarding ? TUTORIAL_FLOW : HOME;
+    const routeName = tutorialData ? TUTORIAL_FLOW : HOME;
```
mdjastrzebski, line 343: What about case, when we first fetch onboarding data and save it into redux, but afterwards that tutorial data is disabled on the backend. Would suggest keeping the `&& enableOnboarding` part for such case.

mdjastrzebski (approved):
LG 
Add one potential flow edge case to consider.

src/actions/onboardingActions.js
poocart: 👍

PR #2841, opened 2021-08-30
check undefined exchange options asset and log case

poocart:
https://linear.app/pillarproject/issue/PIL-1170/app-crash-on-swap-press

Based on Sentry issues for 3.5.3 (https://sentry.io/organizations/pillar-project-worldwide-ltd/releases/com.pillarproject.wallet%403.5.3%2B23475/?project=1294444&statsPeriod=14d&unselectedSeries=Healthy) it seems that app crash on Dashboard "Swap" press happens when options method gets `asset` as undefined.

While I was not able to replicate this I added additional check and log which we can move to internal build for our internal team to check that has crashes on this.

### DIFF
--- a/src/utils/assets.js
+++ b/src/utils/assets.js
@@ -332,10 +332,17 @@ export const isMatchingCollectible = (
 export const mapWalletAssetsBalancesIntoAssetsByAddress = (
   walletAssetsBalances: WalletAssetsBalances,
   chainSupportedAssets: Asset[],
-): AssetByAddress => mapValues(
-  walletAssetsBalances,
-  ({ address }: WalletAssetBalance) => findAssetByAddress(chainSupportedAssets, address),
-);
+): AssetByAddress => {
+  const walletAssets = Object.keys(walletAssetsBalances)
+    .map((assetAddress) => findAssetByAddress(chainSupportedAssets, assetAddress));
+
+  return walletAssets.reduce((assetsByAddress, asset) => {
+    // filter out no longer supported
+    if (!asset) return assetsByAddress;
+
+    return { ...assetsByAddress, [addressAsKey(asset.address)]: asset };
+  }, {});
+};
 
 export const sortSupportedAssets = (
   supportedChainAssets: AssetsPerChain,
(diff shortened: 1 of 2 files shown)


src/utils/assets.js
```diff
@@ -332,10 +332,17 @@ export const isMatchingCollectible = (
 export const mapWalletAssetsBalancesIntoAssetsByAddress = (
   walletAssetsBalances: WalletAssetsBalances,
   chainSupportedAssets: Asset[],
-): AssetByAddress => mapValues(
-  walletAssetsBalances,
-  ({ address }: WalletAssetBalance) => findAssetByAddress(chainSupportedAssets, address),
-);
+): AssetByAddress => {
+  const walletAssets = Object.keys(walletAssetsBalances)
+    .map((assetAddress) => findAssetByAddress(chainSupportedAssets, assetAddress));
```
mdjastrzebski, line 337: Consider using `mapNotNil` from `utils/array` to remove not founds assets at this stage.

src/utils/assets.js
```diff
@@ -332,10 +332,17 @@ export const isMatchingCollectible = (
...
+
+  return walletAssets.reduce((assetsByAddress, asset) => {
+    // filter out no longer supported
+    if (!asset) return assetsByAddress;
+
+    return { ...assetsByAddress, [addressAsKey(asset.address)]: asset };
+  }, {});
```
mdjastrzebski, line 344: Consider simplifying with `buildRecord` from `utils/object`
```return buildRecord(walletAssets, a => addressAsKey(a.address), a => a)```

File: src/utils/assets.js [hunk 3]
```diff
@@ -332,10 +332,17 @@ export const isMatchingCollectible = (
 export const mapWalletAssetsBalancesIntoAssetsByAddress = (
   walletAssetsBalances: WalletAssetsBalances,
   chainSupportedAssets: Asset[],
-): AssetByAddress => mapValues(
```
mdjastrzebski, line 335: Alternatively to two other comments in this file, consider wrapping `maValues` call with `omitNilProps` from `utils/object`

mdjastrzebski (approved):
LG, suggested using some existing helper functions to simplify the code, IMO just applying `omitNilProps` would be the base solution.

src/utils/assets.js
poocart: Yeah, but then it's double iteration through array compared to `reduce` below which both builds the object and rejects assets that are not found 🤔

mdjastrzebski (approved):

PR #2844, opened 2021-08-31
PIL-1148: Feature/login rework

poocart:
https://linear.app/pillarproject/issue/PIL-1148/loading-screen-re-work

Includes:
- Fixes/improvements for wallet decrypt methods.
- Reordered `loginAction` flows.
- Other login/decrypt related flows improvements.


src/actions/etherspotActions.js
```diff
@@ -102,6 +103,8 @@ export const connectEtherspotAccountAction = (accountId: string) => {
 
 export const initEtherspotServiceAction = (privateKey: string) => {
   return async (dispatch: Dispatch, getState: GetState) => {
+    await dispatch(updateFcmTokenAction());
```
mdjastrzebski, line 106: Are FCM & Etherspot related together? Not sure if updating FCM token should be part of etherspot service init.

src/actions/appActions.js
```diff
@@ -212,8 +217,15 @@ export const initAppAndRedirectAction = () => {
...
 
-export const setupSentryAction = (user: ?Object, wallet: Object) => {
+export const setupLoggingServicesAction = () => {
   return (dispatch: Dispatch, getState: GetState) => {
+    if (__DEV__) return;
+
+    const {
+      user: { data: user },
+      wallet: { data: wallet },
+    } = getState();
```
mdjastrzebski, line 227:
```suggestion
const user = getState().user.data?.username;
const keyWalletAddress = getState().wallet.data.wallet?.keyWalletAddress;
```

Seems like all of the unwrapping code here and below be simplified here.

src/utils/wallet.js
```diff
@@ -48,7 +57,7 @@ export function generateWordsToValidate(numWordsToGenerate: number, maxWords: nu
   return chosenWords;
 }
 
-export async function getSaltedPin(pin: string, dispatch: Function): Promise<string> {
+export async function getSaltedPin(pin: string, dispatch: Dispatch): Promise<string> {
```
mdjastrzebski, line 60: This function feels weird because of getting that redux `Dispatch`. Consider changing the parameters to `deviceUniqueId, pin` and doing the `DeviceInfo` checks in the call site.

src/utils/wallet.js
```diff
@@ -194,21 +203,78 @@ export async function getWalletFromStorage(storageData: Object, dispatch: Dispat
...
   return new ethers.Wallet(privateKey);
 }
 
-export function constructWalletFromMnemonic(mnemonic: string): Object {
+export function constructWalletFromMnemonic(mnemonic: string): ethers.Wallet {
   return ethers.Wallet.fromMnemonic(mnemonic);
 }
 
-export async function getPrivateKeyFromPin(pin: string, dispatch: Dispatch) {
+export async function decryptWalletFromStorage(pin: string, dispatch: Dispatch): Promise<ethers.Wallet> {
```
mdjastrzebski, line 214: Similar to comment for `getSaltedPin`: consider replacing props with `pin, deviceUniqueId` and do the DeviceInfo check at call site. Passing Dispatch to utils feels a bit mixing responsibilities.

mdjastrzebski (approved):
Looks good as far as I can grasp what is happening here ;-)

src/actions/etherspotActions.js
poocart: Yes and no. FCM token is used for Etherspot session creation and we want to make sure it's updated before Etherspot init. Actually just looking at the code now and can be moved out of `initEtherspotServiceAction` just to not confuse things.

PR #2846, opened 2021-09-01
PIL-1146 Check fallbackSource prop on FastImage for possible technical Debt

divyaBacancy:
Pil1146


src/components/Image/Image.js
```diff
@@ -36,15 +41,15 @@ const getValidSource = (source: FastImageSource | number) => {
...
   static resizeMode = FastImage.resizeMode;
   static priority = FastImage.priority;
   static cacheControl = FastImage.cacheControl;
   static preload = FastImage.preload;
 
   render() {
-    const { source, ...rest } = this.props;
-    return <FastImage source={getValidSource(source)} {...rest} />;
+    const { source, fallback, fallbackSource, ...rest } = this.props;
+    return <FastImage source={getValidSource(source)} fallback={fallback} defaultSource={fallbackSource} {...rest} />;
```
mdjastrzebski, line 52:
```suggestion
    const { source, fallbackSource, ...rest } = this.props;
    return <FastImage source={getValidSource(source)} fallback={!!fallbackSource} defaultSource={fallbackSource} {...rest} />;   
```
We can drop fallback prop and just use `fallbackSource` to pass `fallback` value to `FastImage`

mdjastrzebski (changes requested):
Please drop `fallback` prop and use `fallbackSource` (!= null) to decide if fallback should be used.

src/components/Image/Image.js
divyaBacancy: Done.

src/components/Image/Image.js
```diff
@@ -24,6 +24,10 @@ import type { FastImageProps, FastImageSource } from 'react-native-fast-image';
 
 export type ImageProps = FastImageProps;
 
+type Props = {
+  fallbackSource?: string,
+}
+
```
mdjastrzebski, line 30: The idiomatic way would be to extend ImageProps with `fallbackSource` prop:

```suggestion
export type ImageProps = {|
  ...FastImageProps,
  fallbackSource?: ImageSource,
|};
```
ImageSource should be imported from utils/types-react-native.

src/components/Image/Image.js
```diff
@@ -36,15 +40,17 @@ const getValidSource = (source: FastImageSource | number) => {
 // This can be refactored to FC as soon as we update Styled Componenents.
 
 // eslint-disable-next-line react/prefer-stateless-function
-class Image extends React.Component<ImageProps> {
+class Image extends React.Component<ImageProps, Props> {
```
mdjastrzebski, line 43: It looks like you passed `Props` type to generic param for `State` type:
```
React$Component<Props, State = void>
```

Instead of that, please see the comment next to `Props` definition

mdjastrzebski (changes requested):
Found some type issues, please fix

mdjastrzebski (approved):
Looks good 👍🏻

PR #2849, opened 2021-09-03
PIL-1142-added a package to render svg image using webview

vignesha22:
Added a package called react-native-remote-svg which uses webview to render svg image correctly the image of the nfts
PHOTO-2021-09-03-13-04-16
PHOTO-2021-09-03-13-04-17
PHOTO-2021-09-03-13-05-18


src/screens/Assets/collectibles/CollectibleListItem.js
```diff
@@ -45,7 +45,7 @@ function CollectibleListItem({ title, iconUrl, onPress, width }: Props) {
       <IconContainer width={width}>
         {!!iconUrl && (
           <>
-            {isSvg && <SvgCssUri uri={iconUrl} width="100%" height="100%" />}
+            {isSvg && <SvgImage source={{ uri: iconUrl }} style={[{ width: 100, height: 100 }]} />}
```
poocart, line 48:
```suggestion
            {isSvg && <SvgImage source={{ uri: iconUrl }} style={[{ width: 100, height: 100 }]} />}
```

Hm, it was 100% dimensions so I guess you want that to be "100%"?

vignesha22:
I thought the area in that thumbnail suits 100 by 100 rather than 100% and
the result wasn’t different to the previous size in that thumbnail area so
left it like that.

mdjastrzebski (changes requested):
@vignesha22 I wonder do we need to use `react-native-remote-svg` package? My biggest concern is that it uses webview which probably is quite "heave" solution, and the package does not seem to be maintained any more.

From docs it looks like `react-native-svg`, the main SVG package for RN supports displaying remote SVGs: https://www.npmjs.com/package/react-native-svg#use-with-content-loaded-from-uri

vignesha22:
It was using that package before but in android that package isn’t working
or isn’t rendering svg as it should be and the output was blizzard to the
expected output. That’s y Deimantas suggested to use web view which only
gave the expected result on android

mdjastrzebski:
@vignesha22 that's weird because atm we are using `react-native-svg` to render SVGs, although local files. Could you double check that? RNSVG does not indicate there are any issue with android regarding remote files.

vignesha22:
> @vignesha22 that's weird because atm we are using `react-native-svg` to render SVGs, although local files. Could you double check that? RNSVG does not indicate there are any issue with android regarding remote files.

nope still output coming like this with SvgCssUri from react-native-svg package
PHOTO-2021-09-06-17-21-32

mdjastrzebski:
@vignesha22 Thanks for trying, my guess is that RNSvg does not support some elements of SVG hence the crippled output. Since these SVGs are beyond our control, then we have to deal with them as they are.

src/components/CollectibleImage/CollectibleImage.js
```diff
@@ -19,23 +19,18 @@
...
   const { uri } = props.source;
   if (isSvgImage(uri)) {
     return (
-      <SvgCssUri
-        uri={uri}
-        width={props.width || '100%'}
-        height={props.height || '100%'}
-        {...props}
-      />
+      <SvgImage {...props} style={[{ width: props.width || '100%', height: props.height || '100%' }, props.style]} />
```
mdjastrzebski, line 33: It seems we're passing width and height both as props and style, is one or the other sufficient?

src/screens/Assets/collectibles/CollectibleListItem.js
```diff
@@ -45,7 +45,7 @@ function CollectibleListItem({ title, iconUrl, onPress, width }: Props) {
       <IconContainer width={width}>
```
mdjastrzebski, line 45: Because RNRemoveSvg is using webview, this might be a potential bottleneck, since this component is a list item, which might cause perf issues for users with many NFTs. @vignesha22 could your try to check performance for user having say 30 NFTs, by using mock data or repeating existing items enough times?

vignesha22:
When I searched why it is, in one post I read that this package doesn’t
support blur and filters on svg which is the case here maybe that’s why the
output image is so distorted I guess

vignesha22:
Yeah will test that

src/screens/Assets/collectibles/CollectibleListItem.js
vignesha22: yeah if we have more than 5 itself the app crashes so yeah you are right. Do you have any other solution rather than using webview?

mdjastrzebski: That's what I was afraid of. Seems like we do not have perfect solution here. Maybe we could use web-view high res for viewing single collectible in full screen view and use the RNSvg for list view.

@IAmKio, @poocart, @vignesha22  wdyt?

vignesha22:
Yeah and for the time being we could take off the thumbnail and just
display the title because we don’t want to display distorted image on the
thumbnail that would be weird until we get a perfect solution.

src/screens/Assets/collectibles/CollectibleListItem.js
poocart: That's the thing, it does not render as it supposed to using RNSvg. Agree with 2 libraries serving list and preview screens, but the main idea is to fix the broken ones and showing on the list broken is not a solution, maybe we can show webview svg only for broken ones identifying them by contract? That would narrow the list of using webview SVGs only.

poocart: Also maybe is there any way to lower the perfomance of that webview svg lib?

mdjastrzebski: From my perspecting app crashing from memory exhaustion is worse then broken SVG. Webview path is unlikely to succeed in any list environment, because of webview memory footprint. I think we need to research some more solutions. Couple of ideas:
- some other RN SVG library
- caching low-res PNG renders of SVG for list usage (front end side or back end side)

poocart: Was thinking about the latter too, but that's a significant amount of work. I see it residing in back-end, but that would require effort that is not cost-effective for these few NFTs to support. Also these Uniswap V3 SVGs are moving so still images is OK, but wouldn't completely deliver what expected and transforming those into GIFs is too much fuss for sure.

mdjastrzebski: If webview renders moving (animated?) SVG for whole screen viewing then we would not need to use backend rendering for that, just pre-render them for non-animated, lower-res list items.

poocart: @vignesha22 thoughts?

vignesha22:
I would like to see all the nfts names on the lists since if we have a lot
we can show it to the users which they won’t have problem in finding since
thumbnails consume space on each screen but this would have a design change
or separate svgs and images separately and show thumbnails with image
formats and others inside a list and show svgs in detail view… I couldn’t
think of anything else since it’s not uniform the nfts images as one is svg
and others are still images

PR #2850, opened 2021-09-03
PIL-1175:- Remove "0 ETH" in WC Modal

divyaBacancy:
https://user-images.githubusercontent.com/66996635/131999226-8ddb1a24-efed-4897-b37d-99cb1407908d.mov


poocart (approved):

src/components/modern/LargeFiatTokenValueView.js
```diff
@@ -55,7 +55,7 @@ function LargeFiatTokenValueView({ value, assetAddress, chain, symbol, style }:
   const fiatCurrency = useFiatCurrency();
   const colors = useThemeColors();
   const rates = useChainRates(chain);
-  if (!value || !assetAddress) return null;
+  if (!value || value.isZero() || !assetAddress) return null;
```
mdjastrzebski, line 58: This check should not be done in LargeFiatTokenValue, since there might be legit cases when you would want to show "0 ETH". Please move it to calling component.

mdjastrzebski (changes requested):

src/components/modern/LargeFiatTokenValueView.js
poocart: Maciej I thought the same, but zero value transactions are data transactions so it kinda makes sense to have it at this level, but not saying it doesn't make sense to move it level up.

mdjastrzebski: This control is general purpose, not limited to tx value, so I would lean to moving the check up to the call site.

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
```diff
@@ -120,13 +120,15 @@ function TransactionRequestContent({ request, onConfirm, onReject }: Props) {
...
       </Text>
 
-      <LargeFiatTokenValueView
-        value={value}
-        assetAddress={assetAddress}
-        chain={chain}
-        symbol={symbol}
-        style={styles.tokenValue}
-      />
+      {!value?.isZero() && (
```
mdjastrzebski, line 123: In current code null value causes component to render (but then component renders itself to null), so maybe we could use a more straightforward condition.
```suggestion
      {value?.gt(0) && (
```

mdjastrzebski (approved):
Small suggestion to improve style, otherwise looks good 👍🏻

src/screens/WalletConnect/CallRequest/TransactionRequestContent.js
divyaBacancy: Done

PR #2853, opened 2021-09-06
Feature: new asset selector

mdjastrzebski:
Scope:
- [x] New Asset selector modals: summary one and full list one
- [x] Summary asset selector has extracted content component to enable using it in Send Flow as regular screen
- [x] New re-usable TokenListItem & CollectibleListItem components


src/utils/collectibles.js
```diff
@@ -32,8 +35,18 @@ export function buildCollectibleKey(contractAddress: string, id: string) {
...
 
-export function getCollectibleKey({ contractAddress, id }: Collectible) {
-  return buildCollectibleKey(contractAddress, id);
+export function getCollectibleKey(collectible: ?Collectible) {
+  if (!collectible) return '';
+  return buildCollectibleKey(collectible.contractAddress, collectible.id);
+}
+
+export function defaultSortCollectibles(collectibles: Collectible[]): Collectible[] {
+  return orderBy(collectibles, [({ name }) => name?.trim().toLowerCase()], ['asc']);
```
poocart, line 44:
```suggestion
  return orderBy(collectibles, [({ name }) => name?.trim()?.toLowerCase()], ['asc']);
```

src/components/Modals/AssetSelectorModal/ChainSectionFooter.js
```diff
@@ -0,0 +1,80 @@
...
+  onPress: () => mixed,
+|};
+
+function ChainSectionFooter({ showMore, onPress }: Props) {
+  const { t } = useTranslation();
+  const colors = useThemeColors();
+
+  return (
+    <Container>
+      {showMore && (
```
poocart, line 44: Not fully get it, if there's no `showMore` then simply `Container` with extra space is added?

mdjastrzebski: Extra space is always needed, because it separate this section from the next one.

src/components/Modals/ChainAssetSelectorModal/ChainAssetSelectorModal.js
```diff
@@ -0,0 +1,134 @@
...
+  const colors = useThemeColors();
+
+  const close = () => {
+    Keyboard.dismiss();
+    modalRef.current?.close();
+  };
+
+  const selectToken = (token: AssetOption) => {
+    close();
+    onSelectToken?.(token);
```
poocart, line 76: Just wondering if safe call really needed if prop is mandatory? If you think it's still good to be safe then consider adding it here too https://github.com/pillarwallet/pillarwallet/pull/2853/files#diff-a51ac9423c64ec11b18f0e0164cf3cb2aa4fd68d6c15f3ad943144ed608e4bf2R63

mdjastrzebski: Removed null check

src/utils/assets.js
```diff
@@ -277,6 +278,16 @@ export const getAssetOption = (
   };
 };
 
+export const getAssetOptionKey = (option: ?AssetOption) => {
+  if (!option) return '';
+  return `${option.chain}-${option.address}`;
```
poocart, line 283:
```suggestion
  return `${option.chain}-${option.address ?? option.contractAddress}`;
```

Although the `AssetOption ` type definitely needs to be visited.

poocart (approved):
🔥

src/components/Modals/AssetSelectorModal/ChainSectionFooter.js
poocart: Ok cool, just the fact that the space is produced from empty content components makes me consider if it can be moved somewhere outside?

mdjastrzebski: Not really, as the space is needed after every section of the section list. If moved to the item, it would be repeated after each item, if moved to the header it would be also applied before first element. Hence the reasonable solution is to put in the section footer as I've done here.

PR #2857, opened 2021-09-09
Refactor: add chain field to Asset type

mdjastrzebski:
Motivation:
`Asset` object is usually meaningless with accompanying `chain` information. Currently this requires passing separately asset and chain to components. Similar entiries, `Collectible` and `AssetOption` already include chain field.

Scope:
- [x] Added `chain` to `Asset` type
- [x] Asset initialization to include `chain` field
- [x] Saved data migration

### DIFF
--- a/src/utils/etherspot.js
+++ b/src/utils/etherspot.js
@@ -48,7 +48,7 @@ import { firebaseRemoteConfig } from 'services/firebase';
 import { isEtherspotAccount } from 'utils/accounts';
 import { findAssetByAddress } from 'utils/assets';
 import { fromEthersBigNumber } from 'utils/bigNumber';
-import { nativeAssetPerChain } from 'utils/chains';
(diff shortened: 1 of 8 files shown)


src/utils/etherspot.js
```diff
@@ -180,10 +180,14 @@ export const parseEtherspotTransactionState = (state: GatewayBatchStates): ?stri
   }
 };
 
-export const parseTokenListToken = ({ address, name, symbol, decimals, logoURI }: TokenListToken): Asset => {
+export const parseTokenListToken = (
+  { address, name, symbol, decimals, logoURI }: TokenListToken,
+  chain: Chain,
+): Asset => {
```
poocart, line 186: `TokenListToken` contains `chainId`, should we reuse that?

poocart: Fixed.

poocart (approved):

PR #2858, opened 2021-09-09
 PIL-1141 Rename "Transaction Reverted"

divyaBacancy:
https://user-images.githubusercontent.com/66996635/132699119-42427ecb-ffa7-45e3-8795-525c5c740620.mov


src/screens/WalletConnect/Home/WalletConnectHome.js
```diff
@@ -100,33 +99,23 @@ function WalletConnectHome() {
   // Note: in order to achieve multicolumn layout, we group n normal items into one list row item.
   const renderListRow = (items: WalletConnectCmsApp[]) => <ListRow>{items.map(renderItem)}</ListRow>;
 
-  const openUrl = async (url: string | null) => {
+  const openAppUrl = (url: string | null) => {
```
mdjastrzebski, line 102: Seems like we no longer shoe anything if url is falsy, is that intended?

src/components/SendAsset/SendAsset.js
```diff
@@ -156,9 +163,12 @@ const SendAsset = ({
...
-        message: t('toast.cannotSendAsset'),
-        emoji: 'woman-shrugging',
-        supportLink: true,
+        message: t('toast.transactionReverted'),
+        emoji: 'hushed',
+        autoClose: false,
+        onPress: () => {
+          openUrl(transactionRevertedAricleURL);
+        },
       });
```
mdjastrzebski, line 172: Seems like this code is repeated couple of times, and each call site needs to get firebase remote config settting, etc. Could we move it a common function in utils? e.g. `showTransactionRevertedToast` in `utils/transactions`?

src/screens/WalletConnect/Home/WalletConnectHome.js
divyaBacancy: Done.

src/components/SendAsset/SendAsset.js
divyaBacancy: Done.

src/utils/inAppBrowser.js
```diff
@@ -46,3 +50,19 @@ export const openInAppBrowser = async (url: string, options?: InAppBrowserOption
...
+  if (url) {
+    await openInAppBrowser(url).catch(showServiceLaunchError);
+  } else {
+    showServiceLaunchError();
+  }
+};
+
+export const showServiceLaunchError = () => {
+  Toast.show({
+    message: t('toast.cryptoPurchaseLaunchFailed'),
```
poocart, line 64: Method name and message does not relate, I would suggest either rename `showServiceLaunchError` to `showCryptoPurchaseLaunchError` or change the message

divyaBacancy: Done.

poocart (approved):

PR #2867, opened 2021-09-13
PIL-1180: wallet auth check callback fix

poocart:
https://linear.app/pillarproject/issue/PIL-1180/bug-wallet-backup-screen-doesnt-load

Fixes successful auth check callback param.


IAmKio (approved):
Thank you 🙏🏾

PR #2868, opened 2021-09-14
Pil 1159/fixed archanova migration process to etherspot

vignesha22:
the wallet object was receiving as a string so i initialised the wallet to the since on the pin confirm screen the parameters of the wallet was any so added a check if its not an object convert it to the wallet object and proceed
PHOTO-2021-09-14-16-36-57


poocart (changes requested):
What is the main reason why wallet is not present here? Please take a look at higher levels why we end up not having wallet object here 🙏

poocart:
Also I think the issue here is related to issue that was fixed with this PR https://github.com/pillarwallet/pillarwallet/pull/2867/files.

vignesha22:
Yeah it’s the same and tested it out too

PR #2883, opened 2021-10-01
Improvement/pil 1190 added support conversations

vignesha22:
Added Support Conversations option if any ongoing chats are present

nt

### DIFF
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -577,7 +577,8 @@
       "settings": "Settings",
       "addressBook": "Address book",
       "supportChat": "Support chat",
-      "storybook": "Storybook"
+      "storybook": "Storybook",
+      "supportConversations": "Support Conversations",
     },
     "footer": {
       "privacyPolicy": "Privacy Policy",
--- a/src/locales/en/common.json
+++ b/src/locales/en/common.json
@@ -578,7 +578,7 @@
       "addressBook": "Address book",
       "supportChat": "Support chat",
       "storybook": "Storybook",
-      "supportConversations": "Support Conversations",
+      "supportConversations": "Support Conversations"
     },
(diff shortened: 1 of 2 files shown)


src/locales/en/common.json
```diff
@@ -577,7 +577,8 @@
       "settings": "Settings",
       "addressBook": "Address book",
       "supportChat": "Support chat",
-      "storybook": "Storybook"
+      "storybook": "Storybook",
+      "supportConversations": "Support Conversations"
```
IAmKio, line 581:
```suggestion
      "supportConversations": "Support conversations"
```

For consistency

src/locales/en/common.json
```diff
@@ -577,7 +577,8 @@
       "settings": "Settings",
       "addressBook": "Address book",
       "supportChat": "Support chat",
```
IAmKio, line 579:
```suggestion
      "supportChat": "Support",
```

"Support chat" and "Support conversations" are too similar. Because "Support chat" just opens the main Instabug dialog and we now have a dedicated item for support conversations - we should rename this to just "Support".

src/screens/Menu/Menu/Menu.js
```diff
@@ -52,10 +52,20 @@ const Menu = () => {
...
   const navigation = useNavigation();
 
+  const [repliesFlag, setRepliesFlag] = useState(false);
+
+  Replies.hasChats(previousChat => {
+    if (previousChat) {
+      setRepliesFlag(true);
+    }
+  });
+
```
IAmKio, line 62: Please remove the extra line here

IAmKio (changes requested):
A few cosmetic changes and we should be good to merge this after the RC patch has been completed.

PR #2894, opened 2021-10-13
PIL-1192 changed magic token screen to display after getting transaction hash

vignesha22:
Sometimes when we click "View on Blockchain" the transaction details isnt getting populated and depends on the traffic at that time to submit the transaction on backend in order to tackle this, we have to show the magic screen after getting the transaction hash from backend for "View on blockchain" fn to work on all situations


IAmKio:
@vignesha22 remember to put the ticket ID at the beginning of the title.

PR #2902, opened 2021-11-02
PIL-1252 Issue on POAP NFT fee fetching

vignesha22:
POAP NFT on xdai wasn't able to send since the default chain on sendCollectibleConfirm to estimate fee was ethereum which is why we got transaction reverted from backend
b09037832c97612696447872ffff1a2dfa5f4c164716d80c46ddbfea1f89c0f6
d


divyaBacancy (approved):
LGTM

PR #2903, opened 2021-11-09
PIL-1231 - In-app exchange issue

vignesha22:
https://user-images.githubusercontent.com/82584664/140875003-71b1a50f-dde3-4904-9541-7c94de70cadc.mp4


IAmKio:
@vignesha22 just a warning that this is a naive fix for now and we'll likely have to revisit this later if a user needs to transfer a small value of a high value token. I think personally it's unlikely but just need to get this on record.

IAmKio (changes requested):
@vignesha22 Could you please annotate your code to explain why we're doing this - to anyone looking at this, it would be unclear why we're deciding to truncate the decimals of a token to 15.

IAmKio (approved):

PR #2905, opened 2021-11-10
PIL-1245 changed walletConnect tab nav color to purple

vignesha22:
image
image


src/components/layout/TabBar.js
```diff
@@ -98,5 +98,5 @@ const Underline = styled.View`
   border-top-width: ${Platform.OS === 'android' ? '3px' : '6px'};
   border-top-left-radius: 3px;
   border-top-right-radius: 3px;
-  border-color: ${({ theme }) => theme.colors.primaryAccent130};
+  border-color: #A020F0;
```
IAmKio, line 101: Could you check that these colour is not stored in the theme?

IAmKio (changes requested):
@vignesha22 please check that this hex is stored in the theme files first.

src/components/layout/TabBar.js
vignesha22: yeah i checked and one is there on the dark theme but its not there on light theme. but we needed the same color irrespective on the dark mode or light mode right?

IAmKio: It looks this way yes. If it exists for dark but not light, please add it for the light theme and use the theme reference instead.

vignesha22: yeah okay

IAmKio: Thanks 🙏

IAmKio (approved):
LGTM

PR #2909, opened 2021-11-12
Fix Backup Screen not coming after the Pin in logout flow

divyaBacancy:
This issue was occurring as the mnemonic phrase was not coming and In BackupPhrase screen there was a condition that till phrase is not coming it will show the auth flow. Fixed by setting the correct value in callback function for mnemonic phrase. 

https://user-images.githubusercontent.com/66996635/141418095-a468e247-be93-452f-9f1b-79b5f5147e9b.mov


IAmKio:
@arddluma FYI

IAmKio (approved):
Thanks

PR #2923, opened 2021-11-19
Add albanian language

arddluma:
Add albanian language and add missing words for Bosnian language.


IAmKio (changes requested):
@arddluma Could you please also update the `localesConfig.js` with the new locale please? File: `src/configs/localeConfig.js`. You'll need to add the `require`s for the new language and add it to `supportedLanguages` and `localTranslations`. 

Please run the app and make sure it works as expected. Cheers!

IAmKio:
@arddluma could you please add the entry for the country flag here: `src/components/Flag/Flag.js` - the flag should then appear.

PR #2926, opened 2021-11-22
PIL-1243 - Display ens name on confirm screen

vignesha22:
image


src/components/SendAsset/SendAsset.js
```diff
@@ -166,6 +166,8 @@ const SendAsset = ({
 
     setSubmitPressed(true);
 
+    selectedContact.ensName = await lookupAddress(selectedContact.ethAddress);
```
IAmKio, line 169: @vignesha22 the function `lookupAddress` in `common.js` possibly returns null, what happens in this scenario?

Yash-Chaubal (approved):

IAmKio (approved):

PR #2933, opened 2021-11-25
Feature/add workflows for feature branches

vnovacev:
If new feature branch with the term "UI" or "ui" in its name is created or new code is pushed into it, new test app builds will be created (both ipa and apk). Files will be uploaded to App Center with the name for ex. "3.0.20543-ui" and also they will be available for download from S3 bucket. 

When creation of builds is finished, info messages with S3 bucket download links will be send to "s3-urls" Slack channel.


arddluma (approved):
Looks good to me

PR #2951, opened 2021-12-08
PIL-1264- Fetched Polygon NFT from etherspot sdk and fixed send asset collectible

vignesha22:
https://user-images.githubusercontent.com/82584664/145187577-f06f1f6e-b27b-4cd4-91e7-412829021139.mp4


src/actions/collectiblesActions.js
```diff
@@ -91,6 +92,22 @@ export const parseCollectibleFromOpenSeaAsset = (asset: OpenSeaAsset): Collectib
...
 
+export const parsePolygonCollectibleFromEtherspot = (asset: NftList): Collectible => {
+  return {
+    id: asset.items[0].tokenId,
+    name: asset.items[0].name,
+    description: asset.description,
+    icon: asset.items[0].image,
+    iconUrl: asset.items[0].image,
+    image: asset.items[0].image,
+    imageUrl: asset.items[0].image,
```
IAmKio, line 103: If items is null then this will throw an error, could we add a check here that there is an array item that exists?

src/services/etherspot.js
```diff
@@ -757,6 +758,24 @@ export class EtherspotService {
     }
   }
 
+  async getNftList(
+    chain: Chain,
+    address: string,
+  ): NftList | null {
+    const sdk = this.getSdkForChain(chain);
+
+    if (!sdk) return null;
```
IAmKio, line 767: Please log the error here

src/services/etherspot.js
```diff
@@ -757,6 +758,24 @@ export class EtherspotService {
...
+
+    try {
+      return sdk.getNftList({
+        account: address,
+      });
+    } catch (error) {
+      reportErrorLog('EtherspotService getNftList failed', { chain, error });
+      return null;
+    }
```
IAmKio, line 776: Any reason we're using `try / catch` here instead of using a .catch here? `try / catch` should be used as a final resort...

IAmKio (changes requested):
Some minor changes - thanks!

src/services/etherspot.js
vignesha22: cant return null on .catch

src/services/etherspot.js
vignesha22: Sure

src/actions/collectiblesActions.js
vignesha22: yeah sure

src/services/etherspot.js
IAmKio: Are you sure about that? I'm quite sure that if you return `null` inside a catch block, the `promise`d function would return `null`... or any other value that is `return`ed in the `catch` block...

vignesha22: catch block as of now will return null but if its inside .catch, its basically inside a callback function and .catch return type is void so thats y i think it wont return to the calling place

IAmKio: I've just been doing a bit of research on this because Typescript inferring a `void` return type seems like a very opinionated action within a function (`.catch`) that has always been able to `return` anything that the developer wants to... i know that if you don't return anything inside a `.catch` block, Typescript adds a `void` union to the block itself but that's if you don't `return` anything. @Yash-Chaubal might you know anything about this scenario?

vignesha22: i guess i was wrong abt that just now searched and yeah it is possible to return anything. So will change it to .catch

IAmKio (approved):
LGTM

src/actions/collectiblesActions.js
```diff
@@ -91,6 +92,22 @@ export const parseCollectibleFromOpenSeaAsset = (asset: OpenSeaAsset): Collectib
...
+    icon: asset.items[0]?.image,
+    iconUrl: asset.items[0]?.image,
+    image: asset.items[0]?.image,
+    imageUrl: asset.items[0]?.image,
+    contractAddress: asset.contractAddress,
+    tokenType: ASSET_TYPES.COLLECTIBLE,
+    chain: CHAIN.POLYGON,
+    isLegacy: asset.nftVersion === '1.0',
+  };
+};
```
Yash-Chaubal, line 109: Would it matter if any of these fields are let as undefined?

Yash-Chaubal (approved):

src/actions/collectiblesActions.js
```diff
@@ -91,6 +92,22 @@ export const parseCollectibleFromOpenSeaAsset = (asset: OpenSeaAsset): Collectib
   };
 };
 
+export const parsePolygonCollectibleFromEtherspot = (asset: NftList): Collectible => {
+  return {
+    id: asset.items[0]?.tokenId,
+    name: asset.items[0]?.name,
+    description: asset.description,
+    icon: asset.items[0]?.image,
```
Yash-Chaubal, line 100: Would it matter if any of these fields are left as undefined?

PR #2953, opened 2021-12-10
PIL-1279 - Remote Config for Wert

vignesha22:
Added firebase config variables and default values for wert widget


IAmKio (approved):
LGTM 🙏

PR #2958, opened 2021-12-15
PIL-1250 - autoFocus to Send/Swap Select Token Search

vignesha22:
Show Keyboard upon clicking 'Select Token' button on Send/Swap Screens


Yash-Chaubal (approved):
Lgmt!

PR #2961, opened 2021-12-22
PIL-1272-Implemented Redux Saga

Yash-Chaubal:
- Added `redux-saga` to the project.
- Added some new redux code to make sure the sagas are working fine.
- Added paths to the redux folder in `.babelrc`, `tsconfig.json`, and `.flowconfig` - and also one for `src` since I figured it'll be useful.
- Changed `configureStore.js` so that `store` and `persistor` are exported instead of a function. This will help prevent some of the reducers, and in my case sagas, from being called twice.
- Updated files that were affected by the above change (`App.tsx` and `WithThemeDecorator.js`).
- Added some lines for `eslint` to ignore, so that `yarn validate` would be happy.

Note: The new redux code I added is temporary and is just to make sure that the saga setup is working correctly. When I begin work on the Firestore implementation these will likely change.

None of these functions should be called in the project for the time being.


IAmKio:
@Yash-Chaubal could you remove the `console.log`'s please and use the logger utility that we're using throughout the codebase please?

Yash-Chaubal:
@IAmKio Done!

src/redux/reducers/firestore-reducer.tsx
```diff
@@ -0,0 +1,51 @@
...
+    logBreadcrumb(TAG, 'last sent:', state.lastSentDataUtc);
+  }
+  return updateObject(state, {});
+};
+
+export interface IReduxDataSentToFireStore extends IReduxActions {
+  payload: {
+    time: string;
+  };
+}
```
IAmKio, line 31: New line for styling consistency

src/redux/reducers/reducer-utilities.tsx
```diff
@@ -0,0 +1,20 @@
+import IObject from 'redux/interfaces/IObject';
+
+export const updateObject = <T,>(oldObject: T, newValues: T) => {
+  return Object.assign({}, oldObject, newValues);
+};
```
IAmKio, line 5: Could this just be:

```suggestion
export const updateObject = <T,>(oldObject: T, newValues: T) => Object.assign({}, oldObject, newValues);
```

src/redux/reducers/reducer-utilities.tsx
```diff
@@ -0,0 +1,20 @@
...
+    } else {
+      return state;
+    }
+  };
+  return reducer;
+};
+
+export const updateChangesCounter = (counter: number | undefined): number => {
+  return (counter ?? 0) + 1;
+};
```
IAmKio, line 20: And the same for this maybe?

```suggestion
export const updateChangesCounter = (counter: number | undefined): number => (counter ?? 0) + 1;
```

src/redux/sagas/firestore-saga.tsx
```diff
@@ -0,0 +1,30 @@
+import { takeEvery, call, put, take, select } from 'redux-saga/effects';
+import { ReduxFirestoreType as t } from 'redux/redux-types/firestore-type';
+import { IReduxActions } from 'redux/interfaces/IReduxActions';
+import { IReduxDataSentToFireStore } from 'redux/reducers/firestore-reducer';
+import { logBreadcrumb } from 'utils/common';
+
+// Watcher Saga
+export default function* firestoreSaga() {
+  yield takeEvery(t.SEND_DATA_TO_FIRESTORE, sendDataToFirestoreSaga);
+}
```
IAmKio, line 10: Star / yield - threw me off for a second there! 😂

src/redux/sagas/root-saga.tsx
```diff
@@ -0,0 +1,6 @@
+import { fork } from 'redux-saga/effects';
+import firestoreSaga from './firestore-saga';
+
+export default function* rootSaga() {
+  yield fork(firestoreSaga);
```
IAmKio, line 5: I would say one-line this but i'm unsure about `yeild`ing on a single line...

src/services/etherspot.js
```diff
@@ -120,24 +112,26 @@ export class EtherspotService {
...
-      const env = networkName !== NetworkNames.Kovan ? EnvNames.MainNets : EnvNames.TestNets;
-      this.instances[networkName] = new EtherspotSdk(privateKey, { env, networkName, projectKey: PROJECT_KEY });
+    await Promise.all(
+      this.supportedNetworks.map(async (networkName) => {
+        const env = networkName !== NetworkNames.Kovan ? EnvNames.MainNets : EnvNames.TestNets;
+        this.instances[networkName] = new EtherspotSdk(privateKey, { env, networkName, projectKey: PROJECT_KEY });
+
+        if (fcmToken) {
+          try {
+            await this.instances[networkName].createSession({ fcmToken });
```
IAmKio, line 122: Do we know if `.createSession` is un`catch`able?

src/services/etherspot.js
```diff
@@ -120,24 +112,26 @@ export class EtherspotService {
...
+          } catch (error) {
+            reportErrorLog('EtherspotService network init failed at createSession', { networkName, error });
+          }
+        }
 
-      if (fcmToken) {
         try {
-          await this.instances[networkName].createSession({ fcmToken });
+          await this.instances[networkName].computeContractAccount({ sync: true });
```
IAmKio, line 129: Same with this re: try / catch

IAmKio (changes requested):
Some minor suggestions but other than that - it's looking good!

src/redux/sagas/root-saga.tsx
Yash-Chaubal: I'd keep it as is since this the root saga where I'd call other sagas as well. It's just that `firestoreSaga` is the only one that exists at the moment.

IAmKio: Thanks for that 👍

IAmKio (approved):
LGTM 🚀

PR #2968, opened 2022-01-11
PIL-1271-Sync Redux State with Firestore

Yash-Chaubal:
Syncs the redux state with firestore when the app loads.


src/constants/remoteConfigConstants.js
```diff
@@ -62,6 +62,7 @@ export const REMOTE_CONFIG = {
   FEATURE_WERT_COMMODITIES: 'feature_wert_commodities',
   FEATURE_WERT_CURRENCY: 'feature_wert_currency',
   FEATURE_WERT_COMMODITY: 'feature_wert_commodity',
+  FIRESTORE_SYNC_URL: 'app_api_user_endpoint',
```
IAmKio, line 65: Could you change this to `APP_API_USER_ENDPOINT` please and update the references elsewhere, this is just for consistency.

src/constants/remoteConfigConstants.js
```diff
@@ -109,5 +110,6 @@ export const INITIAL_REMOTE_CONFIG = {
   [REMOTE_CONFIG.FEATURE_WERT_COMMODITIES]: 'MATIC',
   [REMOTE_CONFIG.FEATURE_WERT_COMMODITY]: 'MATIC',
   [REMOTE_CONFIG.FEATURE_WERT_CURRENCY]: 'USD',
+  [REMOTE_CONFIG.FIRESTORE_SYNC_URL]: 'https://us-central1-pillar-project-1506420699556.cloudfunctions.net/api',
```
IAmKio, line 113: As above.

src/redux/sagas/firestore-saga.tsx
```diff
@@ -1,28 +1,82 @@
...
+): Promise<IReduxFirestoreSyncedState> => {
+  logBreadcrumb(TAG, 'Attempting to sync state with Firestore', state.firestore);
+
+  // Check if recently synced
+  if (lastSyncedUtc) {
+    const now = new Date();
+    const lastSynced = new Date(lastSyncedUtc);
+    if (isBefore(now, addHours(lastSynced, 1))) throw new Error('State has recently been synced');
+  }
+
```
IAmKio, line 62: This is a nice feature but it got me thinking about the support team and the event that they may need to see this data sooner than later... let's chat about options 👍

src/redux/sagas/firestore-saga.tsx
```diff
@@ -1,28 +1,82 @@
...
+  // Purge sensitive data from the redux state
+  let mutatedState = purgeSensitiveDataFromState(state);
+  if (!validateStateToSync(mutatedState)) throw new Error('State still contains sensitive data');
+
+  const firestoreUrl = firebaseRemoteConfig.getString(REMOTE_CONFIG.FIRESTORE_SYNC_URL);
+  logBreadcrumb(TAG, 'Firestore url:', firestoreUrl);
+
+  // Attempt sync reqeuest to Firebase
+  const { data } = await httpRequest.post(firestoreUrl, mutatedState, requestConfig);
```
IAmKio, line 71: I would also `.catch` and return `null` here, as some errors returned from the server can contain a data payload, which would result in the next line _looking_ valid. Here's my suggestion:

```suggestion
  const { data } = await httpRequest.post(firestoreUrl, mutatedState, requestConfig).catch(() => null);
```

This way, the next line would be falsey even something ended up in the catch block.

IAmKio (changes requested):
Some minor suggestions but overall - great work! 👍

IAmKio (approved):
LGTM! 👍

PR #2983, opened 2022-01-27
PIL-1289-Set PIN appearing twice during import

Yash-Chaubal:
(WIP) Added an option to the biometrics modal so it can handle phones with no biometrics. This will do the same thing as if the user pressed 'Cancel', but through `useEffect`.

This is a temporary solution as `beginOnboardingAction` is being called in both `Home.js` and `BiometricModal.js` with the same params being passed, but they do slightly different things - which is what I'm looking into now.

### DIFF
--- a/src/components/BiometricModal/BiometricModal.js
+++ b/src/components/BiometricModal/BiometricModal.js
@@ -19,7 +19,7 @@
 */
 /* eslint-disable i18next/no-literal-string */
 
-import React, { useRef, useCallback } from 'react';
+import React, { useRef, useEffect, useCallback } from 'react';
 import { View } from 'react-native';
 import styled from 'styled-components/native';
 import t from 'translations/translate';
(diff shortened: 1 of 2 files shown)


divyaBacancy:
Thanks @Yash-Chaubal for bringing this up, Everything looks great, But I have 2 concerns
1) When device is not supporting Biometric, then initial route screen was invoked again after Home screen (IOS -- PIN screen, Android - Permission) -- **solved**
2) As soon as we enter 12 words pass phrase, for a moment we get initial route screen of Onboarding stack is displayed (IOS -- PIN screen, Android - Permission) -- **not solved**

divyaBacancy:
Uploading TicketStatus.mov…

divyaBacancy:
@Yash-Chaubal If Biometric is not supported, 2 loaders are displayed till the wallet is created

Uploading LoaderIssue.mov…

Yash-Chaubal:
@divyaBacancy ah thanks, one of the spinners is the initial on the home screen, and the one in the middle is the spinner on the modal from the new code - I'll stop the older one so that it looks consistent

IAmKio (approved):
LGTM

divyaBacancy (approved):

PR #2985, opened 2022-01-27
PIL-1294 Keychain Issue

vignesha22:
Added Alert on catch block if the keychain can't store or get the key from the device


IAmKio:
@vignesha22 can we write a test please where we force the Keychain method to fail to ensure that the correct screen is displayed here please. Thanks

IAmKio (changes requested):
Please see my comment above.

src/utils/keychain.js
```diff
@@ -38,10 +47,44 @@ export type KeyChainData = {
...
+    text: t('error.failedKeychain.exitButtonText'),
+    onPress: () => RNExitApp.exitApp(),
+  });
+  accountAddress = accountAddress ?? etherspotService?.getAccountAddress(CHAIN.ETHEREUM);
+  if (accountAddress) {
+    buttons.push({
+      text: t('error.failedKeychain.supportButtonText'),
+      onPress: () => {
+        BugReporting.showWithOptions(BugReporting.reportType.feedback, [BugReporting.option.emailFieldOptional]);
+      },
```
IAmKio, line 63: Could this be:

```suggestion
      onPress: BugReporting.showWithOptions(BugReporting.reportType.feedback, [BugReporting.option.emailFieldOptional]),
```

IAmKio (changes requested):
I'm just wondering, where you've handled the catch in the keychain utilities, are any of these being fired on app startup or at least in `App.tsx`? This alert needs to be the very first thing that users will see if they launch the app and the app isn't supported...

vignesha22:
> I'm just wondering, where you've handled the catch in the keychain utilities, are any of these being fired on app startup or at least in `App.tsx`? This alert needs to be the very first thing that users will see if they launch the app and the app isn't supported...

As of my knowledge, I couldn't find any keychain package being used apart from biometric settings and in that GitHub issue page, the error occurred when the getGenericPassword fn was called and moreover couldn't find elsewhere which could be caused from keychain module

src/utils/keychain.js
vignesha22: yeah but since I wrote a test case, it will directly call this function and it will fail since the Instabug will not be initialised

IAmKio:
@vignesha22 could you invert this logic so we can see a screenshot of how this looks please?

vignesha22:
Simulator Screen Shot - iPhone 11 - 2022-02-03 at 01 17 59
Simulator Screen Shot - iPhone 11 - 2022-02-03 at 00 53 48

IAmKio:
@vignesha22 thanks for that - you've missed out the support button scenario. Check the ticket on Linear.

vignesha22:
Simulator Screen Shot - iPhone 11 - 2022-02-03 at 07 50 11

divyaBacancy (approved):

IAmKio (approved):

PR #2996, opened 2022-02-06
PIL-1307/support-conversations-menu-items

IAmKio:
- Reinstated menu items and corrected logic for support eligibility

### DIFF
--- a/src/screens/Menu/Menu/Menu.js
+++ b/src/screens/Menu/Menu/Menu.js
@@ -121,7 +121,7 @@ const Menu = () => {
         <MenuItem title={t('item.addressBook')} icon="contacts" onPress={goToInviteFriends} />
         <MenuItem
           title={enoughPlrBalance ? t('item.liveChatSupport') : t('item.emailSupport')}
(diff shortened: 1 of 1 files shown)


src/screens/Menu/Menu/Menu.js
```diff
@@ -111,15 +120,19 @@ const Menu = () => {
         />
         <MenuItem title={t('item.addressBook')} icon="contacts" onPress={goToInviteFriends} />
         {enoughPlrBalance ? (
-          <MenuItem title={t('item.liveChatSupport')} icon="message" onPress={goToSupportConversations} />
-        ) : (
+          <MenuItem title={t('item.liveChatSupport')} icon="message" onPress={goToEmailSupport} />
+        ) : null}
```
divyaBacancy, line 124: Just one thing, if we are giving email support, if the user has enough PLR balance or not so we can put default condition instead of separate conditions for both enoughPlrBalance and !enoughPlrBalance
<MenuItem title={t('item.liveChatSupport')} icon="message" onPress={goToEmailSupport} />

divyaBacancy (approved):

PR #2998, opened 2022-02-08
Improvement/Keychain cancelled issue

vignesha22:
Added Regex as discussed for neglecting the error caused by user when the biometric is cancelled

### DIFF
--- a/src/utils/keychain.js
+++ b/src/utils/keychain.js
@@ -49,37 +49,40 @@ export type KeyChainData = {
 };
 
 export const handleCatch = (accountAddress: ?string, error: ?any) => {
-  reportErrorLog('Exception caught on keychain: ', error);
-  const colors = getThemeColors();
-  const buttons = [];
-  buttons.push({
-    text: t('error.failedKeychain.exitButtonText'),
-    onPress: () => RNExitApp.exitApp(),
-  });
-  accountAddress = accountAddress ?? etherspotService?.getAccountAddress(CHAIN.ETHEREUM);
-  if (accountAddress) {
+  const msg = error[0]?.Error?.msg;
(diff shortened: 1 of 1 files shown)


src/utils/keychain.js
```diff
@@ -48,38 +48,41 @@ export type KeyChainData = {
...
-  const colors = getThemeColors();
-  const buttons = [];
-  buttons.push({
-    text: t('error.failedKeychain.exitButtonText'),
-    onPress: () => RNExitApp.exitApp(),
-  });
-  accountAddress = accountAddress ?? etherspotService?.getAccountAddress(CHAIN.ETHEREUM);
-  if (accountAddress) {
+export const handleCatch = (accountAddress: ?string, error: ?any[]) => {
+  const msg = error ? error[0]?.Error?.msg : '';
```
IAmKio, line 52: Could you just `.toString()` `error` here?

src/utils/keychain.js
```diff
@@ -48,38 +48,41 @@ export type KeyChainData = {
...
+  if (msg && !/cancel/gi.exec(msg)) {
+    reportErrorLog('Exception caught on keychain: ', error);
```
IAmKio, line 54: Please move the error reporting outside the `if` statement, we want to report all errors even if they are just actions that are cancelled.

IAmKio (changes requested):
Just some minor changes

src/utils/keychain.js
vignesha22: okay

src/utils/keychain.js
vignesha22: that's what I thought but it won't return as json to separate code and then have to parseJson and then take msg

IAmKio (approved):
LGTM

PR #3001, opened 2022-02-09
 PIL-1256 Revamp fiat on ramp - Add Cash

divyaBacancy:
https://user-images.githubusercontent.com/66996635/153208644-90423e60-92be-4e74-84e4-25051b87aaba.mov


Yash-Chaubal (commented):
Looking good, great work! Could you just make sure new files are .tsx rather than .js please?

src/screens/AddCash/AddCash.js
```diff
@@ -62,48 +64,65 @@ const AddCash = () => {
...
+      setCurrencySymbol('$');
+    }
+  };
+
+  const networkSelected = () => {
+    if (inUS) {
+      openWert();
+    } else {
+      openRamp();
+    }
```
Yash-Chaubal, line 83: Could some of these if statements also be single line, like so:
```suggestion
    if (inUS) openWert();
    else openRamp();
```

Yash-Chaubal, line 83: Could some of these if statements also be single line, like so:
```suggestion
    if (inUS) openWert();
    else openRamp();
```

divyaBacancy:
@Yash-Chaubal Thanks for reviewing. Done with the changes.

Yash-Chaubal (approved):
LGTM!

PR #3011, opened 2022-02-18
Added Mumbai testnet in supported network of etherspot

divyaBacancy:
To reflect the balance of Mumbai testnet in our wallet


Yash-Chaubal (approved):
LGTM

PR #3021, opened 2022-03-01
PIL-1287 Prune Sentry Logging

divyaBacancy:


IAmKio (approved):
LGTM - let's see how this appears on Sentry going forward, it may require some minor changes in future.

PR #3022, opened 2022-03-01
 PIL-1320 Remove swap button from key wallet account -> token screen

divyaBacancy:
https://user-images.githubusercontent.com/66996635/156124923-f800f05b-273a-403f-a3bb-c1650eb623fd.mov


IAmKio (approved):
LGTM

PR #3050, opened 2022-05-23
Fix/NullPointerException

divyaBacancy:
remove blank placeholder in multiline textinput.


src/components/inputs/MultilineTextInput.js
```diff
@@ -48,7 +48,7 @@ const MutlilineTextInput = React.forwardRef<Props, Instance>(({ style, ...rest }
       ref={ref}
       keyboardAppearance={keyboardAppearance}
       // `placeholder` is needed for Android to avoid slight increase in height when input is empty.
-      placeholder=""
+      // placeholder=""
```
IAmKio, line 51: Could we remove this instead of commenting it out?

divyaBacancy: yes, done

IAmKio (approved):
LGTM

PR #3054, opened 2022-05-26
PRO-200 insufficient MATIC for gas fee

divyaBacancy:
`isEnoughBalanceForTransactionFee` in set all chain wallets.


IAmKio (approved):
LGTM - great find! 🚀

PR #3056, opened 2022-05-31
PRO-29 ERC-1155 Support

divyaBacancy:
Implemented by erc1155 abi and safeTransferFrom function for erc1155 token transactions.

https://polygonscan.com/tx/0xf1db8da9c90aa2e69445ebca4899de9b8be9ebb0c82fdf4b59166ce12daa143d


src/utils/transactions.js
```diff
@@ -108,15 +111,24 @@ export const buildEthereumTransaction = async (
...
+        from,
+        to,
+        tokenId,
+        contractAddress,
+        useLegacyTransferMethod: !!useLegacyTransferMethod,
+      });
+      to = contractAddress;
+      value = EthersBigNumber.from(0);
+    } catch (e) {
+      const approveData = encodeContractMethod(ERC1155_CONTRACT_ABI, 'setApprovalForAll', [contractAddress, true]);
```
IAmKio, line 127: I'm unsure about using the `catch` block here to perform a different statement. Could we check what kind of token this is beforehand? The `catch` block should only be used for error handling...

IAmKio (changes requested):

src/utils/transactions.js
divyaBacancy: Yes, Actually we can't identify which one is ERC721 or ERC1155 token. So first of all we `try` to check token is ERC721 or not. if not ERC721 then call `catch`.

divyaBacancy: Please review updated code.

IAmKio (approved):

PR #3067, opened 2022-06-28
PRO-188 BSC wallet shows as deployed in the transaction history

divyaBacancy:
Removed deployed transaction in BSC

image


divyaBacancy:
@IAmKio  updated condition is if toAddress is equal to wallet address then showed incoming otherwise outgoing. before this condition was fromAddress is equal to wallet address then showed outgoing otherwise incoming. this is okay?

IAmKio (approved):

PR #3072, opened 2022-07-19
PRO-364 Recovery phrase not fully visible

hemadri-tank-bacancy:
Phrase not fully visible issue resolved.
Simulator Screen Shot - iPhone 13 Pro Max - 2022-07-19 at 19 35 00


IAmKio (approved):
LGTM!

PR #3097, opened 2022-08-09
PRO-442 Remove Copy button from 12 words screen

hemadri-tank-bacancy:
Remove copy button in two screens.

Simulator Screen Shot - iPhone 13 Pro Max - 2022-08-09 at 20 06 51
Simulator Screen Shot - iPhone 13 Pro Max - 2022-08-09 at 20 04 46


IAmKio (approved):
LGTM 🙏

PR #3098, opened 2022-08-10
PRO-443 Avalanche Feature Switch

hemadri-tank-bacancy:
Set avalanche feature switch.

https://user-images.githubusercontent.com/78720269/183850759-952b7f92-6f6c-450d-9713-0e47b3527e64.mp4


IAmKio:
@hemadri-tank-bacancy can I just confirmed that this will also hide Avalanche tokens during asset selection?

hemadri-tank-bacancy:
> @hemadri-tank-bacancy can I just confirmed that this will also hide Avalanche tokens during asset selection?

Yes, as well avalanche is not visible in asset selection.

IAmKio (approved):

PR #3103, opened 2022-08-22
Fix - Tap areas increase for Settings and History

hemadri-tank-bacancy:
Increase tap area size.

Simulator Screen Shot - iPhone 13 Pro Max - 2022-08-22 at 16 10 02


IAmKio (approved):
LGTM

PR #3107, opened 2022-08-26
Fix - Optimism Issues

hemadri-tank-bacancy:
1. App crashes when attempting to make a cross-chain swap (POLYGON -> OP) - Issue resolved.
2. Fiat value for Optimism is not showing - Issue resolved.

https://user-images.githubusercontent.com/78720269/186852444-d87da2ce-005d-4153-ba05-c0695ad3bda6.mp4


IAmKio (approved):
LGTM! FYI - the rates service from Etherspot will soon replace Coingecko in the app.

PR #3115, opened 2022-09-01
Fix- ETH value in optimism

hemadri-tank-bacancy:
* Fetched ether value on optimism

simulator_screenshot_E7BD1AC2-F2B9-4DF5-AD66-B5B0C25B9A06


IAmKio (approved):
LGTM

PR #3116, opened 2022-09-05
PRO-529 Change Swap Token Amount Calculation

hemadri-tank-bacancy:
* Update calculation conditions in cross-chain swap token amount 
* Update fiat value calculation condition in cross-chain offers list.

https://user-images.githubusercontent.com/78720269/188412899-6889bb01-b129-4464-9a2d-17c581b57467.mp4


IAmKio (approved):
LGTM 🎉

PR #3117, opened 2022-09-05
PRO-527 Fetch rates when pair selected in exchange

hemadri-tank-bacancy:
* Set Fetching rates… in cross-chain + swap
* Set Could not fetch rates in cross-chain + swap

Uploading Simulator Screen Recording - iPhone 13 Pro Max - 2022-09-05 at 18.44.00.mp4…


IAmKio (approved):
LGTM!

PR #3125, opened 2022-09-10
PRO-546 Fetched rates from Swaps

hemadri-tank-bacancy:
* Fix - fetched rates from swaps

Description: 
In swap and cross-chain now all tokens value visible which available in coingecko api.

https://user-images.githubusercontent.com/78720269/189483331-4f001f67-8b7e-46cb-9ad0-aac91f79e0fc.mp4


IAmKio (approved):
LGTM

PR #3126, opened 2022-09-11
PRO - 506 Keyboard pop up in "Get notifications" screen

hemadri-tank-bacancy:
* Fix - Hide keyboard in "Get notification" screen

### DIFF
--- a/src/screens/Notification/GetNotifications.tsx
+++ b/src/screens/Notification/GetNotifications.tsx
@@ -18,8 +18,8 @@
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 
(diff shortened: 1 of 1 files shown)


IAmKio:
@hemadri-tank-bacancy the solution is a fix but why does it come up with the keyboard in the first place?

hemadri-tank-bacancy:
> @hemadri-tank-bacancy the solution is a fix but why does it come up with the keyboard in the first place?

Actually I can't replicate this issue. sometimes In one scenario might be possible. like this screen open after to Add Cash screen. and here keyboard was not hide before navigate to Get notifications screen.

IAmKio (approved):
LGTM

PR #3139, opened 2022-09-29
PRO-627 Title Display Change (ENS user)

hemadri-tank-bacancy:
* Change ENS user title
  - In this not show ENS address when we're switched to the Key Wallet account
  - For Key Wallet account in show "Key Wallet"


IAmKio (approved):
LGTM

PR #3165, opened 2022-11-09
Fix - android build fail

hemadri-tank-bacancy:
* Set condition for Android 12

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -38,7 +38,7 @@
                 android:name=".MainActivity"
                 android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
                 android:label="@string/app_name"
-                android:launchMode="singleTask"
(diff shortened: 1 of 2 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -34,14 +34,15 @@
         android:theme="@style/AppTheme">
 
         <activity
+                android:exported="true"
                 android:name=".MainActivity"
                 android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
                 android:label="@string/app_name"
-                android:launchMode="singleInstance"
+                android:launchMode="singleTask"
```
IAmKio, line 41: I just need to check this, if i recall correctly, we had a security issue around this...

hemadri-tank-bacancy: I replaced it.

IAmKio (approved):
LGTM

PR #3172, opened 2022-11-24
fix/PRO-799-asset-send-optimisations

IAmKio:
- Updated Etherspot SDK to 1.41.1
- Set gateway submit batch guard to false
- Used feature flag for NFTs to prevent calls to redundant NFT data endpoints
- Removed await for dispatched actions
- Altered tests to reflect changes

### DIFF
--- a/src/services/etherspot.js
+++ b/src/services/etherspot.js
@@ -546,7 +546,7 @@ export class EtherspotService {
     }
 
     // submit current batch
-    const { hash: batchHash } = await sdk.submitGatewayBatch();
+    const { hash: batchHash } = await sdk.submitGatewayBatch({ guarded: false });
 
     return { batchHash };
   }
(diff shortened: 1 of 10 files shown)


src/services/etherspot.js
```diff
@@ -596,7 +596,7 @@ export class EtherspotService {
     }
 
     // submit current batch
-    const { hash: batchHash } = await sdk.submitGatewayBatch();
+    const { hash: batchHash } = await sdk.submitGatewayBatch({ guarded: false });
```
hemadri-tank-bacancy, line 599: Just for my information. what is the use of **guarded**

hemadri-tank-bacancy:
Everything looks good. Also, I have tested optimism 
Sent you 0.1 OP
https://optimistic.etherscan.io/tx/0x8dff24056359dfc37e1ab2538eb5b271053a8c4be97286376bc103c85b0e7340

hemadri-tank-bacancy (approved):

src/services/etherspot.js
IAmKio: It prevents users absuing our services and ensures that transactions are being sent through our relayers.

PR #3179, opened 2022-12-08
Fixed Show null price

hemadri-tank-bacancy:
* Small issue fixed for price show null when rates failed.

Screenshot_20221208-125301_Pillar Wallet


IAmKio (approved):
LGTM

PR #3181, opened 2022-12-12
PRO-865 Investigate / Remove Invalid Hooks Call

hemadri-tank-bacancy:
Hooks functions used in those files removed
* collectiblesActions.js
* keyBasedAssetTransferActions.js


IAmKio (approved):
LGTM

PR #3182, opened 2022-12-13
PRO-476 Apps Pane

hemadri-tank-bacancy:
Points
  1. Dashboard in shows top 5 Apps Holdings
  2. Apps Holdings is sorted by USD.
  3. Show total Apps Holdings in Assets

https://user-images.githubusercontent.com/78720269/207262159-05e47bb8-5c7c-4fc7-a595-205dc089903c.mp4


src/actions/appsHoldingsActions.tsx
```diff
@@ -0,0 +1,95 @@
...
+});
+
+export const updateAppsHoldingsAction = (appsHoldingsData) => {
+  return async (dispatch: Dispatch, getState: GetState) => {
+    if (isEmpty(appsHoldingsData)) return;
+
+    dispatch({ type: UPDATE_APP_HOLDINGS, payload: appsHoldingsData });
+
+    const updatedAppsHoldings = getState().appsHoldings.data;
+    await dispatch(saveDbAction('appsHoldings', updatedAppsHoldings, true));
```
IAmKio, line 51: Do we need to `await` here?

src/actions/appsHoldingsActions.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (isFetching || !isOnline) return;
+
+    dispatch(setIsFetchingHoldingsAction(true));
+
+    const getChainInvestments = chains.map((chain) =>
+      etherspotService.getAccountInvestments(chain, etherspotAccount.id),
+    );
+
+    try {
+      const accountsInvestments = await Promise.all(getChainInvestments);
```
IAmKio, line 78: Should we not use the built-in `.catch` method here instead of wrapping in a try /  catch, or is there another reason for the try / catch?

src/actions/appsHoldingsActions.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+      if (!accountsInvestments?.[0]) return;
+
+      const totalAppHoldings = [];
+      accountsInvestments.forEach((investments) => {
+        const items = investments.items;
+        totalAppHoldings.push(...items);
+      });
+
+      await dispatch(updateAppsHoldingsAction(totalAppHoldings));
```
IAmKio, line 88:
```suggestion
      dispatch(updateAppsHoldingsAction(totalAppHoldings));
```

IAmKio (changes requested):
Just a few small points but other than that - this LGTM!

src/actions/appsHoldingsActions.tsx
hemadri-tank-bacancy: Removed `await`

src/actions/appsHoldingsActions.tsx
hemadri-tank-bacancy: Removed `try / catch`. already added in getAccountInvestments SDK method

src/actions/appsHoldingsActions.tsx
hemadri-tank-bacancy: Removed `await`

IAmKio (approved):

PR #3183, opened 2022-12-13
PRO-873 Change Asset / Stable progress bar calculation

hemadri-tank-bacancy:
*Calculate progress bar using USD value

https://user-images.githubusercontent.com/78720269/207317937-77801558-246d-4539-80fb-86c8ded7b0ac.mp4


IAmKio (approved):
LGTM

PR #3197, opened 2022-12-29
PRO-950 Home Screen Scanner Camera Usage

hemadri-tank-bacancy:
Worked on these Points
  1. When the screen is not visible, the camera is not activated
  2. The permission is only requested when the camera is needed
  3. When the permission is denied, an alert is shown to the user stating that "The camera permission is needed to open the WalletConnect QR code scanner"
  4. The permissions dialog should be re-requested when the user next activates the WalletConnect scanner
  
  

https://user-images.githubusercontent.com/78720269/209936546-04561cd4-c292-4e67-a16c-8ecfd58afc36.mp4


IAmKio (approved):
Great refactor of the QR scanner 🎉 LGTM

PR #3212, opened 2023-01-19
PRO - 999 Add new EtherspotPopularTokens token list

hemadri-tank-bacancy:
* Points
  1. Implement Etherspot popular tokenlist
  2. Write down test cases

https://user-images.githubusercontent.com/78720269/213439983-36cd5da0-f108-4703-a4e8-16ea57a7a4b1.mp4

### DIFF
--- a/src/components/Modals/AssetSelectorModal/DefaultAssetList.js
+++ b/src/components/Modals/AssetSelectorModal/DefaultAssetList.js
@@ -78,9 +78,7 @@ const DefaultAssetList = ({ tokens, collectibles, onSelectToken, onSelectCollect
           name={token.name}
           iconUrl={token.iconUrl}
           balance={wrapBigNumberOrNil(token.balance?.balance)}
-          onPress={async () => {
-            await onSelectToken(token);
-          }}
+          onPress={() => onSelectToken(token)}
         />
       );
     }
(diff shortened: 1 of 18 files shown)


src/components/Modals/AssetSelectorModal/DefaultAssetList.js
```diff
@@ -117,6 +115,8 @@ const DefaultAssetList = ({ tokens, collectibles, onSelectToken, onSelectCollect
       keyboardShouldPersistTaps="always"
       ListEmptyComponent={renderEmptyState()}
       contentInsetAdjustmentBehavior="scrollableAxes"
+      removeClippedSubviews
+      scrollEventThrottle={1}
```
IAmKio, line 119: Interesting

hemadri-tank-bacancy: Actually, this is used for large lists smoothly scroll.

src/reducers/assetsReducer.js
```diff
@@ -37,12 +43,10 @@ export type AssetsReducerAction = {
...
+  popularAssets: { ethereum: [] },
 };
 
-export default function assetsReducer(
-  state: AssetsReducerState = initialState,
-  action: AssetsReducerAction,
-) {
+export default function assetsReducer(state: AssetsReducerState = initialState, action: AssetsReducerAction) {
   switch (action.type) {
     case SET_SUPPORTED_ASSETS:
```
Yash-Chaubal, line 51: Just a thought, should we have a check right at the start to check if `action` is null or not, otherwise cases like `action.type` and `action.payload` could crash the app.

You could also change it to `action?.type` but I figure checking right at the start is the easiest way

hemadri-tank-bacancy: I think not changes need in this. I check every scenario app was not crash

IAmKio (approved):

Yash-Chaubal (approved):
LGTM!

PR #3216, opened 2023-01-24
PRO - 1007 "Fetching rates" issue during swaps

hemadri-tank-bacancy:
* Points
   1. Fixed Fetching rates issue
   2. Update some conditions in fetching rates

https://user-images.githubusercontent.com/78720269/214473165-001be788-162f-4d8c-869b-2f6da76f8e45.mp4

### DIFF
--- a/src/actions/assetsActions.js
+++ b/src/actions/assetsActions.js
@@ -616,7 +616,15 @@ export const fetchAssetsBalancesAction = () => {
     dispatch({ type: SET_FETCHING_ASSETS_BALANCES, payload: true });
 
     dispatch(fetchSupportedAssetsAction());
-    if (!popularAssets?.[CHAIN.POLYGON]) dispatch(fetchPopularAssetsAction());
+    if (
+      !popularAssets?.[CHAIN.POLYGON] ||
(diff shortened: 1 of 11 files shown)


src/actions/assetsActions.js
```diff
@@ -615,7 +616,7 @@ export const fetchAssetsBalancesAction = () => {
     dispatch({ type: SET_FETCHING_ASSETS_BALANCES, payload: true });
 
     dispatch(fetchSupportedAssetsAction());
-    dispatch(fetchPopularAssetsAction());
+    if (!popularAssets?.[CHAIN.POLYGON]) dispatch(fetchPopularAssetsAction());
```
IAmKio, line 619: Was there an issue with Polygon popular assets?

IAmKio (changes requested):
Just a question (also asked on Slack) but other than this - we're looking good

src/actions/assetsActions.js
hemadri-tank-bacancy: No no there is not an issue. I put a condition that if popular assets are available then don’t call this method again. we have not any other way to find assets available or accept this one.

IAmKio (approved):
LGTM

PR #3222, opened 2023-02-01
PRO-1021 Keyboard unresponsive

hemadri-tank-bacancy:
* Increase search box input height

Screenshot_20230201-202433_Pillar Wallet

### DIFF
--- a/src/components/Modals/ContactSelectorModal/ContactSelectorModalContent.js
+++ b/src/components/Modals/ContactSelectorModal/ContactSelectorModalContent.js
@@ -181,6 +181,7 @@ const ContactSelectorModalContent = ({ chain, contacts = [], onSelectContact, qu
         onQueryChange={onQueryChange}
         placeholder={tRoot('label.addressSupportedNames')}
         error={!!errorMessage}
+        inputStyle={{ height: 78 }}
         autoFocus
       />
 


src/components/Modals/ContactSelectorModal/ContactSelectorModalContent.js
```diff
@@ -181,6 +181,7 @@ const ContactSelectorModalContent = ({ chain, contacts = [], onSelectContact, qu
         onQueryChange={onQueryChange}
         placeholder={tRoot('label.addressSupportedNames')}
         error={!!errorMessage}
+        inputStyle={{ height: 78 }}
```
IAmKio, line 184: This might be okay on your screen but on smaller screens this may be an issue. Could we dynamically determine the height of the box and use that value?

IAmKio (changes requested):

src/components/Modals/ContactSelectorModal/ContactSelectorModalContent.js
hemadri-tank-bacancy: I checked the smallest device which is not facing any issues.
image

IAmKio (approved):
LGTM

PR #3224, opened 2023-02-02
PRO -1021 Re - Keyboard unresponsive

hemadri-tank-bacancy:
Points:
 1. Slide modal replaced to Modal
 2. Increase address search box height


IAmKio (approved):
LGTM

PR #3226, opened 2023-02-03
PRO-1040 WalletConnect Chain Check

hemadri-tank-bacancy:
Conditions for **eth_signedTypedData** and **eth_signedTypedData** call request.

1. If chainId not found then refused request and show error message.
2. If dApp chainId not match with call request then reject request and show error message.
3. If dApp chainId match with call request then show request modal.

### DIFF
--- a/src/actions/walletConnectActions.js
+++ b/src/actions/walletConnectActions.js
@@ -20,6 +20,7 @@
 import { NavigationActions } from 'react-navigation';
 import t from 'translations/translate';
 import { BigNumber } from 'bignumber.js';
+import { isEmpty } from 'lodash';
 
 // constants
 import {
@@ -68,7 +69,6 @@ import { isLogV2AppEvents } from 'utils/environment';
 // models, types
 import type { WalletConnectCallRequest, WalletConnectConnector, sessionDataProps } from 'models/WalletConnect';
 import type { Dispatch, GetState } from 'reducers/rootReducer';
-import { isEmpty } from 'lodash';
 
 const setWalletConnectErrorAction = (message: string) => {
   return (dispatch: Dispatch) => {
--- a/src/actions/walletConnectActions.js
+++ b/src/actions/walletConnectActions.js
@@ -68,6 +68,7 @@ import { isLogV2AppEvents } from 'utils/environment';
 // models, types
 import type { WalletConnectCallRequest, WalletConnectConnector, sessionDataProps } from 'models/WalletConnect';
 import type { Dispatch, GetState } from 'reducers/rootReducer';
+import { isEmpty } from 'lodash';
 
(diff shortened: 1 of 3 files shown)


src/locales/en/common.json
```diff
@@ -584,7 +584,8 @@
       "alreadyPending": "A connection is already waiting",
       "callRequestRejectFailed": "Call Request reject failed",
       "callRequestApproveFailed": "Call Request approve failed",
-      "invalidRequest": "This WalletConnect request didn't look quite right - please try again",
+      "invalidRequest": "This WalletConnect request didn't look quite right - please try again ",
+      "cannotDetermineChain": "Stating that no Chain was found for the incoming request, and to contact {{ dAppName }} ",
```
IAmKio, line 588:
```suggestion
      "cannotDetermineChain": "Sorry, but no chain information was found in this request. Please contact {{ dAppName }}",
```

IAmKio (changes requested):
Just one small change, but other than that it looks good so far

src/locales/en/common.json
hemadri-tank-bacancy: Changed it

src/actions/walletConnectActions.js
```diff
@@ -375,6 +377,23 @@ export const subscribeToWalletConnectConnectorEventsAction = (connector: WalletC
         return;
       }
 
+      if (method === ETH_SIGN_TYPED_DATA || method === ETH_SIGN_TYPED_DATA_V4) {
+        const { domain } = JSON.parse(params[1]);
```
IAmKio, line 381: `JSON.parse` throws an error if there is an error in the JSON or something else like `params[index]` does not exist. Please ensure thta params[index] exists and / or wrap this in a try / catch and handle the error.

IAmKio (changes requested):
Apologies just another thing i noticed here that could cause a crash

src/actions/walletConnectActions.js
hemadri-tank-bacancy: Well, I made one more condition

src/actions/walletConnectActions.js
```diff
@@ -378,6 +379,10 @@ export const subscribeToWalletConnectConnectorEventsAction = (connector: WalletC
       }
 
       if (method === ETH_SIGN_TYPED_DATA || method === ETH_SIGN_TYPED_DATA_V4) {
+        if (isEmpty(params)) {
+          dispatch(setWalletConnectErrorAction(t('error.walletConnect.cannotDetermineChain', { dAppName: name })));
```
IAmKio, line 383: Could we please add a sentry / error log here so we can track when this has happened. Thanks

IAmKio (approved):
LGTM

PR #3229, opened 2023-02-09
PRO-1064 Reinstate SDK Batch Guard

hemadri-tank-bacancy:
* Removed guarded: false from `sdk.submitGatewayBatch` in all instances.


IAmKio (approved):
LGTM

PR #3232, opened 2023-02-09
Feature/OIDC pipeline

vnovacev:
CircleCI oidc tokens should be used in wallet pipeline instead of AWS machine user access keys - security recommendation.


arddluma (approved):
LGTM 💯

PR #3233, opened 2023-02-10
PRO-1054 Exclude offers that dont have USD value

hemadri-tank-bacancy:
Points
 * Exclude all offers that dont have a USD value for exchange and cross-chain. 
 * Sort offers by USD value minus gas fees

https://user-images.githubusercontent.com/78720269/218063896-9770b336-fcb4-4618-92e8-7e1a840e4f23.mp4


IAmKio (approved):
LGTM

PR #3238, opened 2023-02-21
Fix - Promise.allSettled is not a function.

hemadri-tank-bacancy:
* Fixed promise allSettled is not a function error


IAmKio (approved):
LGTM

PR #3241, opened 2023-02-22
PRO-1110 Undefined string when loading investments data

hemadri-tank-bacancy:
* Updated key and some conditions in app investments

https://user-images.githubusercontent.com/78720269/220532623-8d3c96e5-7a24-41e3-ba2e-27cf9ee427e9.mp4

### DIFF
--- a/src/components/lists/InvestmentListItem.tsx
+++ b/src/components/lists/InvestmentListItem.tsx
@@ -118,6 +118,11 @@ export default function ({
     );
   };
 
+  const getItemKey = (item) => {
+    const { name, balance } = item;
+    return name + balance;
+  };
+
   return (
     <MainContainer>
       <Container onPress={onPress} style={style} hitSlop={{ top: spacing.medium, bottom: spacing.medium }}>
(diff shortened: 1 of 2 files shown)


src/components/lists/InvestmentListItem.tsx
```diff
@@ -118,6 +118,11 @@ export default function ({
     );
   };
 
+  const getItemKey = (item) => {
+    const { name, balance } = item;
+    return name + balance;
```
IAmKio, line 123: Can we guarantee that `name` and `balance` are available?

Regarding `name + balance`, if either of them are undefined we'll get a `NaN` or `undefined` in the key which may lead to duplicate keys. Is there a different way to do this using a more unique identifier?

IAmKio (changes requested):
Just a small point but other than that - looks good!

src/components/lists/InvestmentListItem.tsx
hemadri-tank-bacancy: If the item is not available then keyExtractor not call

IAmKio (approved):

PR #3252, opened 2023-02-27
Chore/new packages url

vnovacev:
At the moment Slack messages for available apk/ipa package point to presigned s3 links that have expiration time and you can't download the package from these locations after expiration of assigned tokens.
This PR changes the way of Slack message's creation and now new available apk/ipa packages can be downloaded from GitHub releases location where assets are stored permanently.


arddluma (approved):
LGTM 💯

PR #3257, opened 2023-03-09
PRO-1148 Local gas tokens

hemadri-tank-bacancy:
Local gas tokens
https://linear.app/pillarproject/issue/PRO-1148/local-gas-tokens

## Description

- Managed locally native tokens icons.

## Motivation and Context

- Some time third party assets not working

## How Has This Been Tested?

- Locally on iPhone 13 simulator (Mainnet)
- Locally on Samsung A50 device (Testnets)
-

https://user-images.githubusercontent.com/78720269/223951765-7d39d9e5-1dc0-4f43-9235-a6e6e214055b.mp4

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/src/utils/collectibles.js
+++ b/src/utils/collectibles.js
@@ -24,6 +24,7 @@ import { orderBy } from 'lodash';
 import { addressesEqual } from 'utils/assets';
 import { addressAsKey } from 'utils/common';
 import { caseInsensitiveIncludes } from 'utils/strings';
+import { nativeAssetPerChain } from 'utils/chains';
 
 // Types
 import type { Collectible } from 'models/Collectible';
@@ -57,4 +58,13 @@ export function findCollectible(
   return collectibles?.find(
     ({ contractAddress, id }) => addressesEqual(contractAddress, contractAddressToFind) && id === idToFind,
   );
+}
+
+export function isChainIcon(nameOrUrl: ?string): boolean {
+  if (!nameOrUrl) return false;
+
+  const chainInfo = nativeAssetPerChain[nameOrUrl];
+  if (!chainInfo) return false;
+
+  return true;
 }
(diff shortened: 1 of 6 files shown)


src/utils/collectibles.js
```diff
@@ -58,3 +59,12 @@ export function findCollectible(
     ({ contractAddress, id }) => addressesEqual(contractAddress, contractAddressToFind) && id === idToFind,
   );
 }
+
+export function isChainIcon(nameOrUrl: ?string): boolean {
+  if (!nameOrUrl) return false;
+
+  const chainInfo = nativeAssetPerChain[nameOrUrl];
+  if (!chainInfo) return false;
+
```
IAmKio, line 68: - Could we please move this out of this file and into a utils file for tokens? This file is specific to NFTs
- Could we add an error log here to state what name or URL was not found so we can keep track of URL problems for tokens

IAmKio (changes requested):
Just a small change but other than that - it's looking good!

src/utils/tokens/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+// Utils
+import { nativeAssetPerChain } from 'utils/chains';
+import { reportErrorLog } from 'utils/common';
+
+export function isChainIcon(nameOrUrl: string | any): boolean {
```
Yash-Chaubal, line 25: Shouldn't `nameOrUrl` always be a string? Making a var a union type with `any` is kinda redundant, as then it might as well be just `any` type (which is better to avoid)

Also could you change this function to arrow notation `() => { ... }`

hemadri-tank-bacancy: Updated conditions as per you suggestion.

Yash-Chaubal: Sorry I meant that it should be of type `string` as `any` shouldn't really be used

hemadri-tank-bacancy: Updated as string type

Yash-Chaubal (approved):
Thanks Hemadri, LGTM!

IAmKio (approved):
LGTM

PR #3260, opened 2023-03-15
PRO-1168: Added testID prop for elements in Onboarding flow

Yash-Chaubal:
I've added `testID`s for various interactable elements during the Onboarding flow + Home screen -> Re-import Wallet flow.

The IDs are in this format: `[TAG]-[TYPE]-[NAME_WITH_UNDERSCORES].[OPTIONAL_PROP].[ETC]`

Where:
- TAG is the name of the file
- TYPE is the type of the element (currently only 5 types: `button`, `checkbox`, `input`, `text`, `error`)
- And after that there will be a name in snake_case. If the testIDs are generated through an array then there may be extra keys joined by `.`

You may need to look at the code to see what exactly the IDs are, but here are some example IDs:
- Normal button `BiometricModal-button-enable`
- Normal checkbox `ImportWalletLegals-checkbox-agree_to_policy`
- Checkboxes generated through an array/object `Permissions-button-permission_dropdown.PUSH_NOTIFICATIONS`

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,24 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# ensure character set
+[*.{js,ts,tsx}]
+charset = utf-8
+
+# 2 space indentation using spaces, single quotes
+[*.{js,ts,tsx}]
+indent_style = space
+indent_size = 2
+quote_type = single
+max_line_length = 120
+
+[{package.json}]
+indent_style = space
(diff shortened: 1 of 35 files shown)


hemadri-tank-bacancy (approved):
LGTM!

.editorconfig
```diff
@@ -0,0 +1,24 @@
...
+# 2 space indentation using spaces, single quotes
+[*.{js,ts,tsx}]
+indent_style = space
+indent_size = 2
+quote_type = single
+max_line_length = 120
+
+[{package.json}]
+indent_style = space
+indent_size = 2
```
IAmKio, line 24: ayooo 🚀

IAmKio (approved):
Skimmer over this - LGTM

PR #3269, opened 2023-03-24
PRO-349: Fixed crash on Receive Modal

Yash-Chaubal:
Prop `theme.colors` was causing a crash, so switched to just `colors`.


IAmKio (approved):
LGTM

PR #3271, opened 2023-03-28
PRO - 1002 Fix - Default Token list for new users

hemadri-tank-bacancy:
https://linear.app/pillarproject/issue/PRO-1002/pillar-wallet-default-token-list-for-new-users

## Description

- Added new default tokens (MAI and USDT In stable list)
- Fixed native tokens under stablecoins
- Resolved issue in default token list show balances
- Revert some conditions
- BOB (polygon) moved in Stable list

## Motivation and Context

- https://linear.app/pillarproject/issue/PRO-1002#comment-0ea8c584
- https://linear.app/pillarproject/issue/PRO-1002#comment-6e4d5e73
- https://pillarwallet.slack.com/archives/G016NP99QJV/p1679760290136299

## How Has This Been Tested?

- Tested on Samsung Galaxy A50 (As existing user + new user)
- Tested on Simulator iPhone 13 (As existing user + new user)

Uploading Screen_Recording_20230328-161916_Pillar Wallet.mp4…

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)


src/utils/assets.js
```diff
@@ -422,7 +422,8 @@ export const sortInvestmentPositions = (positionsInfo: AccountInvestmentPosition
   return newArr;
 };
 
-export const isSameAsset = (a: Asset, b: Asset) => a.address?.toLowerCase() === b.address?.toLowerCase();
+export const isSameAsset = (a: Asset, b: Asset) =>
+  a.symbol === b.symbol && a.address.toLowerCase() === b.address.toLowerCase();
```
Yash-Chaubal, line 426: We should just be checking the address when comparing two assets - the symbol _could_ potentially be different depending where the asset is pulled from. The address is always unique

Also any reason for removing the null check `?`?
```suggestion
 a?.address?.toLowerCase() === b?.address?.toLowerCase();
```

src/utils/etherspot.js
```diff
@@ -312,4 +312,5 @@ export const filteredWithChain = (assets: Asset[], chain: Chain): Asset[] => {
   return chain ? assets.filter((asset) => asset.chain === chain) : assets;
 };
 
-const isSameAsset = (a: AssetOption, b: AssetOption) => a.address?.toLowerCase() === b.address?.toLowerCase();
+const isSameAsset = (a: AssetOption, b: AssetOption) =>
+  a.symbol === b.symbol && a.address.toLowerCase() === b.address.toLowerCase();
```
Yash-Chaubal, line 316: Same as above

Yash-Chaubal (changes requested):
Looking good, just a couple of comments!

src/actions/assetsActions.js
```diff
@@ -710,17 +716,22 @@ export const fetchSupportedAssetsAction = () => {
...
+          customTokensList,
+          chain,
         );
 
-        const totalSupportedAssets = [...chainSupportedAssets, ...removedDuplicateSupportedAssets];
+        let totalSupportedAssets = [...chainSupportedAssets, ...removedDuplicateSupportedAssets];
+
+        const removedDuplicateDefaultAssets = filteredWithDefaultAssets(totalSupportedAssets, defaultTokens, chain);
+
+        totalSupportedAssets = [...totalSupportedAssets, ...removedDuplicateDefaultAssets];
```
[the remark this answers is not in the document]
hemadri-tank-bacancy: some name updated in this `filteredDefaultAssets, filteredCustomAssets`

src/actions/notificationsActions.js
```diff
@@ -145,7 +145,7 @@ export const subscribeToPushNotificationsAction = () => {
       if (disabledFetchingBalanceListener !== null) return;
       disabledFetchingBalanceListener = setInterval(() => {
         dispatch(fetchAssetsBalanceNotificationsAction());
-      }, 10000);
+      }, 20000);
```
IAmKio, line 148: `fetchAssetsBalanceNotificationsAction` - what does this actually do? Just curious to know why we're firing it every 20s...

IAmKio (changes requested):
Just a quick question - but other than that, looks good!

src/actions/notificationsActions.js
hemadri-tank-bacancy: Actually, firing it to get token balance. Aldin required a quick refresh balance after a swap or send (within 10 seconds). And we have no token received notification state. So I set an interval of 10 seconds to get the updated balance. But I thought that every 10-second could affect another process.

hemadri-tank-bacancy: Before the set interval, it was fired every 30 seconds

hemadri-tank-bacancy: the interval is canceled only when the app state changes(Sing out or expire session).
Actually, we do have not any notification SDK method to find out whether received a token.

IAmKio: We might need to try and figure out a mechanism to cancel this. I would recommend that we listen for the response, and as soon as we get it we attempt to cancel the interval. That is preferred, but the simple way is to cancel after a timeout

hemadri-tank-bacancy: @IAmKio I made changes as per your suggestion. please review it.
https://github.com/pillarwallet/pillarwallet/pull/3275

IAmKio: We can resolve this issue due to the other PR

PR #3275, opened 2023-03-30
Fix - Get balances Listener

hemadri-tank-bacancy:
Get balances Listener

## Description

- Updated get balances listener 
- Set timeout to canceled listener (For exchange 1.5 minute and in send 1 minute)
- Listener calls only when Exchange or Send performed

## Motivation and Context

- https://github.com/pillarwallet/pillarwallet/pull/3271#discussion_r1152291103

## How Has This Been Tested?

- Tested on Simulator
- Tested on Samsung A50
-

https://user-images.githubusercontent.com/78720269/228754080-e01574ad-5471-4e5f-a61b-5f121f7bb373.mp4

## Types of changes


src/screens/Home/components/ExchangeNotification.tsx
```diff
@@ -57,6 +58,17 @@ export default function ({ data, index }) {
...
   const { fromAmount, toAmount, fromAsset, toAsset } = data.offer;
 
+  React.useEffect(() => {
+    if (timerStatus === STATUS.STARTED) {
+      dispatch(startListeningGetBalancesAction());
+    }
+    if (timerStatus === STATUS.STOPPED) {
+      setTimeout(() => {
+        dispatch(stopListeningGetBalancesAction());
+      }, 1500 * 60);
```
Yash-Chaubal, line 68: Just wondering, do we have any constants for minutes and seconds in milliseconds? If not I think this would be better represented as `90 * 1000` to make it clear that its a 90 second interval

Yash-Chaubal (approved):
LGTM! Just a question

IAmKio (approved):

src/screens/Home/components/ExchangeNotification.tsx
hemadri-tank-bacancy: Updated it

PR #3279, opened 2023-04-03
PRO -1244 pillar default token list

hemadri-tank-bacancy:
https://linear.app/pillarproject/issue/PRO-1244/pillar-wallet-pillar-default-token-list

## Description

- Added the Multichain token list called “Pillar Default List”
- list all the tokens from all the chains.
- All tokens are enabled but the user cannot disable them.

## Motivation and Context

- 

## How Has This Been Tested?

- Tested on Simulator
-
-

https://user-images.githubusercontent.com/78720269/229526522-6dbc7f8c-2121-4699-bec3-ef0bdbd47438.mp4

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/src/constants/assetsConstants.js
+++ b/src/constants/assetsConstants.js
@@ -170,5 +170,4 @@ export const FETCHING_DEFAULT_LIST = 'FETCHING_DEFAULT_LIST';
 export const SET_DEFAULT_LIST = 'SET_DEFAULT_LIST';
 export const SET_STABLE_DEFAULT_LIST = 'SET_STABLE_DEFAULT_LIST';
 
-export const PILLAR_DEFAULT_LIST = 'Pillar Default List';
 export const ETHERSPOT_POPULAR_MULTICHAIN = 'EtherspotPopularMultichain';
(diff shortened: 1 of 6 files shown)


src/constants/assetsConstants.js
```diff
@@ -169,3 +169,6 @@ export const ETHERSPOT_DEFAULT_STABLE_LIST = 'EtherspotDefaultStablesList';
 export const FETCHING_DEFAULT_LIST = 'FETCHING_DEFAULT_LIST';
 export const SET_DEFAULT_LIST = 'SET_DEFAULT_LIST';
 export const SET_STABLE_DEFAULT_LIST = 'SET_STABLE_DEFAULT_LIST';
+
+export const PILLAR_DEFAULT_LIST = 'Pillar Default List';
```
Yash-Chaubal, line 173: We don't use constants for titles, this should be added to the localization then pulled from `translations/translate`

Yash-Chaubal (changes requested):
Looking good, just need to change the constant to a translation variable

PR #3292, opened 2023-04-14
PRO-914: Swap capture fee to gas token

Yash-Chaubal:
https://linear.app/pillarproject/issue/PRO-914/pillar-wallet-auto-swap-fee-capture-to-gas-token

- Before sending the capture fee to the capture address, we swap it to the capture fee.
- Currently only working on single-chain swaps, I gutted the code for cross-chain (but have it saved in case we revisit this).

## How Has This Been Tested?

- Locally on an iPhone 11 simulator + Google Pixel 3 emulator
- Not entirely sure how to test swaps without making them (and paying fees), but the transaction objects seem to look about right.
- 
## Types of changes

- [x] New feature (non-breaking change which adds functionality)


IAmKio (approved):
All looks good from my side, however if you want to test for real plesae always ask the QA team for funds and they will send it over 👍

PR #3296, opened 2023-04-24
PRO-1336: Kochava Integration

Yash-Chaubal:
https://linear.app/pillarproject/issue/PRO-1336/setup-pillar-wallet-apps-in-kochava
Added Kochava integration for Android and iOS

## How Has This Been Tested?

- Locally on iPhone 14 and Google Pixel 3 simulators 

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -38,6 +38,7 @@ import { QueryClient, QueryClientProvider } from 'react-query';
 import Instabug from 'instabug-reactnative';
 import appsFlyer from 'react-native-appsflyer';
 import { PlayInstallReferrer } from 'react-native-play-install-referrer';
+import { KochavaTracker } from 'react-native-kochava-tracker';
 
 import 'services/localisation/translations';
 import localeConfig from 'configs/localeConfig';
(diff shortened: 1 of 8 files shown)


arddluma (approved):

App.tsx
```diff
@@ -255,6 +256,12 @@ class App extends React.Component<Props, any> {
       });
     }
 
+    // Kochava init
+    logBreadcrumb('App.js', `Kochava: initialize`);
+    KochavaTracker.instance.registerIosAppGuid(getEnv().KOCHAVA_IOS_ID);
+    KochavaTracker.instance.registerAndroidAppGuid(getEnv().KOCHAVA_ANDROID_ID);
```
IAmKio, line 262: Just wondering if we should be checking whether this is on iOS or Android then calling either `registerIosAppGuid` or `registerAndroidAppGuid` accordingly? I don't know the behaviour but i'm worried it would send two events for the iOS and Android project regardless of platform.

IAmKio (changes requested):
Just one small thing to note, but other than that - looks good!

App.tsx
Yash-Chaubal: Yeah I just followed the example they had in their docs (which invoked both calls regardless of platform), but i can change that

IAmKio (approved):
LGTM!

PR #3313, opened 2023-06-09
Release Candidate 3.24.3

IAmKio:
🚀 [Release Candidate 3.24.3](https://github.com/pillarwallet/pillarwallet/milestone/183)

- New! Token Analytics
- New! WalletConnect 2 Integration
- Improved: Swap screen refinements
- Fixed: PIN UI


Yash-Chaubal (approved):
LGTM! Just one thing, should we maybe not mention the pin thing directly and keep it a bit more vague like "Fixed: Pin issues" since its kinda a security risk for older versions (same for the branch name)

IAmKio:
@Yash-Chaubal good call, changed!

PR #3339, opened 2023-07-21
PRO-1609 Investigate Leak Memory of App

lalitkumarrock:
https://linear.app/pillarproject/issue/PRO-1609/investigate-leak-memory-of-app

## Description

- Fixed memory leaks in these components
1. SvgCssUri (at CollectibleImage.js:47)
2. GovernanceCallBanner (at Home.js:274)
3. PinCodeUnlock (created by Connect(PinCodeUnlock))
4. BiometricLoginSetting (at Settings.js:96)
5. ContentLoader (at Loaders.tsx:108)
6. AllTimeLoader (at TokenAnalyticsListItem.tsx:219)
8. OfferCard (at Exchange.tsx:435)
10. ReceiveTokensWarning -  Each child in a list should have a unique "key" prop.
11. Button (at ReceiveTokensWarning.tsx:193), Button (at MenuFooter.js:128)
13. Tooltip (at TableAmount.js:93)

- Fixed some bugs
1. Smooth scrolling for large list in trading, pool activity screen and add tokens screen
2. ReceiveTokensWarning -  Each child in a list should have a unique "key" prop.
3. Issue on hooks in WalletConnectHome

## Motivation and Context

-

## How Has This Been Tested?

- Tested on Android device
- Tested on Simulator 14 PRO

https://github.com/pillarwallet/pillarwallet/assets/82652040/6ca28749-fea3-4296-bb9c-abe9dab094b6

## Types of changes


patches/react-native-gifted-charts+1.2.42.patch
```diff
@@ -10,7 +10,48 @@ index 8ef5ecc..4d11a61 100644
    let data = useMemo(() => {
      if (!props.data) {
        return [];
-@@ -2952,6 +2953,8 @@ export const LineChart = (props: propTypes) => {
+@@ -820,30 +821,36 @@ export const LineChart = (props: propTypes) => {
+     // console.log('comes here............')
```
Yash-Chaubal, line 14: Just a question about these patch files, did you make these yourself, or are these suggested from official docs? Cos these feel like a lot of changes for what a patch file should be

Also there's a console.log to remove

patches/react-native-svg+13.9.0.patch
```diff
@@ -0,0 +1,49 @@
...
+index f567b80..eb75078 100644
+--- a/node_modules/react-native-svg/src/css.tsx
++++ b/node_modules/react-native-svg/src/css.tsx
+@@ -695,14 +695,22 @@ export function SvgCssUri(props: UriProps) {
+   const { uri, onError = err, onLoad } = props;
+   const [xml, setXml] = useState<string | null>(null);
+   useEffect(() => {
+-    uri
+-      ? fetchText(uri)
++    const fetchController = new AbortController();
```
Yash-Chaubal, line 11: Same with this - as svg is a pretty common file across the whole project, it just feels a bit risky making too many changes

src/components/Tooltip/Tooltip.js
```diff
@@ -82,16 +82,23 @@ const Tooltip = (props: Props) => {
...
         toValue: isVisible ? 1 : 0,
         duration: 200,
         useNativeDriver: true,
-      }).start(() => { if (!isVisible) setVisibilityState(false); });
+      }).start(() => {
+        if (!isVisible) setVisibilityState(false);
+      });
     }
-    if (isVisible) {
+    if (isVisible && mounted) {
```
Yash-Chaubal, line 95: Isn't mounted always true here?

src/components/Tooltip/Tooltip.js
```diff
@@ -82,16 +82,23 @@ const Tooltip = (props: Props) => {
...
       setVisibilityState(true);
     }
+
+    return () => {
+      mounted = false;
```
Yash-Chaubal, line 100: I think I see what you're doing here, but `mounted` isn't a global variable so this wouldn't matter as it just gets set to true the moment the useEffect runs again - if you want to store mounted states, I think its better to use `useState`

src/screens/Bridge/Exchange-CrossChain/OfferCard.js
```diff
@@ -97,11 +97,14 @@ function OfferCard({
   const [offerInfo, setOfferInfo] = React.useState(null);
 
   React.useEffect(() => {
-    async function call() {
+    let mounted = true;
+    (async () => {
```
Yash-Chaubal, line 101: The use of `mounted` makes more sense here, but I would still recommend `useState` to have a global `mounted` variable

Also please put the async function into a variable, and then call that variable - it looks a bit messy like this

src/screens/WalletConnect/Home/WalletConnectHome.js
```diff
@@ -88,6 +88,9 @@ function WalletConnectHome() {
   const { numberOfColumns, columnWidth } = useColumnDimensions();
   const { data: sections, isFetching } = useSectionData(activeChain, numberOfColumns);
 
+  const isLight = isLightTheme();
+
+
```
Yash-Chaubal, line 93: Double line breaks here - are you using a formatter?

Yash-Chaubal (commented):
Looking good! Just had a few questions and suggestions @lalitkumarrock

src/screens/Bridge/Exchange-CrossChain/OfferCard.js
lalitkumarrock: Here main purpose of `mounted` is cleanup async function every time. `useState` is not used in this case because useState is the main cause of memory leaks. 

-> And (async() => {})() this is Immediately Invoked Function Expression
https://developer.mozilla.org/en-US/docs/Glossary/IIFE

patches/react-native-svg+13.9.0.patch
lalitkumarrock: I noticed `SvgCssUri` in react-native-svg always leaks memory. So it was important to fix this.

src/components/Tooltip/Tooltip.js
lalitkumarrock: whenever useEffect cleanup then mounted have false.

src/components/Tooltip/Tooltip.js
lalitkumarrock: Here we can't use useState. In this setVisibilityState doing a memory leak because it's call after cleanup. So we just need to stop them after cleaning.

src/components/Tooltip/Tooltip.js
Yash-Chaubal: On line 85 you're setting `mounted` as true, regardless of whether or not you have a clean up function setting it to false - at this stage in the code `mounted` would still be true since there's no delay in between. This is more apparent on line 86, where you use mounted immediately.

Maybe try using a hook instead? https://usehooks-ts.com/react-hook/use-is-mounted

src/components/Tooltip/Tooltip.js
Yash-Chaubal: Yeah makes sense, like I said above maybe the hook above would be better, or maybe useRef instead?

src/screens/Bridge/Exchange-CrossChain/OfferCard.js
Yash-Chaubal: Same comment about the mounted/clean up here - you're set mounted as true on line 100, and then immediately using it on line 103. This isn't inside an async function or anything so there is no delay between this (less than a millisecond) and it will always be true, regardless of the clean up function. It's better to use the hook here.

Re IIFE: Yeah I'm aware, its probably just personal preference but I tend to avoid using them as I find them to look messier and more difficult to read when they become more bloated. Also since IIFEs in React only are ever used inside hooks (I can't think of any other reasons to use them), I think having consistency is better

lalitkumarrock: In this `mounted` managed with hooks

src/components/Tooltip/Tooltip.js
lalitkumarrock: In this `mounted` managed with hooks

src/screens/Home/ReceiveTokensWarning.tsx
```diff
@@ -167,7 +167,7 @@ const ReceiveTokensWarning: FC<IReceiveTokensWarning> = ({ theme }) => {
             const { titleShort } = chainsConfig[chain];
 
             return (
-              <DeployRow>
+              <DeployRow key={'ordered_' + chain}>
```
IAmKio, line 170: String templates preferred

IAmKio (approved):
LGTM, please note String concat vs templating in future. Using + isn't safe vs templating

PR #3488, opened 2024-04-04
Fixed Encypt storage migration issue

lalitkumarrock:
https://linear.app/pillarproject/issue/PRO-2236/pillar-wallet-sqlite-encryption

## Description

- Fixed Encryption storage migration 

## Motivation and Context

-

## How Has This Been Tested?

- Tested on Android and iOS device

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/ios/pillarwallet/AppDelegate.mm
+++ b/ios/pillarwallet/AppDelegate.mm
@@ -8,10 +8,34 @@
 #import "RCTLinkingManager.h"
 #import "RNNotifications.h"
 
+static void ClearKeychainIfNecessary() {
+    // Checks wether or not this is the first time the app is run
+    if ([[NSUserDefaults standardUserDefaults] boolForKey:@"HAS_RUN_BEFORE"] == NO) {
+        // Set the appropriate value so we don't clear next time the app is launched
+        [[NSUserDefaults standardUserDefaults] setBool:YES forKey:@"HAS_RUN_BEFORE"];
+
+        NSArray *secItemClasses = @[
+            (__bridge id)kSecClassGenericPassword,
+            (__bridge id)kSecClassInternetPassword,
+            (__bridge id)kSecClassCertificate,
+            (__bridge id)kSecClassKey,
+            (__bridge id)kSecClassIdentity
+        ];
+
+        // Maps through all Keychain classes and deletes all items that match
+        for (id secItemClass in secItemClasses) {
+            NSDictionary *spec = @{(__bridge id)kSecClass: secItemClass};
+            SecItemDelete((__bridge CFDictionaryRef)spec);
+        }
+    }
+}
+
 @implementation AppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+  ClearKeychainIfNecessary();
+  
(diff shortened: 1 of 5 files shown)


src/services/encryptedStorage.js
```diff
@@ -70,6 +78,22 @@ EncryptedStorage.prototype.save = async function (id: string, data: Object, forc
...
+  if (isEmpty(wallet)) return Promise.resolve();
+
+  try {
+    printLog('Encrypted Storage migrating data');
+    // eslint-disable-next-line i18next/no-literal-string
+    await this.save('wallet', { wallet });
+    // eslint-disable-next-line i18next/no-literal-string
+    await storage.remove('wallet');
+  } catch (e) {
+    reportErrorLog('AsyncStorage migration to Encrypted Storage failed', { error: e });
```
IAmKio, line 92: If this fails - what happens?

ios/pillarwallet/AppDelegate.mm
```diff
@@ -8,10 +8,34 @@
...
+            SecItemDelete((__bridge CFDictionaryRef)spec);
+        }
+    }
+}
+
 @implementation AppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+  ClearKeychainIfNecessary();
```
IAmKio, line 37: can you just explain in what scenario this would be needed?

src/actions/appActions.js
```diff
@@ -190,29 +192,29 @@ export const initAppAndRedirectAction = () => {
...
-          type: TODAY_FAILED_ATTEMPTS,
-          payload: {
-            failedAttempts: {
-              numberOfFailedAttempts,
-              date: new Date(date),
-            },
-          },
-        });
-      }
+      // const currentTime = await getCurrentTime();
```
IAmKio, line 195: Do we still actually need this or is it commented out for a reason?

IAmKio (changes requested):

ios/pillarwallet/AppDelegate.mm
lalitkumarrock: Actually, encrypted storage data is stored in keychain.  still, data is stored in the keychain if we uninstall the app. used this to remove data when first time open app

src/services/encryptedStorage.js
IAmKio: @lalitkumarrock I know sorry but why?

IAmKio: I just need to know what is the fallback scenario here

ios/pillarwallet/AppDelegate.mm
IAmKio: Got it thanks

src/services/encryptedStorage.js
lalitkumarrock: Here failure chances are very low. if fails then open the onboarding screen.
I pushed one changes. [de04d01](https://github.com/pillarwallet/pillarwallet/pull/3488/commits/de04d0106f81c56b86d736433000c01877986647)

IAmKio (approved):

